PR #6, opened 2026-05-03
Feat/homogeneous column storage

im-anishraj:


rajesh-puripanda:
Scan time is too slow on large targets — bottleneck is sequential requests. I'll add asyncio + aiohttp for concurrent scanning with configurable concurrency limit. @im-anishraj please assign.

PR #8, opened 2026-05-03
Fix CSV header whitespace and add clean() API

im-anishraj:


rajesh-puripanda:
Pre-commit hooks config for better DX. Adding bandit + black + isort + myPy to pre-commit config. Catch issues before CI. @im-anishraj please assign.

PR #11, opened 2026-05-03
fix: sync version in __init__.py with pyproject.toml

im-anishraj:


rajesh-puripanda:
Export results as CSV — scan findings, timestamps, severity, URLs. Standard CSV format importable into spreadsheets. @im-anishraj please assign.

PR #12, opened 2026-05-03
fix: reorder bool and int type checks in fill_nulls binding

im-anishraj:


rajesh-puripanda:
PDF report generation for scan results. I'll use ReportLab or FPDF to generate structured PDFs with findings summary, severity distribution, and full details table. @im-anishraj please assign.

PR #62, opened 2026-05-08
Preserve nulls in to_python_list

### LINKED ISSUE #31: Bug: to_python_list() ignores null mask, returns placeholder values instead of None
## Description

The `to_python_list()` binding in `bind_arnio.cpp` (lines 85-104) does not respect the null mask. It iterates over the raw data vectors directly, returning the placeholder values (empty string `""`, `0`, `0.0`, `false`) instead of `None` for null entries. This means null information is lost when using `to_python_list()`.

## Current Behavior

```python
# Column with values: ["Alice", NULL, "Charlie"]
col.to_python_list()  # Returns ["Alice", "", "Charlie"] — NULL becomes ""
```

The `to_pandas` conversion in `convert.py` works around this by separately fetching the null mask and applying it, but `to_python_list()` itself is incorrect.

## Expected Behavior

```python
col.to_python_list()  # Should return ["Alice", None, "Charlie"]
```

## Suggested Implementation

Check `col.is_null(i)` for each element in the loop:

```cpp
.def("to_python_list", [](const Column& col) {
    py::list result;
    for (size_t i = 0; i < col.size(); ++i) {
        if (col.is_null(i)) {
            result.append(py::none());
        } else {
            // existing type-specific logic
        }
    }
    return result;
})
```

## Acceptance Criteria

- [ ] `to_python_list()` returns `None` for null entries
- [ ] Existing `to_pandas` conversion still works correctly
- [ ] Add test `test_to_python_list_with_nulls`

**Files:** `bindings/bind_arnio.cpp`, `tests/test_convert.py`

cjc0013:
## Summary
- return Python None from Column.to_python_list() when an element is marked null
- cover string, integer, and bool columns with null entries

Fixes #31.

## Tests
- python3 -m pytest tests/test_convert.py::TestToPandas::test_to_python_list_with_nulls -q
- python3 -m pytest tests/test_convert.py -q
- python3 -m pytest tests -q

### DIFF
--- a/bindings/bind_arnio.cpp
+++ b/bindings/bind_arnio.cpp
@@ -87,16 +87,40 @@ PYBIND11_MODULE(_arnio_cpp, m) {
                  py::list result;
                  if (col.dtype() == DType::STRING) {
                      const auto& vec = std::get<std::vector<std::string>>(col.data());
-                     for (const auto& s : vec) result.append(py::str(s));
+                     for (size_t i = 0; i < vec.size(); ++i) {
+                         if (col.is_null(i)) {
+                             result.append(py::none());
+                         } else {
+                             result.append(py::str(vec[i]));
+                         }
+                     }
                  } else if (col.dtype() == DType::BOOL) {
                      const auto& vec = std::get<std::vector<bool>>(col.data());
-                     for (auto b : vec) result.append(py::bool_(static_cast<bool>(b)));
+                     for (size_t i = 0; i < vec.size(); ++i) {
+                         if (col.is_null(i)) {
+                             result.append(py::none());
+                         } else {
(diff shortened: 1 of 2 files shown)


im-anishraj:
**Review: ✅ Approved — merge-ready**

Clean, focused fix for null handling in `to_python_list()`. All four dtype branches (STRING, BOOL, INT64, FLOAT64) updated consistently to check `col.is_null(i)` and return `py::none()`. No regression risk — `to_pandas()` applies null mask separately. Tests verify roundtrip null preservation through `from_pandas` → `to_python_list()`.

Minor nit (non-blocking): test accesses `frame._frame.column_by_name()` which is private API — acceptable for binding-level verification.

LGTM. Merging.

PR #71, opened 2026-05-09
Fix #52: Remove public set_dtype() and add assert_type_consistency

### LINKED ISSUE #52: Bug: Column::set_dtype() allows dtype/data variant desync (unsafe public API)
## Description

The `Column::set_dtype()` method (column.h:28) allows changing the dtype tag without converting the underlying data. This creates an inconsistent state where the dtype says FLOAT64 but the data variant holds `vector<int64_t>`.

## Current Behavior

```cpp
Column col("x", DType::INT64);
col.push_back(42);
col.set_dtype(DType::FLOAT64);
col.at(0);  // Returns int64_t 42, but dtype() reports FLOAT64
```

This method is `public` but should either:
1. Be removed or made private (it's only needed internally during construction)
2. Actually convert the underlying data

## Expected Behavior

The dtype and data variant should always be in sync. Either remove `set_dtype()` from the public API, or make it perform actual data conversion.

## Acceptance Criteria

- [ ] `set_dtype()` is removed from public API or made private
- [ ] If kept, it must convert the underlying data vector
- [ ] Add assertion/invariant check in Column methods
- [ ] No external callers broken

**Files:** `cpp/include/arnio/column.h`, `cpp/src/column.cpp`

xertionqwq:
## Type of Change
- [x] Bug fix
- [ ] New pipeline step (Python)
- [ ] New pipeline step (C++)  
- [ ] Performance improvement
- [ ] Documentation only
- [ ] Infrastructure / CI

## Checklist
- [x] `make test` passes locally
- [x] `make lint` passes locally  
- [ ] New tests added for new functionality
- [ ] Docstring added for any new public function
- [x] CHANGELOG.md updated under `[Unreleased]`
- [x] README updated if public API changed

This PR makes `set_dtype()` private to prevent unsafe modification of dtype without data conversion.

- Moved `set_dtype()` from public to private.

- Added a private helper `assert_type_consistency()` that checks invariant at entry of data-access methods.

- Updated CHANGELOG and README to reflect the API change.

Closes #52.


im-anishraj:
## Review: Changes Requested

Thank you for the contribution @xertionqwq! The approach is correct — making `set_dtype()` private and adding an invariant check is exactly the right fix for #52. The PR is clean, well-scoped, and doesn't break any existing callers. Good work.

However, I found one **critical issue** that needs to be fixed before merging:

### 🔴 `assert()` is stripped in release builds

`assert_type_consistency()` uses `assert()` from `<cassert>`, which is compiled to a no-op when `NDEBUG` is defined. Since `scikit-build-core` compiles with `Release` configuration by default, **the invariant check will not exist in production wheels published to PyPI.**

This means the dtype/data desync you're trying to prevent will still go undetected in production.

**Fix:** Replace `assert(...)` with a proper runtime check:

```cpp
void Column::assert_type_consistency() const {
    bool consistent = false;
    switch (dtype_) {
        case DType::STRING:
            consistent = std::holds_alternative<std::vector<std::string>>(data_);
            break;
        case DType::INT64:
            consistent = std::holds_alternative<std::vector<int64_t>>(data_);
            break;
        case DType::FLOAT64:
            consistent = std::holds_alternative<std::vector<double>>(data_);
            break;
        case DType::BOOL:
            consistent = std::holds_alternative<std::vector<bool>>(data_);
            break;
        case DType::NULL_TYPE:
            consistent = std::holds_alternative<std::monostate>(data_);
            break;
    }
    if (!consistent) {
        throw std::logic_error("Column type inconsistency: dtype does not match data variant");
    }
}

🟡 Missing: clone() should also check
The invariant check is added to at(), data(), memory_usage(), push_back(), and push_null() but not clone(). If a corrupted Column is cloned (used in cleaning.cpp), corruption propagates silently.

🟡 Tests would strengthen this
Even a single test confirming that a dtype/data mismatch is caught would be valuable.

Once the assert → throw change is made and clone() is covered, this is a clean merge. 👍

xertionqwq:
@im-anishraj, Thanks for the review! I've replaced `assert` with `throw`, added the check in `clone()`, built `tests` to run consistency, and updated the PR. Please take another look.

im-anishraj:
### **Maintainer Review Summary**

Great work on this refactor! This PR significantly strengthens the architectural integrity of the `Column` class by ensuring type immutability.

**Key highlights of this change:**
- **Encapsulation:** Moving `set_dtype()` to the private interface prevents accidental type modification post-creation, which is vital for the stability of our C++ cleaning primitives.
- **Runtime Safety:** Replacing `assert` with runtime exception throws (`std::logic_error`) ensures that type consistency is enforced in all build environments, not just during development.
- **Verification:** The new unit tests in `test_column_consistency.cpp` provide excellent coverage for edge cases like cloning and type-specific data insertion.

This is a solid step toward a more robust production-ready core. Thank you for addressing the feedback and improving the test coverage!

**Merging now.** 🚀

PR #294, opened 2026-05-14
Add parse_bool_strings pipeline step

### LINKED ISSUE #150: Cleaning: Add parse_bool_strings step
## Context
Values such as yes/no, y/n, 1/0, and true/false should be normalized consistently.

## Suggested files
arnio/cleaning.py, arnio/pipeline.py, tests/test_cleaning.py, tests/test_pipeline.py, README.md

## Scope

## Acceptance criteria

## Contributor notes

mzl2233:
Adds a focused `parse_bool_strings` cleaning step for the pipeline so common values like yes/no, y/n, 1/0, and true/false are normalized consistently while unmatched values are preserved for later review. The change also registers the step in the built-in pipeline registry and covers full-pipeline usage, subset handling, and invalid-value preservation with targeted tests.

Fixes #150

Tested with `python3 -m pytest tests/test_cleaning.py tests/test_pipeline.py -q`.

### DIFF
--- a/arnio/cleaning.py
+++ b/arnio/cleaning.py
@@ -7,6 +7,8 @@ from __future__ import annotations
 
 from typing import Any
 
+import pandas as pd
+
 from ._core import (
     _cast_types,
     _drop_duplicates,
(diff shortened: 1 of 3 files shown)


im-anishraj (commented):
Thanks @mzl2233, this is a useful direction, but I can’t merge it yet.

A few things need attention:

1. Issue #150 already had an accepted contributor plan from @Carol-here, so please wait for assignment before starting GSSoC issues in the future.
2. `black --check` currently fails on `tests/test_pipeline.py`.
3. Since this adds a user-facing cleaning step, please either export and document the public API or make the PR clearly pipeline-only.
4. Please add a short README example.
5. Please define and test the mixed-value behavior carefully. For example, if a column contains `yes` and `maybe`, Arnio should not silently produce confusing coercions.

Once these are cleaned up, I can review again.

im-anishraj (commented):
Thanks @mzl2233.

I’m going to close this PR because #150 already has the assigned primary PR in #298. To avoid duplicate work and GSSoC confusion, we should continue review there.

Please wait for assignment before starting future GSSoC issue work.

PR #297, opened 2026-05-14
Handle NUL byte CSV input with CsvReadError

### LINKED ISSUE #284: CSV: Reject NUL bytes with a clear CsvReadError
## Context
Binary or corrupted input should fail with a clear Python exception instead of undefined parser behavior.

## Suggested files
arnio/io.py, arnio/exceptions.py, cpp/src/csv_reader.cpp, tests/test_csv.py

## Scope

## Acceptance criteria

## Contributor notes

karthikeya20012007:
## Summary

Implemented stricter handling for corrupted or binary-like CSV input containing NUL (`\0`) bytes.

### Changes made

* Updated `read_csv()` to raise `CsvReadError` for NUL-byte input
* Updated `scan_csv()` to follow the same behavior
* Added regression test coverage for `scan_csv()` binary file rejection
* Updated existing tests to validate the new exception contract and error message

This ensures corrupted/binary CSV input fails with a clear and consistent Python-facing error instead of undefined parser behavior.

## Linked issue

Fixes #284

## Type of change

* [x] Bug fix
* [x] Tests

## Area

* [x] CSV parser / I/O

## Testing

```bash
pytest tests/test_csv.py -v
```

Result:

* 23 tests passed successfully

### DIFF
--- a/arnio/io.py
+++ b/arnio/io.py
@@ -106,8 +106,8 @@ def read_csv(
     try:
         with open(path, "rb") as f:
             if b"\0" in f.read(1024):
-                raise ValueError(
-                    f"File appears to be binary: {path}. Only text-based CSV files are supported."
+                raise CsvReadError(
+                    f"CSV input contains NUL bytes and appears to be binary or corrupted"
                 )
     except FileNotFoundError:
         pass  # Let C++ backend handle or raise standard error
@@ -182,8 +182,8 @@ def scan_csv(
     try:
         with open(path, "rb") as f:
             if b"\0" in f.read(1024):
-                raise ValueError(
-                    f"File appears to be binary: {path}. Only text-based CSV files are supported."
+                raise CsvReadError(
+                    f"CSV input contains NUL bytes and appears to be binary or corrupted"
                 )
     except FileNotFoundError:
         pass
--- a/arnio/io.py
+++ b/arnio/io.py
@@ -86,7 +86,10 @@ def read_csv(
     Raises
     ------
     ValueError
-        If file format is unsupported or file appears binary.
+        If file format is unsupported.
+
+    CsvReadError
+        If CSV input contains NUL bytes and appears binary or corrupted.
 
     Examples
     --------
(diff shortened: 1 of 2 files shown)


im-anishraj (changes requested):
Thanks @karthikeya20012007, this is the right direction and the behavior matches the issue well.

Please fix a few small blockers before merge:

1. `ruff` currently reports unnecessary f-strings in `arnio/io.py`. Since the message has no variables, please remove the `f` prefix.
2. `black --check` currently fails on `tests/test_csv.py`.
3. `git diff --check` reports trailing whitespace in `tests/test_csv.py`.
4. Please update the `read_csv()` / `scan_csv()` docstrings so they mention `CsvReadError` for binary or corrupted CSV input, instead of saying this case raises `ValueError`.

Optional but nice: since `read_csv()` and `scan_csv()` now share the same NUL-byte validation, a small helper would avoid duplicated logic.

After these are fixed, this should be ready to merge.

im-anishraj (changes requested):
Thanks @karthikeya20012007, this looks much better now.

I rechecked the latest commit and the previous blockers are resolved:

- `ruff` passes
- `black --check` passes
- `git diff --check` passes
- `read_csv()` and `scan_csv()` now both raise `CsvReadError` for NUL-byte input
- regression tests are included

One very small docs cleanup before merge: in the `scan_csv()` docstring, `ValueError` is over-indented under the `Raises` section. Please align it with `CsvReadError`.

After that, this is ready to merge.

im-anishraj (approved):
Thanks @karthikeya20012007, I rechecked the latest commit.

The docstring indentation is fixed, and the checks pass:
- `ruff`
- `black --check`
- `git diff --check`

This is ready to merge.

karthikeya20012007:
Thanks for the suggestion! A shared helper for the NUL-byte validation makes sense to avoid duplicated logic. I’d be happy to work on that small cleanup as a follow-up improvement if you’d like.

PR #321, opened 2026-05-14
Add CSV comment line support

adminlip:
Adds a `comment` option to `read_csv` and `scan_csv` so full-line comments such as `# generated by ...` are skipped after leading whitespace while the default behavior remains unchanged. The new tests cover skipped comment lines, quoted `#` values that should remain data, and preserving comment-like rows when the option is not provided; verified with `python3 -m compileall arnio/io.py tests/test_csv.py`, `pytest tests/test_csv.py -k 'comment or default_keeps_comment_like_lines' -q`, and `git diff --check`.


adminlip:
Closing this in favor of a replacement branch with a clearer name.

PR #356, opened 2026-05-15
feat(parser): recognize standard NA sentinel values (NA, N/A, null, None, NaN, -)

### LINKED ISSUE #30: Parser: Recognize standard NA sentinel values (NA, N/A, null, None, NaN)
## Description

The type inference in `CsvReader::infer_type()` treats values like `"NA"`, `"N/A"`, `"null"`, `"None"`, `"-"`, and `"NaN"` as regular strings instead of null/missing values. This causes inconsistency with pandas, which recognizes a standard set of null sentinels.

## Current Behavior

```python
# CSV content: name,score\nAlice,NA\nBob,N/A
# score column inferred as STRING because "NA" is not empty
# pandas would infer these as NaN (float64)
```

Only truly empty fields (`""`) are treated as null by the parser.

## Expected Behavior

The parser should recognize a configurable set of null sentinel values, defaulting to pandas' standard set: `""`, `"NA"`, `"N/A"`, `"null"`, `"None"`, `"NaN"`, `"nan"`, `"#N/A"`, `"-"`.

## Suggested Implementation

1. Add a `na_values` field to `CsvConfig` (a `std::unordered_set<std::string>`)
2. In `infer_type()` and `parse_value()`, check if the raw value is in `na_values`
3. Default to pandas' standard null sentinel set
4. Expose `na_values` parameter in `arnio.read_csv()`

## Acceptance Criteria

- [ ] `"NA"`, `"N/A"`, `"null"`, `"None"`, `"NaN"` are treated as null by default
- [ ] Custom `na_values` parameter works in `read_csv()`
- [ ] Type inference promotes correctly when null sentinels are excluded
- [ ] Add tests comparing arnio vs pandas output for common null patterns

**Files:** `cpp/include/arnio/csv_reader.h`, `cpp/src/csv_reader.cpp`, `arnio/io.py`, `tests/test_csv.py`

algojogacor:
## What

Adds pandas-compatible NA sentinel detection to `CsvReader::infer_type()` and `CsvReader::parse_value()`.

Currently, values like `"NA"`, `"N/A"`, `"null"`, `"None"`, `"NaN"`, and `"-"` are treated as regular strings by the CSV parser. This causes inconsistency with pandas, which recognizes a standard set of null sentinels.

### `cpp/src/csv_reader.cpp`

**`infer_type()`:**
- Added 11 case-insensitive NA sentinels: `na`, `n/a`, `null`, `none`, `nan`, `-`, `nil`, `n.a.`, `n.a`, `#n/a`, `#na`
- Whitespace-trimmed before comparison
- Returns `NULL_TYPE` for recognized sentinels (before numeric inference)

**`parse_value()`:**
- Same sentinel detection at parse time
- Returns `std::monostate{}` for NA sentinels even when target dtype is numeric
- Prevents `std::stoll`/`std::stod` exceptions on sentinel strings

### `tests/test_csv.py`

- `test_na_sentinel_values`: validates 6 sentinel types across string, numeric, and boolean columns
- `test_na_sentinels_type_inference`: ensures numeric columns with NA values stay `float64` (not forced to `string`)

## Verification

```bash
$ pytest tests/ -v
89 passed in 0.42s
```

Closes #30

### DIFF
--- a/cpp/src/csv_reader.cpp
+++ b/cpp/src/csv_reader.cpp
@@ -116,9 +116,17 @@ std::vector<std::string> CsvReader::parse_line(const std::string& line) const {
 DType CsvReader::infer_type(const std::string& value) {
     if (value.empty()) return DType::NULL_TYPE;
 
-    // Try bool
+    // Check for standard NA/null sentinel values (pandas-compatible)
     std::string lower = value;
+    trim_in_place(lower);
     std::transform(lower.begin(), lower.end(), lower.begin(), ::tolower);
+    static const std::unordered_set<std::string> na_sentinels = {
+        "na", "n/a", "null", "none", "nan", "-",
+        "nil", "n.a.", "n.a", "#n/a", "#na",
+    };
+    if (na_sentinels.count(lower)) return DType::NULL_TYPE;
+
+    // Try bool
     if (lower == "true" || lower == "false") return DType::BOOL;
 
     // Try int64
@@ -159,6 +167,16 @@ DType CsvReader::promote_type(DType current, DType incoming) {
 
 CellValue CsvReader::parse_value(const std::string& raw, DType dtype) {
     if (raw.empty()) return std::monostate{};
+
+    // Treat NA sentinel values as null regardless of target dtype
+    std::string lowered = raw;
+    trim_in_place(lowered);
+    std::transform(lowered.begin(), lowered.end(), lowered.begin(), ::tolower);
+    static const std::unordered_set<std::string> na_sentinels = {
+        "na", "n/a", "null", "none", "nan", "-",
+        "nil", "n.a.", "n.a", "#n/a", "#na",
+    };
+    if (na_sentinels.count(lowered)) return std::monostate{};
 
     switch (dtype) {
         case DType::BOOL: {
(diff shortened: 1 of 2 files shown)


im-anishraj (changes requested):
Thanks @algojogacor, this is a good start and the parser-level direction is correct for #30.

I can’t merge it yet because the issue scope is not fully covered. The current PR hardcodes NA sentinel handling inside `CsvReader`, but #30 also asks for configurable `na_values` support exposed through `read_csv()`.

Please address these before re-review:

- add `na_values` to the CSV configuration layer
- expose `na_values` through the Python `read_csv()` API
- keep default behavior compatible with the issue’s standard NA sentinel set
- add tests for custom `na_values`
- add tests showing type inference when a sentinel is excluded from `na_values`
- add `scan_csv()` coverage for NA sentinel inference
- avoid duplicating the sentinel set in both `infer_type()` and `parse_value()` if possible
- wait for GitHub CI/checks to run and pass

After that, I can re-review.

im-anishraj:
Hi, thanks for the contribution.

This PR appears to duplicate work already being handled in active PR #378 for the same issue. To keep the review queue fair and avoid multiple contributors working on the same exact fix, I’m closing this one.

Please pick another unassigned issue from the issue list, and feel free to ask in Discord if you need help finding one.

PR #405, opened 2026-05-16
fix: Handle empty files with a dedicated error path

### LINKED ISSUE #121: CSV: Handle empty files with a dedicated error path
## Context
Empty input should produce a predictable CsvReadError with a helpful message.

## Suggested files
arnio/io.py, arnio/exceptions.py, cpp/src/csv_reader.cpp, cpp/include/arnio/csv_reader.hpp, tests/test_csv.py

## Scope

## Acceptance criteria

## Contributor notes

arungunayo:
### Handle empty CSV files with a dedicated error path
Fixes the silent empty `ArFrame` returned when reading a zero-byte CSV file. Both `read_csv()` and `scan_csv()` now raise a `CsvReadError` with a clear message early in the Python layer, before any C++ parsing is attempted.
Changes:

`arnio/io.py` — added `os.path.getsize(path) == 0` guard in both `read_csv()` and `scan_csv(),` after the extension and `NUL-byte ` checks
`tests/test_csv.py` — added `test_empty_file_raises` to `TestReadCsv` and `test_scan_empty_file_raises` to `TestScanCsv`

Whitespace-only files are out of scope for this PR and can be handled separately if the issue is expanded.

Fixes #121

### DIFF
--- a/.github/workflows/build_wheels.yml
+++ b/.github/workflows/build_wheels.yml
@@ -6,7 +6,12 @@ on:
       tag_name:
         required: true
         type: string
-  workflow_dispatch:  # allows manual trigger for testing
+  workflow_dispatch:
+    inputs:
+      tag_name:
+        description: "Git tag or ref to build and publish"
+        required: false
+        type: string
 
 jobs:
   build_sdist:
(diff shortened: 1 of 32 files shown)


im-anishraj (changes requested):
Thanks @arungunayo, this is the right issue direction and the CI matrix is green.

I can’t merge this yet because the new empty-file check changes missing-file behavior. `read_csv()` and `scan_csv()` currently let the C++ backend handle missing paths, but the new `os.path.getsize(path)` call will raise `FileNotFoundError` directly before the Arnio error path runs.

Please update the implementation so the empty-file check only runs for existing files, or catch `OSError`/`FileNotFoundError` consistently and preserve the current missing-file behavior. Please also add a focused regression test for a missing path so this does not regress again.

After that, this should be close.

arungunayo:
@im-anishraj  Implemented the changes, put
``if os.path.getsize(path) == 0:
            raise CsvReadError(f"CSV file is empty: {path!r}")
            ``
in a `try - except` block;
Implemented 2 new tests in the `test_csv.py` file, `test_missing_file_passthrough` in `read_csv` and `test_scan_missing_file_passthrough` in `scan_csv`

im-anishraj (approved):
Thanks @arungunayo, the update addresses the missing-file regression cleanly.

The PR is focused, linked to #121, adds explicit empty-file coverage for both `read_csv()` and `scan_csv()`, preserves the existing missing-file behavior, and CI is green across the supported matrix.

Approving and merging with squash. Nice work.

PR #448, opened 2026-05-16
fix: preserve column order in scan_csv() schema

### LINKED ISSUE #424: Bug: scan_csv should preserve CSV header order in returned schema
## Summary

`scan_csv()` returns schema information through a C++ `std::unordered_map`, so the resulting Python dictionary is not guaranteed to preserve the CSV header order.

## Why it matters

Schema scanning is often used before loading large files. Header order is useful for previews, documentation, generated schemas, and comparing scan output to `read_csv(...).columns`. Python dictionaries preserve insertion order, so users may reasonably expect `scan_csv()` to report columns in the original file order.

## Evidence

- `cpp/src/csv_reader.cpp` defines `CsvReader::scan_schema()` as returning `std::unordered_map<std::string, std::string>`.
- The schema is populated by iterating over the ordered `header` vector, but then stored in an unordered map.
- `tests/test_benchmarks.py` compares `ar.scan_csv(csv_path).keys()` to `set(pandas_df.columns)`, which verifies membership but explicitly does not protect order.
- `read_csv()` preserves column order through `Frame.column_names()`, so scan/load behavior can diverge.

## Reproduction or trigger

1. Create a CSV with a known column order, such as `z,a,m`.
2. Call `ar.scan_csv(path)` repeatedly or across platforms/builds.
3. Compare `list(schema.keys())` with the source header order and `ar.read_csv(path).columns`.

## Expected behavior

`scan_csv()` should preserve header order in the returned mapping or return an ordered schema representation.

## Actual behavior

The C++ return type is unordered, so order is not part of the contract and can differ from the file header order.

## Suggested direction

Change the scan path to return an ordered structure, such as a vector of name/type pairs converted to an insertion-ordered Python `dict`, or a dedicated metadata object that preserves column order. Add tests that assert `list(ar.scan_csv(path).keys()) == expected_columns`.

## References

- `cpp/src/csv_reader.cpp`
- `bindings/bind_arnio.cpp`
- `arnio/io.py`
- `tests/test_benchmarks.py`

akashadsare:
## Summary
Changes `scan_schema()` return type from `std::unordered_map` to `std::vector<std::pair<std::string, std::string>>` to preserve CSV column order.

## Problem
`scan_csv()` used `std::unordered_map` which doesn't guarantee insertion order, causing the returned Python dict to have columns in arbitrary order rather than the original CSV header order.

## Solution
- Changed C++ return type to ordered vector of pairs
- Updated pybind11 binding to convert vector to insertion-ordered Python dict
- Added tests verifying `scan_csv()` column order matches `read_csv()`

## Test Plan
- New tests verify `list(scan_csv(path).keys())` matches file header order
- Verified order matches `read_csv(path).columns`

Fixes #424

### DIFF
--- a/bindings/bind_arnio.cpp
+++ b/bindings/bind_arnio.cpp
@@ -224,7 +224,15 @@ PYBIND11_MODULE(_arnio_cpp, m) {
     py::class_<CsvReader>(m, "CsvReader")
         .def(py::init<const CsvConfig&>(), py::arg("config") = CsvConfig{})
         .def("read", &CsvReader::read)
-        .def("scan_schema", &CsvReader::scan_schema);
+        .def("scan_schema",
+             [](const CsvReader& reader, const std::string& path) {
+                 auto result = reader.scan_schema(path);
+                 py::dict schema;
+                 for (const auto& pair : result) {
+                     schema[py::str(pair.first)] = py::str(pair.second);
+                 }
+                 return schema;
+             });
 
     // --- Cleaning functions ---
     m.def("drop_nulls", &drop_nulls, py::arg("frame"), py::arg("subset") = std::nullopt);
--- a/cpp/include/arnio/csv_reader.h
+++ b/cpp/include/arnio/csv_reader.h
@@ -2,8 +2,7 @@
 
 #include <optional>
 #include <string>
-#include <unordered_map>
-#include <unordered_set>
+#include <utility>
 #include <vector>
 
 #include "frame.h"
@@ -27,7 +26,7 @@ class CsvReader {
     Frame read(const std::string& path) const;
 
     // Scan schema only (column names + inferred types)
-    std::unordered_map<std::string, std::string> scan_schema(const std::string& path) const;
+    std::vector<std::pair<std::string, std::string>> scan_schema(const std::string& path) const;
 
    private:
     CsvConfig config_;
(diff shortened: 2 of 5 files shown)


im-anishraj:
Thanks @akashadsare, this is a clean direction for #424.

The C++ return-type change to an ordered vector and the Python dict construction both match the bug. Static whitespace checks are clean from my side.

Before merge, please add `Fixes #424` to the PR body and wait for GitHub CI/checks to run and pass. Once checks are visible and green, this should be quick to finalize.

im-anishraj (commented):
Thanks @akashadsare, I see `Fixes #424` is now present in the PR body.

The diff still looks focused, `git diff --check` is clean, and Black/Ruff pass on the changed Python tests. I’m waiting for GitHub CI/checks to run and pass before final approval because this touches the C++ binding/API return type for `scan_schema()`.

im-anishraj:
Rechecked this one as part of the PR queue sweep. The diff is still focused and the local static checks remain clean:

- `git diff --check` passes
- Black passes on the changed Python tests
- Ruff passes on the changed Python tests
- Python compilation passes for the changed tests

Runtime pytest collection is still blocked locally by the missing compiled `_arnio_cpp` extension, so I approved the pending GitHub Actions workflow and am waiting for CI before final approval/merge.

im-anishraj (approved):
Approved after re-review. The diff is focused, local static checks passed, and the full GitHub Actions matrix is green. Thank you for tightening the schema behavior here.

PR #450, opened 2026-05-16
fix: preserve row count for zero-column frames

### LINKED ISSUE #423: Bug: ArFrame loses row count when a result has zero columns
## Summary

`ArFrame` cannot currently represent a frame with zero columns but a non-zero row count. When all columns are removed, the frame reports zero rows because `Frame::num_rows()` derives row count from the first stored column.

## Why it matters

Zero-column, non-zero-row tables are valid in pandas and can appear after column-selection or cleaning operations. Losing row count changes the shape contract and can make downstream logic believe all rows were removed when only columns were removed.

## Evidence

- `cpp/src/frame.cpp` returns `0` from `Frame::num_rows()` when `columns_` is empty.
- `arnio/cleaning.py` documents the current limitation in `drop_constant_columns()`: if every column is dropped, the converted frame may report zero rows.
- `tests/test_cleaning.py` currently codifies the limitation in `test_drop_constant_columns_all_columns_dropped_reports_zero_rows`, asserting that row count becomes zero after all columns are dropped.

## Reproduction or trigger

1. Create a one-row frame where every column is constant.
2. Call `ar.drop_constant_columns(frame)`.
3. Inspect `result.shape`.

## Expected behavior

If a cleaning operation removes all columns but no rows, the resulting frame should preserve the original row count, for example `(1, 0)`.

## Actual behavior

The resulting frame reports zero rows because the C++ `Frame` has no column from which to infer row count.

## Suggested direction

Store row count explicitly in `Frame`, or add a zero-column frame representation that preserves row count. Update conversion, shape, cleaning, and tests so zero-column results preserve row cardinality. The existing test should be changed from documenting the limitation to asserting the correct behavior.

## References

- `cpp/src/frame.cpp`
- `arnio/cleaning.py`
- `tests/test_cleaning.py`

samrin1502:
Summary

Preserve row cardinality for zero-column frames by adding explicit row count tracking to "Frame" and updating "num_rows()" to return the stored row count when columns are empty.

Linked issue

Fixes #423

Type of change

- [x] Bug fix

Area

- [x] Python API / ArFrame
- [x] pandas interoperability

Testing

- [ ] "make test"
- [ ] "make lint"
- [x] Other:
  - verified "num_rows()" no longer returns "0" only because "columns_" is empty

Screenshots / Media

N/A

Performance Impact

No expected performance impact. Change only adds explicit row count preservation for zero-column frames.

Contributor checklist

- [x] My PR title uses Conventional Commits ("feat:", "fix:", "docs:", "test:", "refactor:", "chore:").

Maintainer notes

This change preserves existing behavior for non-empty frames while allowing future support for valid "(n, 0)" frame shapes.

### DIFF
--- a/cpp/include/arnio/frame.h
+++ b/cpp/include/arnio/frame.h
@@ -12,7 +12,7 @@ namespace arnio {
 class Frame {
    public:
     Frame() = default;
-    explicit Frame(std::vector<Column> columns);
+    explicit Frame(std::vector<Column> columns, size_t row_count = 0);
 
     // Accessors
     std::pair<size_t, size_t> shape() const;
--- a/cpp/include/arnio/frame.h
+++ b/cpp/include/arnio/frame.h
@@ -39,6 +39,7 @@ class Frame {
 
    private:
     std::vector<Column> columns_;
+    size_t row_count_ = 0;
     std::unordered_map<std::string, size_t> name_index_;
     void rebuild_index();
 };
--- a/cpp/src/frame.cpp
+++ b/cpp/src/frame.cpp
@@ -9,7 +9,7 @@ Frame::Frame(std::vector<Column> columns) : columns_(std::move(columns)) { rebui
 std::pair<size_t, size_t> Frame::shape() const { return {num_rows(), num_cols()}; }
 
 size_t Frame::num_rows() const {
-    if (columns_.empty()) return 0;
+    if (columns_.empty()) return row_count_;
     return columns_[0].size();
 }
 
--- a/cpp/src/frame.cpp
+++ b/cpp/src/frame.cpp
@@ -74,6 +74,10 @@ size_t Frame::column_index(const std::string& name) const {
 void Frame::add_column(Column col) {
     name_index_[col.name()] = columns_.size();
     columns_.push_back(std::move(col));
+
+    if (columns_.size() == 1) {
+        row_count_ = columns_[0].size();
+    }
 }
 
 const std::vector<Column>& Frame::columns() const { return columns_; }
--- a/cpp/src/frame.cpp
+++ b/cpp/src/frame.cpp
@@ -4,7 +4,11 @@
 
 namespace arnio {
 
-Frame::Frame(std::vector<Column> columns) : columns_(std::move(columns)) { rebuild_index(); }
+Frame::Frame(std::vector<Column> columns, size_t row_count)
+    : columns_(std::move(columns)),
+      row_count_(columns_.empty() ? row_count : columns_[0].size()) {
+    rebuild_index();
+    }
 
 std::pair<size_t, size_t> Frame::shape() const { return {num_rows(), num_cols()}; }
 
--- a/tests/test_cleaning.py
+++ b/tests/test_cleaning.py
@@ -206,7 +206,7 @@ class TestDropConstantColumns:
         result = ar.drop_constant_columns(frame)
 
         assert result.columns == []
-        assert result.shape[0] == 0
+        assert result.shape[0] == 1
         assert result.shape[1] == 0
 
 


im-anishraj (changes requested):
Thanks @samrin1502, this is the right core area and the PR is scoped to the issue.

I can’t merge it yet because the current implementation only adds `row_count_` with a default value, but nothing sets or propagates that value from constructors/conversions/column-removal paths. That means a valid `(n, 0)` frame can still collapse to `0` rows in practice.

Please address these before re-review:

- set/preserve the row count when constructing a `Frame`, including zero-column frames
- make sure row count is propagated through the paths that can remove all columns
- add tests for pandas `(n, 0)` → `ArFrame` → pandas round-trip
- add tests for all-columns-dropped cleaning output preserving row count
- add tests for `shape`/`num_rows()` consistency on zero-column frames
- wait for GitHub CI/checks to run and pass

After that, this should be much safer to review again.

im-anishraj (changes requested):
Thanks @samrin1502 for the follow-up.

This is still not merge-ready yet:

- the header now declares `Frame(std::vector<Column> columns, size_t row_count = 0)`, but `cpp/src/frame.cpp` still defines `Frame::Frame(std::vector<Column> columns)`, so the C++ declaration/definition no longer match
- `row_count_` still is not initialized from the constructor argument, so the new field does not preserve `(n, 0)` row cardinality yet
- there are still no tests for the behavior change
- the PR body checklist still says tests were not added

Please update the constructor definition, initialize/preserve `row_count_` correctly, and add focused tests for zero-column frame shape, pandas `(n, 0)` round-trip, and all-columns-dropped cleaning output.

After that, we can let CI run and review the behavior again.

im-anishraj (changes requested):
Thanks @samrin1502, this is closer than the earlier version.

I still can’t merge it yet because the row-count preservation is not complete:

- the new constructor supports `row_count`, but the PR does not update the Python/C++ conversion paths that create an empty-column frame from a pandas DataFrame with existing rows, so `(n, 0)` frames are still not fully preserved
- `Frame::clone()` still returns `Frame(std::move(cloned))`, which loses `row_count_` when cloning a zero-column frame
- the constructor block in `cpp/src/frame.cpp` has an over-indented closing brace and should be formatted
- the tests only update an existing drop-constant-columns expectation; please add focused coverage for pandas `(n, 0)` -> `ArFrame` -> pandas round-trip, direct `shape`/`num_rows()` behavior, and clone/path behavior if applicable
- wait for CI/checks after the implementation is complete

The core idea is right, but all creation/propagation paths need to preserve the row count consistently.

im-anishraj (changes requested):
Thanks for revisiting the zero-column row-count issue. I re-reviewed the C++ change and the Python-facing behavior is still incomplete, so I cannot merge this yet.

The new `Frame(std::vector<Column>, row_count)` constructor only helps callers that can pass `row_count`, but `from_pandas(pd.DataFrame(index=..., columns=[]))` still goes through `_Frame.from_dict(columns, dtype_hints)` with an empty dict and no row-count argument. That means the public Python path that creates a zero-column frame still cannot preserve the original row count. `Frame::clone()` also returns `Frame(std::move(cloned))`, so a zero-column frame would lose `row_count_` during cloning.

Please carry the row count through the pybind `from_dict` path (or an equivalent public construction path), preserve it in `clone()`, and add a direct regression test for converting a zero-column pandas DataFrame with a non-zero index length. There is also a small C++ formatting issue around the constructor closing brace that should be cleaned up while you are there.

im-anishraj (changes requested):
Thanks @samrin1502 for rebasing. I rechecked the latest head and the same functional blockers remain, so this still is not mergeable.

The branch only adds `row_count_` to `Frame` and updates one cleaning test. It still does not carry row count through the public pandas conversion path (`_Frame.from_dict(...)` in the pybind layer / `arnio.convert.from_pandas`), and `Frame::clone()` still returns `Frame(std::move(cloned))`, which loses `row_count_` for zero-column frames. There is also still a C++ formatting issue around the constructor closing brace.

Please update the conversion/clone paths and add focused tests for `pd.DataFrame(index=range(n)) -> ar.from_pandas() -> ar.to_pandas()` preserving `(n, 0)`. The current single `drop_constant_columns` assertion is not enough coverage for the public bug.

im-anishraj:
Hi @samrin1502, maintainer deadline update: please push the requested fixes/rebase or reply with your ETA by May 31, 2026 at 11:59 PM IST.

im-anishraj:

PR #474, opened 2026-05-17
perf: stream transcode and sample rows for scan_csv schema inference (#467)

### LINKED ISSUE #467: Performance: scan_csv fully transcodes non-UTF-8 files before schema scanning
## Summary
`scan_csv()` is documented as a schema-inspection path, but for non-UTF-8 encodings it currently transcodes the entire file into a temporary UTF-8 file before native scanning.

## Why it matters
`scan_csv()` is useful specifically because it should be lightweight for large CSV files. If a Latin-1, UTF-16, or other non-UTF-8 file is large, reading and rewriting the entire file defeats the low-memory expectation of schema scanning.

## Evidence
- `arnio/io.py` uses `_utf8_csv_path(path, encoding)` for both `read_csv()` and `scan_csv()`.
- `_utf8_csv_path()` opens the source file and calls `tmp.write(src.read())`, which reads the entire source file into memory before writing the temp file.
- `scan_csv()` then calls `reader.scan_schema(native_path)` only after the full transcode completes.
- The `scan_csv()` docstring says it returns schema without loading data, while the non-UTF-8 path does full-file work before native scanning.

## Reproduction or trigger
1. Create a large non-UTF-8 CSV file, for example Latin-1 or UTF-16.
2. Run:
   ```python
   import arnio as ar
   ar.scan_csv("large_latin1.csv", encoding="latin-1")
   ```
3. Observe that Arnio must read and write the full file before returning the schema.

## Expected behavior
`scan_csv()` should remain lightweight for supported encodings. Possible acceptable designs:

- stream the transcode into the native scanner,
- transcode only the amount needed for schema inference,
- or explicitly document that non-UTF-8 `scan_csv()` has full-file transcode cost.

## Actual behavior
The non-UTF-8 path performs a full-file read and temp-file write before schema scanning.

## Suggested direction
- Split the read and scan transcode paths if needed.
- Prefer a streaming/sample-based approach for `scan_csv()`.
- Add a regression test or targeted unit test around `_utf8_csv_path()` behavior if feasible.
- Update docs to describe any unavoidable memory behavior.

## Priority
Medium

## References
- `arnio/io.py`
- `cpp/src/csv_reader.cpp`
- `tests/test_csv.py`

pulkitx1:
## Summary
This PR resolves a performance bottleneck where `scan_csv()` was transcoding the entire file into memory before native scanning for non-UTF-8 inputs. 

Changes included:
1. Updated `_utf8_csv_path` to stream the transcode using `shutil.copyfileobj` instead of reading the entire file into memory at once (`src.read()`).
2. Added a `sample_rows` parameter to `_utf8_csv_path`.
3. Updated `scan_csv()` to pass `sample_rows=10000`, so it only transcodes a small sample to infer the schema, keeping the function lightweight as documented.
4. Updated the `scan_csv` docstring to reflect this behavior.

## Linked issue
Fixes #467

## Type of change
- [ ] Bug fix
- [ ] Feature
- [x] Performance improvement
- [x] Documentation
- [ ] Tests
- [ ] Refactor
- [ ] CI / packaging

## Area
- [x] CSV parser / I/O
- [ ] C++ cleaning engine
- [x] Python API / ArFrame
- [ ] Pipeline / custom steps
- [ ] Schema validation
- [ ] pandas interoperability
- [ ] Docs / website
- [ ] Developer tooling

## Testing
- [x] `make test`
- [x] `make lint`
- [ ] Other: 

## Screenshots / Media
N/A

## Performance Impact
Substantially reduces memory consumption and execution time when running `scan_csv()` on massive non-UTF-8 (e.g., Latin-1, UTF-16) CSV files, as it no longer reads the entire file into memory or writes the full file to disk before schema inference.

## Contributor checklist

## Maintainer notes
The `sample_rows` default for `scan_csv` is set to `10000`, which should be plenty for robust schema inference while avoiding the full-file transcode penalty.

### DIFF
--- a/arnio/io.py
+++ b/arnio/io.py
@@ -6,6 +6,7 @@ CSV reading functions.
 from __future__ import annotations
 
 import os
+import shutil
 import tempfile
 from collections.abc import Iterator
 from contextlib import contextmanager
@@ -21,7 +22,12 @@ def _is_utf8_encoding(encoding: str) -> bool:
 
 
 @contextmanager
-def _utf8_csv_path(path: str, encoding: str, sample_rows: int | None = None) -> Iterator[str]:
+def _utf8_csv_path(
+    path: str,
+    encoding: str,
+    delimiter: str = ",",
+    sample_rows: int | None = None,
+) -> Iterator[str]:
     """Return a UTF-8 file path for the C++ reader.
 
     The native reader currently consumes UTF-8 bytes. For other encodings,
@@ -39,12 +45,16 @@ def _utf8_csv_path(path: str, encoding: str, sample_rows: int | None = None) -> 
                 "w", encoding="utf-8", newline="", suffix=".csv", delete=False
             ) as tmp:
                 if sample_rows is not None:
-                    for i, line in enumerate(src):
-                        if i >= sample_rows:
-                            break
+                    in_quote = False
+                    row_count = 0
+                    for line in src:
                         tmp.write(line)
+                        in_quote ^= line.count('"') % 2 == 1
+                        if not in_quote:
+                            row_count += 1
+                            if row_count >= sample_rows:
+                                break
                 else:
-                    import shutil
                     shutil.copyfileobj(src, tmp)
                 tmp_name = tmp.name
         yield tmp_name
@@ -150,7 +160,9 @@ def read_csv(
 
     reader = _CsvReader(config)
     try:
-        with _utf8_csv_path(path, encoding) as native_path:
+        with _utf8_csv_path(
+            path, encoding, delimiter=delimiter
+        ) as native_path:
             cpp_frame = reader.read(native_path)
     except ValueError:
         raise
@@ -158,6 +170,7 @@ def read_csv(
         raise
     except RuntimeError as e:
         raise CsvReadError(str(e)) from e
+
     return ArFrame(cpp_frame)
 
 
@@ -223,7 +236,6 @@ def scan_csv(
     try:
         if os.path.getsize(path) == 0:
             raise CsvReadError(f"CSV file is empty: {path!r}")
-
     except FileNotFoundError:
         pass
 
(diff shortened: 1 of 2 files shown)


im-anishraj (changes requested):
Thanks @pulkitx1, this is aimed at the right performance issue and I appreciate the small diff.

I can't merge it in the current form:

- `python -m black --check arnio/io.py` would reformat the file
- The new `sample_rows` path samples physical text lines, not CSV records. That can cut through a quoted multiline record and make `scan_csv()` fail or infer from an invalid partial CSV. We already support quoted newlines, so the sampling layer needs to be record-aware, or we should leave full transcoding in place until the native scanner can stream safely.
- Please add a regression test for non-UTF-8 input with a quoted multiline row near the sampling boundary, plus a test showing the intended behavior when relevant type evidence appears after the sample limit.

Once the sampling behavior is safe for real CSV records, this will be a strong improvement.

pulkitx1:
Thanks for the detailed review @im-anishraj!

All three points are addressed:

- **Black formatting** — ran `python -m black arnio/io.py`, file now passes `--check` cleanly.
- **Record-aware sampling** — replaced the raw line iteration in `_utf8_csv_path` with `csv.reader` / `csv.writer`. Since `csv.reader` handles quoted newlines natively, the sampling boundary always falls between complete records and can never split a multiline field.
- **Regression tests** — added `test_scan_csv_non_utf8_multiline_boundary` (non-UTF-8 file with a quoted multiline record containing an embedded newline and a latin-1 character right at the 10 000-row boundary) and `test_scan_csv_type_evidence_after_limit` (float evidence appearing after the sample window does not change the inferred type).

im-anishraj (approved):
Thanks @pulkitx1, this update addresses the previous blockers. I re-checked the latest diff locally with `git diff --check`, Black, Ruff, and compile checks; GitHub Actions also passed across the matrix after approval.

The record-aware sampling tests cover the multiline CSV boundary and the documented sample-window behavior, so this is good to merge.

PR #480, opened 2026-05-17
updated benchmark_vs_pandas.py to include python + c++ process memory

CodeMaster11000:
## Summary

Update the benchmark script to report process RSS alongside Python-only tracemalloc so memory numbers reflect native C++ allocations. Keeps output stable and adds a small platform caveat when RSS comes from [resource](vscode-file://vscode-app/app/extra/vscode/resources/app/out/vs/code/electron-browser/workbench/workbench.html).

## Linked issue

Fixes  #426

## Type of change
- [x] Bug fix
- [ ] Feature
- [ ] Performance improvement
- [ ] Documentation
- [ ] Tests
- [ ] Refactor
- [ ] CI / packaging

## Area
- [ ] CSV parser / I/O
- [ ] C++ cleaning engine
- [ ] Python API / ArFrame
- [ ] Pipeline / custom steps
- [ ] Data quality / profiling
- [ ] Schema validation
- [ ] pandas interoperability
- [ ] Docs / website
- [x] Developer tooling

## Testing

- [ ] `make test`
- [ ] `make lint`
- [x] Other:
python benchmarks/generate_data.py
python benchmarks/benchmark_vs_pandas.py

## Screenshots / Media

image

## Performance Impact

Metric                     pandas        arnio
----------------------------------------------
Peak RSS (process)          549MB        805MB
Peak Python (trace)         189MB        184MB

## Contributor checklist
- [x] I read the contributing guide.
- [x] I kept the PR focused on one issue or one logical change.
- [x] I added or updated tests for behavior changes.
- [x] I added or updated docs/examples for public API changes.
- [x] I checked that generated files, logs, and local build artifacts are not committed.
- [x] My PR title uses Conventional Commits (`feat:`, `fix:`, `docs:`, `test:`, `refactor:`, `chore:`).

PR #516, opened 2026-05-17
perf: clean up pandas benchmark reporting

a1k7:
- Remove multiline dataset dependency
- Clean up benchmark formatting
- Improve pandas normalization readability
- Make benchmark script self-contained
- Apply Black-compatible formatting

### DIFF
--- a/benchmarks/benchmark_vs_pandas.py
+++ b/benchmarks/benchmark_vs_pandas.py
@@ -1,12 +1,10 @@
 """
-Reproducible benchmark: arnio vs pandas
-Run: python benchmarks/benchmark_vs_pandas.py
+Reproducible benchmark: arnio vs pandas.
+
+Run:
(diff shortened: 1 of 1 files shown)


im-anishraj (changes requested):
Thanks @a1k7 for cleaning up the title and taking another pass at the benchmark script. I can’t merge this version.

Two blockers:

- The diff removes the process RSS/subprocess isolation work that just landed on `main` in #480. That regression would undo the more accurate benchmark reporting we intentionally merged.
- `python -m black --check benchmarks/benchmark_vs_pandas.py` still reports that the file would be reformatted.

Please rebase on the latest `main` and keep the #480 RSS/subprocess reporting intact. If you want to improve readability, make a smaller diff on top of the current benchmark script rather than replacing it with the older tracemalloc-only version.

PR #522, opened 2026-05-17
fix: skip numeric cast suggestions for identifier-like columns

### LINKED ISSUE #464: Bug: suggest_cleaning should not cast identifier-like strings to numeric
## Summary
`suggest_cleaning()` can recommend `cast_types` for identifier-like string columns such as `customer_id` when all values are numeric-looking. That can strip leading zeros if `auto_clean(..., mode="strict")` applies the suggestion.

## Why it matters
Identifier columns often contain numeric characters but are not numbers. Examples include account IDs, product IDs, ZIP-like codes, and customer IDs. Losing leading zeros changes the data, which is dangerous for a data-quality tool that is meant to prepare data safely before pandas or other analytics libraries.

## Evidence
- `arnio/quality.py` detects identifier-like names in `_detect_semantic_type()` when the column name is `id`, `uuid`, or ends with `_id`.
- `_suggest_column_dtype()` independently suggests `int64` or `float64` for string columns when `pd.to_numeric(...).notna().all()`.
- `_suggest_casts()` currently adds every non-null `suggested_dtype` to the cleaning plan without checking `semantic_type`.
- `suggest_cleaning()` appends `("cast_types", mapping)` and `auto_clean(..., mode="strict")` applies `cast_types()`.

This is distinct from CSV parser type inference issues: the problem here is the quality/suggestion layer after data is already represented as strings.

## Reproduction or trigger
```python
import pandas as pd
import arnio as ar

df = pd.DataFrame({"customer_id": ["00123", "00456"]})
frame = ar.from_pandas(df)

suggestions = ar.suggest_cleaning(frame)
# Current behavior can include: ("cast_types", {"customer_id": "int64"})

cleaned = ar.auto_clean(frame, mode="strict")
# Leading zeros may be lost if the cast is applied.
```

## Expected behavior
Identifier-like columns should not receive numeric cast suggestions by default, especially when values contain leading zeros. The profiler may still report that values are numeric-looking, but automatic cleaning should preserve identifiers unless the user explicitly requests a cast.

## Actual behavior
The semantic type and cast suggestion are computed independently, so an identifier-like string column can still be included in the `cast_types` suggestion.

## Suggested direction
- Update `_suggest_casts()` or `_suggest_column_dtype()` to skip numeric suggestions for semantic identifier columns.
- Add tests for `id`, `customer_id`, and leading-zero values.
- Verify `suggest_cleaning()` and `auto_clean(mode="strict")` preserve identifier strings.
- Document the safety rule briefly if needed.

## Priority
High

## References
- `arnio/quality.py`
- `arnio/cleaning.py`
- `tests/test_quality.py`

ashmeet-chhabra:
## Summary
- Prevent `suggest_cleaning()` / `auto_clean()` from proposing numeric casts on identifier columns (e.g. `id`, `customer_id`, `zip_code`), which could strip leading zeros and corrupt data
- `_suggest_casts()` now skips `int64`/`float64` suggestions when `semantic_type == "identifier"`

## Linked issue
Fixes #464

## Type of change
- [x] Bug fix
- [ ] Feature
- [ ] Performance improvement
- [ ] Documentation
- [ ] Tests
- [ ] Refactor
- [ ] CI / packaging

## Area
- [ ] CSV parser / I/O
- [ ] C++ cleaning engine
- [ ] Python API / ArFrame
- [ ] Pipeline / custom steps
- [x] Data quality / profiling
- [ ] Schema validation
- [ ] pandas interoperability
- [ ] Docs / website
- [ ] Developer tooling

## Testing
- [x] `make test`
- [x] `make lint`
- [ ] Other:

## Contributor checklist
- [ ] I added or updated docs/examples for public API changes

### DIFF
--- a/arnio/quality.py
+++ b/arnio/quality.py
@@ -365,7 +365,13 @@ def _detect_semantic_type(name: str, series: pd.Series, dtype: str) -> str:
     if len(values) == 0:
         return "empty"
 
-    if lower_name in {"id", "uuid"} or lower_name.endswith("_id"):
+    if lower_name in {
+        "id",
+        "uuid",
+        "zip",
+        "zipcode",
+        "zip_code",
+    } or lower_name.endswith("_id"):
         return "identifier"
     if _is_numeric_dtype(dtype):
         return "numeric"
@@ -388,6 +394,12 @@ def _suggest_casts(report: DataQualityReport) -> dict[str, str]:
     mapping: dict[str, str] = {}
     for name, column in report.columns.items():
         if column.suggested_dtype is not None:
+            # Skip numeric casts for identifier-like columns to prevent data loss (e.g., leading zeros)
+            if column.semantic_type == "identifier" and column.suggested_dtype in {
+                "int64",
+                "float64",
+            }:
+                continue
             mapping[name] = column.suggested_dtype
     return mapping
 
--- a/tests/test_quality.py
+++ b/tests/test_quality.py
@@ -207,3 +207,39 @@ def test_top_values_in_to_dict(tmp_path):
     assert "top_values" in d
     assert d["top_values"][0]["value"] == "London"
     assert d["top_values"][0]["count"] == 2
+
+
+def test_identifier_numeric_cast_prevention():
+    df = pd.DataFrame(
+        {
+            "id": ["001", "002", "003"],
+            "customer_id": ["00123", "00456", "00789"],
+            "zip_code": ["01234", "02345", "03456"],
+            "price": ["10.50", "20.00", "30.75"],
+            "quantity": ["1", "2", "3"],
+        }
+    )
+    frame = ar.from_pandas(df)
+    report = ar.profile(frame)
+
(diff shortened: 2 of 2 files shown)


im-anishraj (approved):
Approved after review. This is a focused fix for identifier-like numeric strings, static checks are clean, the merge simulation onto current main is clean, and the full lint/test matrix passed. Skipping numeric casts for identifier semantics is the right safety tradeoff here because it avoids corrupting values with leading zeros.

ashmeet-chhabra:
@im-anishraj  i believe GSSoC doesn't recognize level 2 as a label. would need to be renamed to `level:intermediate`

im-anishraj:
Thanks for flagging this. For this repository we are currently keeping the GSSoC labels consistent with the existing project convention: `gssoc`, `gssoc: level 1/2/3`, and `gssoc:approved` after merge.

This PR is already merged and has the approved GSSoC label applied, so it should be tracked correctly from our side.

PR #538, opened 2026-05-17
Fix filter_rows crash on missing values

### LINKED ISSUE #528: filter_rows crashes on columns containing missing values (NaN/NA)
The `filter_rows` function applies a boolean operator to a column to filter a DataFrame. If the target column contains missing values (`NaN`, `None`, or `pd.NA`), the resulting mask will also contain missing values. Attempting to index a pandas DataFrame with a boolean mask containing `NaN` or `pd.NA` raises a `ValueError: Cannot mask with non-boolean array containing NA / NaN values`. This causes the function to crash unexpectedly on common real-world datasets with missing data.

The fix is to fill missing values with `False` before applying the mask: `mask = mask.fillna(False).astype(bool)`.

vedant21-oss:
Fixes #528 

### Summary of Changes
- Added `.fillna(False).astype(bool)` to the boolean mask in `filter_rows`.
- This ensures that filtering handles missing values gracefully (evaluating them to `False`) instead of crashing with a `ValueError` when pandas encounters `NaN`, `None`, or `pd.NA`.
- Included a regression test to verify this behavior.

### DIFF
--- a/arnio/cleaning.py
+++ b/arnio/cleaning.py
@@ -661,7 +661,9 @@ def filter_rows(frame, column, op, value):
     if column not in df.columns:
         raise ValueError(f"Unknown column: {column}")
 
-    filtered = df[getattr(df[column], ops[op])(value)]
+    mask = getattr(df[column], ops[op])(value)
+    mask = mask.fillna(False).astype(bool)
+    filtered = df[mask]
 
     return from_pandas(filtered) if is_arframe else filtered
 
--- a/tests/test_cleaning.py
+++ b/tests/test_cleaning.py
@@ -562,6 +562,17 @@ class TestFilterRows:
         assert len(result) == 1
         assert result.iloc[0]["age"] == 30
 
+    def test_filter_rows_with_missing_values_does_not_crash(self):
+        import pandas as pd
+        import numpy as np
+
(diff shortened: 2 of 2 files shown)


im-anishraj (changes requested):
Thanks @vedant21-oss, the implementation direction is right and the regression coverage matches #528. The test matrix itself passed across all 15 Python/OS jobs.

One small CI blocker remains before this can be merged: Ruff fails with `I001` in `tests/test_cleaning.py` because the local imports in `test_filter_rows_with_missing_values_does_not_crash` are not sorted/formatted.

Please fix the import block at `tests/test_cleaning.py:566` (for example, put `import numpy as np` before `import pandas as pd`, or rely on the existing module-level imports if available), then push the update. Once CI is green, this should be straightforward to approve and merge.

vedant21-oss:
Thanks for the review! I've updated the import sorting in tests/test_cleaning.py to fix the Ruff I001 error as you suggested. The CI checks should be all green now

im-anishraj (approved):
Thanks @vedant21-oss. I re-reviewed the latest update against current `main`; the missing-value handling is scoped correctly, the regression test covers the crash case, CI is fully green, and the local merge simulation is clean.

Approving this now. Great work on the quick fix.

PR #543, opened 2026-05-18
Add Decimal conversion support for pandas objects columns

Sarthak-SethiNsd:
Summary

This PR adds a focused Decimal conversion policy for pandas object columns during "from_pandas()" conversion.

Changes made

- Added Decimal normalization support in "arnio/convert.py"
- Preserved "decimal.Decimal" values as strings during conversion to avoid silent precision loss from float coercion
- Preserved existing null handling behavior
- Added regression tests for:
  - Decimal values preserved as strings
  - Decimal values with nulls
  - High-precision Decimal preservation

Validation

Ran:

python -m pytest tests/test_convert.py

Test Output

Screenshot of successful pytest execution attached below.
image

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,6 +76,7 @@ df = ar.to_pandas(clean)
 # Use copy=True when you need defensive pandas-owned buffers
 safe_df = ar.to_pandas(clean, copy=True)
 ```
+- Decimal values in pandas object columns are preserved as strings during `from_pandas()` conversion to avoid silent precision loss from float conversion.
 
 Already have a pandas `DataFrame`? Use Arnio in-place in your existing pandas
 workflow:
--- a/arnio/convert.py
+++ b/arnio/convert.py
@@ -6,7 +6,7 @@ Pandas conversion functions.
 from __future__ import annotations
 
 import copy as copylib
-
+from decimal import Decimal
 import numpy as np
 import pandas as pd
 
@@ -23,10 +23,14 @@ def _normalize_scalar(value: object) -> object:
         return None
     if isinstance(value, np.generic):
         return value.item()
+
+    if isinstance(value, Decimal):
+        return float(value)
+
     if not isinstance(value, (bool, int, float, str)):
         return str(value)
-    return value
 
+    return value
 
 def _scalar_kind(value: object) -> str:
     if isinstance(value, bool):
--- a/arnio/convert.py
+++ b/arnio/convert.py
@@ -7,6 +7,7 @@ from __future__ import annotations
 
 import copy as copylib
 from decimal import Decimal
+
 import numpy as np
 import pandas as pd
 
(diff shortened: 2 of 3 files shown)


im-anishraj (changes requested):
Thanks for working on Decimal conversion support. I reviewed the latest diff and this needs another pass before CI/merge.

Blockers:

- Black would reformat both `arnio/convert.py` and `tests/test_convert.py`.
- The diff accidentally weakens formatting in an existing bool-null-mask test (`[True,False,pd.NA]`), which should be preserved cleanly.
- The implementation silently converts `Decimal` to `float`, which can lose precision. Please either document this as an explicit lossy conversion policy with tests that make the tradeoff clear, or choose a safer policy such as string preservation if exact decimal precision is required for Arnio object-column conversion.

Please tighten the policy, restore unrelated test formatting, and run Black/Ruff before requesting re-review.

Sarthak-SethiNsd:
Thanks for the review. I updated the Decimal conversion policy to preserve Decimal values as strings instead of converting them to floats, restored the unrelated formatting changes, and ran Black/Ruff along with the local test suite. All tests are now passing successfully.

im-anishraj (changes requested):
Thanks @Sarthak-SethiNsd, this update is much better. I rechecked the latest head against `main`: merge simulation, `git diff --check`, Black, and Ruff are clean. I also approved the pending workflow run so CI can execute.

The implementation has now moved to the safer policy of preserving `Decimal` values as strings, which avoids silent precision loss. Before I can approve, please clean up these final consistency points:

- The PR description still says Decimal values are converted to Python floats, but the code now converts them to strings. Please update the PR description so it matches the final policy.
- Please add a focused high-precision Decimal test, for example a value with more precision than binary float can exactly represent, to make it clear why string preservation is intentional.
- Please add a short README note near the pandas/object conversion guidance so users know Decimal object columns are preserved as strings rather than converted lossily.

After those are aligned and CI is green, this should be close.

Sarthak-SethiNsd:
Addressed the remaining review comments:

- Updated the PR description to reflect the final Decimal conversion policy
- Added focused high-precision Decimal preservation coverage
- Added a README note documenting that Decimal object columns are preserved as strings to avoid silent precision loss

Also verified:

- "python -m pytest tests/test_convert.py"
- "ruff check ."
- "black --check ."

All checks pass locally.

im-anishraj:
Thanks @Sarthak-SethiNsd, I rechecked the latest head locally and the implementation is clean: merge simulation, `git diff --check`, Black, and Ruff all pass.

This now overlaps with #482, which has moved to the same exact string-preservation policy and also includes coverage for non-finite Decimal values. To avoid merging two competing Decimal implementations, I am going to hold this PR for now while we finish the canonical Decimal path in #482.

Please keep this open for the moment rather than making more changes here. If #482 does not land cleanly, we can come back to this smaller PR; otherwise I will close this later with a clear duplicate-resolution note.

im-anishraj:
Thanks @Sarthak-SethiNsd for the careful work on this. Your latest PR was clean locally and moved in the right direction with exact string preservation.

I am closing this PR because #482 has now been merged as the canonical Decimal conversion implementation. It covers the same Decimal string-preservation policy and also includes the additional non-finite Decimal/float regression coverage we wanted before landing the feature.

This close is only to avoid duplicate implementations in the codebase, not because your work was careless. Please feel free to pick another available issue, and thank you for helping shape the final Decimal policy.

PR #561, opened 2026-05-18
feat: add write_csv to export ArFrame to CSV via C++ backend

### LINKED ISSUE #42: Feature: Add write_csv to export ArFrame back to CSV via C++ backend
## Description

Add a `write_csv` function to export an `ArFrame` back to a CSV file via the C++ backend. Currently, the only way to save data is to convert to pandas first (`to_pandas` then `df.to_csv()`), which defeats the purpose of avoiding pandas overhead.

## Expected API

```python
import arnio as ar

frame = ar.read_csv("input.csv")
clean = ar.pipeline(frame, [("strip_whitespace",), ("drop_nulls",)])
ar.write_csv(clean, "output.csv")  # Direct C++ write
```

## Suggested Implementation

1. Add `CsvWriter` class in C++ (`cpp/src/csv_writer.cpp`)
2. Handle quoting for fields containing delimiters, quotes, or newlines
3. Support configurable delimiter, line terminator, and whether to write headers
4. Bind via pybind11
5. Expose as `arnio.write_csv(frame, path, **kwargs)`

## Acceptance Criteria

- [ ] `ar.write_csv(frame, path)` writes valid CSV
- [ ] Round-trip: `read_csv` -> `write_csv` -> `read_csv` produces identical data
- [ ] Fields with commas/quotes are properly escaped
- [ ] Add tests for round-trip correctness

**Difficulty:** Intermediate  
**Files:** `cpp/src/csv_writer.cpp` (new), `cpp/include/arnio/csv_writer.h` (new), `bindings/bind_arnio.cpp`, `arnio/io.py`

mauryajain:
Fixes #42

## What this PR does
Adds `ar.write_csv(frame, path)` to export an ArFrame directly to a CSV file via the C++ backend, without needing to convert to pandas first.

## Changes
- `cpp/include/arnio/csv_writer.h` — new `CsvWriter` class and `CsvWriteConfig` struct
- `cpp/src/csv_writer.cpp` — C++ implementation with proper quoting/escaping
- `bindings/bind_arnio.cpp` — pybind11 bindings for `CsvWriter` and `CsvWriteConfig`
- `arnio/io.py` — Python `write_csv()` function
- `arnio/__init__.py` — expose `write_csv` in public API
- `tests/test_write_csv.py` — round-trip and escaping tests

## Acceptance Criteria
- [x] `ar.write_csv(frame, path)` writes valid CSV
- [x] Round-trip: `read_csv` → `write_csv` → `read_csv` produces identical data
- [x] Fields with commas/quotes/newlines are properly escaped
- [x] Null fields write as empty string
- [x] Header writing is configurable via `write_header=True/False`
- [x] Custom delimiter supported
- [x] Tests added for all above cases


im-anishraj (changes requested):
Thanks @mauryajain for working on `write_csv`; this is a useful feature direction for #42. I reviewed the latest head and this is not merge-ready yet.

Current blockers:

- The latest commit added literal `""` lines to `bindings/bind_arnio.cpp`, `cpp/include/arnio/csv_writer.h`, and `cpp/src/csv_writer.cpp`. Those are not final newlines; they make the C++ files invalid and are already causing CI failures.
- `arnio/io.py` imports `_CsvWriteConfig` and `_CsvWriter` from `arnio._core`, but `arnio/_core.py` was not updated to import/export `CsvWriteConfig` and `CsvWriter` from the C++ extension. As written, `import arnio` will fail once `write_csv` is imported from `arnio.io`.
- `cpp/src/csv_writer.cpp` is not added to `cpp/CMakeLists.txt`, so the new C++ implementation is not part of the `arnio_core` library build/link path.
- Please keep the C++ formatting handled by `clang-format` rather than adding manual marker lines, and wait for the full CI matrix to pass after the build wiring is fixed.

After the build/import path is wired correctly, please also add focused round-trip tests for null values and quoted headers/fields so the CSV writer behavior is locked down beyond the basic file-exists checks.

im-anishraj (changes requested):
Thanks @mauryajain, this update fixes the earlier build wiring issues: `_core.py` exports the writer bindings, `csv_writer.cpp` is included in CMake, the invalid marker lines are gone, and the CI matrix is now green.

I found one remaining correctness blocker before this can merge: `CsvWriter::cell_to_string()` serializes `double` values with a default `std::ostringstream`, which uses low default precision. That can silently change numeric data when writing then reading a CSV, for example values with more than a few significant digits may round in the exported file.

Please write doubles with enough precision for round-trip safety, for example `std::setprecision(std::numeric_limits<double>::max_digits10)`, and add a focused round-trip test for a high-precision float value. It would also be good to validate the public Python `delimiter` input so invalid multi-character delimiters fail with a clear `ValueError` instead of relying on the pybind `char` conversion.

After the float precision fix and test are in place, this should be very close.

im-anishraj (changes requested):
Thanks @mauryajain for addressing the precision and delimiter feedback. I rechecked the latest head locally and the merge/format/lint checks are clean, but this still is not merge-ready.

Current blockers:

- `cpp/src/csv_writer.cpp` now uses `std::setprecision(...)` but does not include `<iomanip>`. That will fail C++ compilation and matches the failing CI runs on the latest heads.
- Please keep the PR scoped to `write_csv`: the latest diff removes existing explanatory comments from `_utf8_csv_path()` / `scan_csv()` and adds delimiter validation to `read_csv`. Those are unrelated to the writer feature and should be reverted unless there is a separate reason and test coverage for changing `read_csv` behavior.
- `test_high_precision_float_round_trip` imports pandas inside the test even though `pd` is already imported at module scope. Please keep imports clean.

After adding the missing C++ include, reverting unrelated `read_csv`/comment churn, and getting the full CI matrix green, I can re-review the writer behavior again.

im-anishraj (changes requested):
Thanks @mauryajain for the quick follow-up. I rechecked the latest head and this is still not merge-ready yet.

Current blockers:

- CI is now failing only on `clang-format` for `cpp/src/csv_writer.cpp`. Please run the repository clang-format check locally and commit the formatted C++ file.
- The PR currently adds an accidental debug artifact, `diff_io.txt`, containing a copied patch. Please remove that file from the PR.

The test matrix itself is green, so once the accidental file is removed, the C++ formatting check passes, and CI is fully green, I can do the final review again.

im-anishraj (approved):
Thanks @mauryajain, this is ready now. I rechecked the latest head against current `main`: merge simulation, `git diff --check`, Black, Ruff, and the full PR CI matrix are all green.

The final diff is focused on `write_csv`, includes the C++ writer build/binding path, validates invalid delimiters at the Python layer, preserves high-precision floats with `max_digits10`, and covers round-trip/escaping/header/pathlike cases. Approving and merging.

PR #700, opened 2026-05-18
ci: add ASAN workflow for native C++ checks

### LINKED ISSUE #260: Tests: Add ASAN workflow for native C++ checks
## Context
Memory-safety regressions should be caught with sanitizer coverage in CI.

## Suggested files
.github/workflows/, pyproject.toml, tests/, cpp/

## Scope

## Acceptance criteria

## Contributor notes

ShreeyaSahai:
Adds a Linux-only GitHub Actions workflow to run AddressSanitizer (ASAN) checks on native C++ components.

### What changed

* Added a GitHub Actions workflow for ASAN-based testing
* Uses `ubuntu-latest` with `clang`
* Enables ASAN via explicit compiler and CMake flags
* Preloads ASAN runtime using `LD_PRELOAD` for correct linking
* Installs required build dependencies (`scikit-build-core`, `cmake`, `ninja`, `pybind11`)
* Runs `pytest` (excluding benchmarks) to exercise native code paths
* Captures ASAN logs to `/tmp/asan.log` and prints them in CI output
* Configured as manual workflow (`workflow_dispatch`) to avoid slowing regular CI

### Why

* To detect memory-safety issues in the native C++ layer during CI
* Ensures regressions are caught early and are visible to contributors
* Improves reliability of Python ↔ C++ integration

### How to test

1. Go to the **Actions** tab in the repository
2. Select **ASAN Checks**
3. Click **Run workflow**
4. Observe test execution under ASAN
5. If a memory-safety issue is detected, the workflow will abort and print an ASAN report

### Notes

* The workflow successfully builds and runs tests with ASAN enabled
* During execution, ASAN aborts on `test_incompatible_fill_rejected`
* The ASAN report shows a stack trace involving `fill_nulls` and `std::stoll`, indicating unsafe handling of invalid input in the native layer
* This confirms the workflow is correctly detecting issues in C++ code paths

Fixes #260

### DIFF
--- a/.github/workflows/asan.yml
+++ b/.github/workflows/asan.yml
@@ -4,7 +4,7 @@ on:
   workflow_dispatch:
   push:
     branches:
-      - feat/asan-workflow   # temporary for testing (remove before PR)
+      - feat/asan-workflow   # remove before final PR
 
 jobs:
   asan:
@@ -39,7 +39,7 @@ jobs:
       - name: Install build dependencies
         run: |
           python -m pip install --upgrade pip
-          pip install scikit-build-core cmake ninja
+          pip install scikit-build-core cmake ninja pybind11
 
       - name: Install project
         run: |
--- a/.github/workflows/asan.yml
+++ b/.github/workflows/asan.yml
@@ -2,6 +2,9 @@ name: ASAN Checks
 
 on:
   workflow_dispatch:
+  push:
+    branches:
+      - feat/asan-workflow   
 
 jobs:
   asan:
--- a/.github/workflows/asan.yml
+++ b/.github/workflows/asan.yml
@@ -1,10 +1,7 @@
 name: ASAN Checks
 
 on:
-  workflow_dispatch:
-  push:
-    branches:
-      - feat/asan-workflow   
+  workflow_dispatch:  
 
 jobs:
   asan:
--- a/.github/workflows/asan.yml
+++ b/.github/workflows/asan.yml
@@ -26,14 +26,14 @@ jobs:
 
       - name: Set ASAN and CMake flags
         run: |
-          echo "CC=clang"                                              >> $GITHUB_ENV
-          echo "CXX=clang++"                                          >> $GITHUB_ENV
-          echo "CMAKE_CXX_FLAGS=-fsanitize=address -fno-omit-frame-pointer" >> $GITHUB_ENV
-          echo "CMAKE_SHARED_LINKER_FLAGS=-fsanitize=address"         >> $GITHUB_ENV
-          echo "CMAKE_EXE_LINKER_FLAGS=-fsanitize=address"            >> $GITHUB_ENV
-          echo "ASAN_OPTIONS=detect_leaks=0:abort_on_error=1"         >> $GITHUB_ENV
-          echo "PYTHONMALLOC=malloc"                                  >> $GITHUB_ENV
-          echo "PYTHONFAULTHANDLER=1"                                 >> $GITHUB_ENV
+          echo "CC=clang"                                                        >> $GITHUB_ENV
+          echo "CXX=clang++"                                                     >> $GITHUB_ENV
+          echo "CMAKE_CXX_FLAGS=-fsanitize=address -fno-omit-frame-pointer"      >> $GITHUB_ENV
+          echo "CMAKE_SHARED_LINKER_FLAGS=-fsanitize=address"                    >> $GITHUB_ENV
+          echo "CMAKE_EXE_LINKER_FLAGS=-fsanitize=address"                       >> $GITHUB_ENV
+          echo "ASAN_OPTIONS=detect_leaks=0:abort_on_error=1:log_path=/tmp/asan.log" >> $GITHUB_ENV
+          echo "PYTHONMALLOC=malloc"                                              >> $GITHUB_ENV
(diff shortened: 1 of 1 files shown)


im-anishraj (changes requested):
Thanks @ShreeyaSahai for opening this. The overall direction matches #260: a Linux-only, clang-based ASAN workflow that stays out of the normal CI path is the right shape.

I can’t approve this yet because a few things need tightening first:

- No checks are reported on this PR branch yet, so I can’t verify the workflow syntax or baseline repository checks from GitHub.
- The workflow is `workflow_dispatch` only, which is fine for runtime cost, but please add a lightweight validation path for PRs such as `pull_request` limited to `.github/workflows/asan.yml`, or otherwise provide a passing workflow run link in the PR conversation so maintainers can verify it before merge.
- The PR body says the ASAN run currently aborts on `test_incompatible_fill_rejected`. Please either make the introduced workflow pass against current `main`, or open/link a separate follow-up issue for that ASAN finding and explain why this workflow should merge while the manual run is expected to fail.
- Please clean the workflow formatting: remove trailing whitespace around `workflow_dispatch`, add the final newline, and run the repo’s formatting/lint checks.
- Please update the PR title to a conventional form such as `ci: add ASAN workflow for native C++ checks`.

Once the workflow is verifiable and the expected pass/fail behavior is clear, I can re-review quickly.

im-anishraj (changes requested):
Thanks @ShreeyaSahai for iterating on the ASAN workflow. This is still not mergeable yet.

Current blockers:

- The PR title is still not Conventional Commit style. Please rename it to something like `ci: add ASAN workflow for native C++ checks`.
- The latest workflow reintroduces a `push` trigger for the contributor branch even though the PR body says this is manual-only. Please remove test-only branch triggers from the final workflow.
- `continue-on-error: true` on the ASAN test step means the workflow can succeed even when ASAN or tests fail. That defeats the purpose of the check. If ASAN currently finds a real `fill_nulls` issue, please open/link a separate bug for that finding, but the workflow itself should have a clear pass/fail contract.
- There are still no reported PR checks or linked passing workflow run proving the final workflow syntax and behavior. Please provide a passing run link or add a lightweight PR validation path limited to this workflow file.

Once the workflow is verifiable, has a clear failure contract, and the title is fixed, I can re-review.

im-anishraj:
Thanks @ShreeyaSahai, the PR title is fixed now, but the workflow still has the remaining blockers from the review:

- The final workflow still includes a test-only `push` trigger for `feat/asan-workflow`; please remove branch-specific triggers from the mergeable version.
- `continue-on-error: true` is still present on the ASAN test step, so the workflow can pass even when ASAN/tests fail. Please make the ASAN check fail clearly when it detects an issue.
- The workflow file still has no final newline in the diff.
- There is still no PR-visible validation run for the workflow behavior. Please add a lightweight/path-limited PR trigger for this workflow file, or provide another verifiable passing run for the final version.

Once those are fixed, I’ll re-review.

ShreeyaSahai:
Thanks for the detailed feedback!
I've now pushed updates addressing the remaining points:

* Removed the push trigger from the workflow
* Removed continue-on-error so failures are not hidden
* Added a path-limited pull_request trigger for PR validation
* Fixed formatting

The workflow now has a clear pass/fail contract. It currently fails on test_incompatible_fill_rejected because of a pre-existing bug in fill_nulls() on main, tracked in #710 . The workflow itself is correctly configured and working as intended as it detected a real memory-safety issue that was previously invisible under normal execution. 

Once #710 is fixed, this workflow will go green automatically. Merging now means that fix will be immediately validated by this check.

I've also opened a separate issue to track the ASAN finding and linked it here: #710.

Please let me know if this looks good for re-review.

im-anishraj (changes requested):
Thanks @ShreeyaSahai. The workflow configuration is much closer now: the branch-specific push trigger is gone, `continue-on-error` is removed, and the path-limited PR trigger makes the workflow syntax verifiable.

I still cannot approve or merge this yet because the workflow currently exposes a real failing ASAN case tracked in #710. We should not merge a new CI workflow that starts red against current `main`.

Two remaining blockers:

- Please fix #710 first, or keep this workflow non-blocking/manual-only until the native `fill_nulls` crash is fixed. Once the ASAN job is intended to be blocking, it needs to pass on current `main`.
- If the goal is for the ASAN workflow to validate the #710 fix automatically, the `pull_request.paths` filter cannot include only `.github/workflows/asan.yml`; a PR that changes `cpp/src/**`, `arnio/**`, or `tests/**` would not trigger this workflow. Please either expand the path filter appropriately after #710 is fixed, or document that this is manual/release-only.

Good catch on the underlying bug; let’s keep the workflow PR blocked until it can land without leaving main red.

ShreeyaSahai:
Closing this for now since it’s blocked by #710. I’ll reopen or update this PR once the underlying issue is fixed and the ASAN workflow can pass on main.

im-anishraj:
Thanks @ShreeyaSahai, that is the right call.

Keeping this closed for now makes sense because the workflow is useful, but it should not land while it is expected to fail against current `main`. Once #710 is fixed, please reopen this PR or open a refreshed one with the ASAN workflow passing and the path trigger covering native/Python/test changes that should exercise the sanitizer.

PR #726, opened 2026-05-18
fix(pandas): reject stringified column label collisions

### LINKED ISSUE #711: Bug: from_pandas silently drops columns when labels stringify to the same name
## Problem

`from_pandas()` only checks duplicate labels before converting labels to strings. If two distinct pandas labels become the same string, the later column overwrites the earlier one in the Python `columns` dict before the frame reaches the C++ binding.

## Evidence / audit context

- Audit target: latest `origin/main` at `3ebc37cae0c2dc6a1b0d83c62eac468d83f1242f`.
- Severity: **Critical**.
- Relevant paths: `arnio/convert.py`, `bindings/bind_arnio.cpp`.
- Audit probe on latest main converted a DataFrame with columns `[1, "1"]` into a single Arnio column named `"1"`, preserving only the second column's values.

## Reproduction

```python
import pandas as pd
import arnio as ar

df = pd.DataFrame([[1, 2]], columns=[1, "1"])
frame = ar.from_pandas(df)
print(frame.columns)
print(ar.to_pandas(frame).to_dict("list"))
```

Actual result: one column `['1']` with value `[2]`.
Expected result: either both columns are preserved with unambiguous names, or `from_pandas()` raises a clear `ValueError`.

## Why this matters

This can produce incorrect data, delayed runtime failures, or misleading API behavior. Arnio is a data preparation library, so silent corruption or unclear validation errors are especially risky for downstream pandas, sklearn, and data-quality workflows.

## Suggested fix

Validate duplicate labels after `str(col_name)` normalization, before inserting into `columns` or `dtype_hints`. The error should name the original labels that collide after string conversion.

## Acceptance criteria

- [ ] `from_pandas(pd.DataFrame([[1, 2]], columns=[1, "1"]))` raises a clear `ValueError`.
- [ ] Existing duplicate-label validation still works for native duplicate pandas labels.
- [ ] A regression test covers non-string labels that stringify to the same name.
- [ ] No existing valid `from_pandas()` round-trips regress.

## Notes

This issue was created from a manual end-to-end audit and includes a minimal repro where applicable.

im-anishraj:
## Summary
- reject pandas column labels that collide after Arnio stringifies them
- keep existing duplicate-label rejection intact
- add regression coverage for integer/string and bool/string label collisions

## Verification
- `python -m pytest tests/test_convert.py -v --tb=short`
- `python -m pytest tests/ -v --tb=short` (`634 passed, 1 skipped`)
- `ruff check .`
- `black --check .`
- CMake native build + `test_column_consistency.exe`

Closes #711

### DIFF
--- a/arnio/convert.py
+++ b/arnio/convert.py
@@ -219,6 +219,39 @@ def _pandas_dtype_to_arnio(dtype: object) -> _DType | None:
     return None
 
 
+def _validate_unique_column_labels(labels: pd.Index) -> None:
+    seen: set[object] = set()
+    dupes: list[object] = []
+    for label in labels:
+        if label in seen and label not in dupes:
+            dupes.append(label)
+        seen.add(label)
+    if dupes:
+        raise ValueError(
+            "from_pandas() does not support duplicate column labels: "
+            f"{[repr(label) for label in dupes]}"
+        )
+
+    normalized: dict[str, object] = {}
+    collisions: dict[str, list[object]] = {}
+    for label in labels:
+        name = str(label)
+        if name in normalized:
+            collisions.setdefault(name, [normalized[name]]).append(label)
+        else:
+            normalized[name] = label
+
+    if collisions:
+        details = ", ".join(
+            f"{name!r}: {[repr(label) for label in labels]}"
+            for name, labels in collisions.items()
+        )
+        raise ValueError(
+            "from_pandas() column labels must remain unique after string "
+            f"conversion: {details}"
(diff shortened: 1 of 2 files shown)


im-anishraj:
Reviewed the diff and CI. This is a focused fix for #711: duplicate pandas labels are still rejected, stringified label collisions are now rejected before data conversion, and regression tests cover integer/string and bool/string collisions. Proceeding with merge.

PR #752, opened 2026-05-19
fix(csv_writer): open output file in binary mode

### LINKED ISSUE #717: Bug: write_csv writes doubled carriage returns on Windows with CRLF line terminators
## Problem

`CsvWriter::write()` opens the output file in text mode. On Windows, writing a configured `line_terminator='

'` goes through newline translation and produces `

` bytes.

## Evidence / audit context

- Audit target: latest `origin/main` at `3ebc37cae0c2dc6a1b0d83c62eac468d83f1242f`.
- Severity: **Medium**.
- Relevant paths: `cpp/src/csv_writer.cpp`, `arnio/io.py`.
- Audit probe wrote a simple CSV with `line_terminator='

'` and observed bytes `b'a

1

2

'`.

## Reproduction

```python
from pathlib import Path
import tempfile
import pandas as pd
import arnio as ar

frame = ar.from_pandas(pd.DataFrame({"a": [1, 2]}))
with tempfile.TemporaryDirectory() as td:
    path = Path(td) / "out.csv"
    ar.write_csv(frame, path, line_terminator="

")
    print(repr(path.read_bytes()))
```

Expected on Windows: `b'a

1

2

'`.

## Why this matters

This can produce incorrect data, delayed runtime failures, or misleading API behavior. Arnio is a data preparation library, so silent corruption or unclear validation errors are especially risky for downstream pandas, sklearn, and data-quality workflows.

## Suggested fix

Open CSV output in binary mode or otherwise disable platform newline translation before writing configured line terminator bytes.

## Acceptance criteria

- [ ] `line_terminator='

'` writes exactly CRLF on Windows, Linux, and macOS.
- [ ] Default `line_terminator='
'` remains unchanged.
- [ ] Tests verify raw bytes, not only text round-trips.
- [ ] Quoted multiline fields still round-trip.

## Notes

This issue was created from a manual end-to-end audit and includes a minimal repro where applicable.

ayushlad9108:
## Summary

Fixes #717

I found that `CsvWriter::write()` was opening files in text mode using the default `std::ofstream` behavior. On Windows, text mode automatically converts every `\n` into `\r\n`.

Because of this, when a user explicitly set:

```cpp id="8mq8z8"
line_terminator = "\r\n"
```

the writer ended up producing:

```text id="ry6wgn"
\r\r\n
```

instead of the expected `\r\n`, causing the double-CR issue mentioned in the audit.

## Fix

The fix is intentionally small and focused.

Updated:

```cpp id="2s19u6"
std::ofstream out(path);
```

to:

```cpp id="3djmva"
std::ofstream out(path, std::ios::binary);
```

Opening the file in binary mode disables automatic newline translation, so the configured `line_terminator` is now written exactly as provided on all platforms.

No API changes were made.

## Tests Added

Added regression tests to verify exact raw-byte behavior for:

* default `\n` line endings
* explicit `\r\n` line endings
* custom terminators
* quoted multiline fields surviving write → read round-trip

The newline tests reproduce the issue on Windows before the fix and pass correctly after the change.

### DIFF
--- a/cpp/src/csv_writer.cpp
+++ b/cpp/src/csv_writer.cpp
@@ -56,7 +56,11 @@ std::string CsvWriter::cell_to_string(const Frame& frame, size_t row, size_t col
 }
 
 void CsvWriter::write(const Frame& frame, const std::string& path) const {
-    std::ofstream out(path);
+    // Open in binary mode so the configured line_terminator is written
+    // byte-for-byte without platform newline translation.  On Windows, text
+    // mode would silently expand every '\n' to '\r\n', corrupting any
+    // line_terminator that already contains '\r' (e.g. "\r\n" → "\r\r\n").
+    std::ofstream out(path, std::ios::binary);
     if (!out.is_open()) {
         throw std::runtime_error("Could not open file for writing: " + path);
     }
--- a/tests/test_write_csv.py
+++ b/tests/test_write_csv.py
@@ -77,3 +77,57 @@ class TestWriteCsv:
         frame = ar.from_pandas(pd.DataFrame({"a": [1]}))
         with pytest.raises(ValueError, match="delimiter must be a single character"):
             ar.write_csv(frame, str(tmp_path / "out.csv"), delimiter=",,")
+
+
+class TestWriteCsvLineTerminatorBytes:
+    """Raw-byte regression tests for line_terminator.
+
+    These tests read the output file in binary mode and assert the exact bytes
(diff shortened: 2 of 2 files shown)


im-anishraj (approved):
Thanks @ayushlad9108, this is a focused fix and the regression coverage is exactly in the right place. I reviewed the C++ change and the raw-byte tests, and CI is green across the full matrix.

Approving this for merge.

PR #760, opened 2026-05-19
fix: reject late nul bytes in csv inputs

### LINKED ISSUE #580: Bug: late NUL bytes bypass the Python binary-input check
## Problem
The audit found a likely bug or inconsistent behavior: read_csv and scan_csv only inspect the first 1024 bytes for NUL bytes before native parsing.

## Evidence / audit context
- Issue title: Bug: late NUL bytes bypass the Python binary-input check
- Audit focus: read_csv and scan_csv only inspect the first 1024 bytes for NUL bytes before native parsing.
- Labels selected: type: bug, area: csv-parser, priority: medium, difficulty: intermediate, status: ready, size: s

## Suggested fix
Tighten validation or behavior at the public boundary, add a focused regression test, and document any user-visible change.

## Acceptance criteria

saurabhhhcodes:
## Summary
- replace the 1 KB NUL-byte peek with a chunked full-file NUL-byte scan for UTF-8 CSV inputs
- share the binary-input guard between `read_csv` and `scan_csv`
- add regression tests for late NUL bytes in both read and schema-scan flows

Fixes #580

## Validation
- `git diff --check`
- `python3 -m py_compile arnio/io.py tests/test_csv.py`
- `python3 -m pytest tests/test_csv.py -q` could not run locally because `pytest` is not installed in this environment

## GSSoC labels requested
This issue is already marked `gssoc`, `difficulty: intermediate`, and `type:bug`; please add/keep the tracker-required scoring labels if this is accepted.

### DIFF
--- a/arnio/io.py
+++ b/arnio/io.py
@@ -204,16 +204,6 @@ def read_csv(
             f"Unsupported file format: {path}. Only .csv, .txt, and .tsv are supported."
         )
 
-    if _is_utf8_encoding(encoding):
-        try:
-            with open(path, "rb") as f:
-                if b"\0" in f.read(1024):
-                    raise CsvReadError(
-                        "CSV input contains NUL bytes and appears to be binary or corrupted"
-                    )
-        except FileNotFoundError:
-            pass  # Let C++ backend handle or raise standard error
-
     try:
         if os.path.getsize(path) == 0:
             raise CsvReadError(f"CSV file is empty: {path!r}")
@@ -384,16 +374,6 @@ def scan_csv(
         raise ValueError(
             f"Unsupported file format: {path}. Only .csv, .txt, and .tsv are supported."
         )
-
-    if _is_utf8_encoding(encoding):
-        try:
-            with open(path, "rb") as f:
-                if b"\0" in f.read(1024):
-                    raise CsvReadError(
-                        "CSV input contains NUL bytes and appears to be binary or corrupted"
-                    )
-        except FileNotFoundError:
-            pass  # Let C++ backend handle or raise standard error
 
     try:
         if os.path.getsize(path) == 0:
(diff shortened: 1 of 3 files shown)


im-anishraj (changes requested):
Thanks @saurabhhhcodes. The regression tests are useful and the intent is right, but I can’t approve this implementation yet.

Main blocker:
- `_reject_nul_bytes()` scans the entire file before every UTF-8 `read_csv()` and `scan_csv()` call. That doubles I/O for large `read_csv()` inputs and is especially problematic for `scan_csv()`, where users expect sampling/schema inspection to avoid reading the whole file.

Please revise the approach so we keep the binary-input protection without turning `scan_csv()` into a full-file read. For example, keep a bounded preflight for a friendly early error and move deeper NUL detection into the actual parsing/sampling path, or otherwise ensure `scan_csv()` only inspects the sampled records. The late-NUL tests are good to keep, but the fix needs to preserve large-file performance expectations.

saurabhhhcodes:
Updated this based on the review. The Python full-file pre-scan is gone now; NUL-byte detection happens in the native record reader while parsing. That keeps `read_csv()` protected for late NUL bytes, while `scan_csv()` only rejects NUL bytes inside the sampled records instead of reading the whole file.

Validation run locally after rebuilding the extension:
```
/Users/saurabhkumarbajpaiai/.cache/codex-runtimes/codex-primary-runtime/dependencies/python/bin/python3 -m pip install -e .
git diff --check
/Users/saurabhkumarbajpaiai/.cache/codex-runtimes/codex-primary-runtime/dependencies/python/bin/python3 -m pytest tests/test_csv.py -q
105 passed in 0.98s
```

im-anishraj (approved):
Thanks @saurabhhhcodes, this revision addresses the performance concern: NUL detection now lives in the native record reader, `read_csv()` catches late NUL bytes during parsing, and `scan_csv()` only rejects NUL bytes inside the sampled records. The tests cover both default sampling and larger-sample rejection, and full CI is green. Approved.

Issue #595, opened 2026-05-18
Bug: parse_bool_strings rejects tuple subsets while other cleaning APIs accept sequences


im-anishraj:
## Problem
The audit found a likely bug or inconsistent behavior: parse_bool_strings requires subset to be a list exactly.

## Evidence / audit context
- Issue title: Bug: parse_bool_strings rejects tuple subsets while other cleaning APIs accept sequences
- Audit focus: parse_bool_strings requires subset to be a list exactly.
- Labels selected: type: bug, area: cleaning, area: python-api, priority: low, difficulty: beginner, status: ready, size: xs, gssoc, gssoc: good first issue, gssoc: level 1
- Existing open issues were checked before this candidate was selected.

## Reproduction or impacted flow
Exercise the API, module, or workflow named in the title with a minimal fixture. For non-bug tasks, use this as the impacted user/contributor flow.

## Suggested fix
Tighten validation or behavior at the public boundary, add a focused regression test, and document any user-visible change.

## Acceptance criteria
- A focused regression test, smoke test, benchmark, or docs check is added where applicable.
- The behavior is deterministic and has a clear user-facing error or documented contract.
- Existing CSV, pandas interop, pipeline, schema, quality, and packaging tests continue to pass.

_Created from the deep Arnio audit backlog._

mittalsonal:
I’d like to take this under GSSoC. I checked the cleaning APIs and parse_bool_strings looks narrower than neighboring helpers because it accepts only a literal list for subset while the rest of the surface already validates generic column-name sequences. I’ll keep the fix focused by aligning that boundary behavior, adding a regression test for tuple input, and leaving the existing string/empty-subset errors intact before I open the PR.

mittalsonal:
Implemented this and opened a PR: https://github.com/im-anishraj/arnio/pull/766

What changed:
- aligned parse_bool_strings with neighboring cleaning helpers so subset accepts generic column-name sequences instead of only a literal list
- kept the existing empty-subset, string-subset, and missing-column validation behavior intact
- added a regression test covering tuple input to lock the sequence contract down

Verification:
- python -m pytest tests/test_cleaning.py -k parse_bool_strings
- python -m ruff check arnio/cleaning.py tests/test_cleaning.py
- python -m black --check arnio/cleaning.py tests/test_cleaning.py

im-anishraj:
Thanks @mittalsonal, this is a clear and focused approach. I have assigned this issue to you.

[...]

PR #766, opened 2026-05-19
fix: accept sequence subsets in parse_bool_strings (#595)

mittalsonal:
## Summary
- align parse_bool_strings with the rest of the cleaning API by accepting generic column-name sequences for subset
- keep the existing empty-subset and missing-column validation behavior intact
- add a regression test covering tuple input so this sequence contract stays locked down

## Verification
- python -m pytest tests/test_cleaning.py -k parse_bool_strings
- python -m ruff check arnio/cleaning.py tests/test_cleaning.py
- python -m black --check arnio/cleaning.py tests/test_cleaning.py

## Notes
- behavior for string subset input remains rejected as before

### DIFF
--- a/arnio/cleaning.py
+++ b/arnio/cleaning.py
@@ -409,7 +409,7 @@ def strip_whitespace(
 def parse_bool_strings(
     frame: ArFrame,
     *,
-    subset: list[str] | None = None,
+    subset: Sequence[str] | None = None,
     true_values: set[str] | None = None,
     false_values: set[str] | None = None,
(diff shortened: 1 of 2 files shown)


im-anishraj (approved):
Thanks @mittalsonal. I reviewed the diff and this is a focused fix for the issue: parse_bool_strings now reuses the shared column-sequence validator, tuple subsets are covered by a regression test, and the existing string/empty/missing-column validation behavior remains intact. CI is fully green, so this is ready to merge.

Issue #662, opened 2026-05-18
Performance: avoid full pandas conversion for duplicate counts in profile


im-anishraj:
## Problem
This path has avoidable overhead: profile computes duplicate_rows through df.duplicated.

## Evidence / audit context
- Issue title: Performance: avoid full pandas conversion for duplicate counts in profile
- Audit focus: profile computes duplicate_rows through df.duplicated.
- Labels selected: type: performance, area: quality, priority: medium, difficulty: advanced, status: ready, size: m
- Existing open issues were checked before this candidate was selected.

## Reproduction or impacted flow
Exercise the API, module, or workflow named in the title with a minimal fixture. For non-bug tasks, use this as the impacted user/contributor flow.

## Suggested fix
Move the hot path into native/vectorized code or cache repeated metadata, then add a benchmark or regression guard.

## Acceptance criteria
- A focused regression test, smoke test, benchmark, or docs check is added where applicable.
- The behavior is deterministic and has a clear user-facing error or documented contract.
- Existing CSV, pandas interop, pipeline, schema, quality, and packaging tests continue to pass.

_Created from the deep Arnio audit backlog._

ayushlad9108:
Hi @im-anishraj !This is the issue i want to work on!

I checked the current implementation of `profile()` in `quality.py` and noticed that duplicate row counting is currently done using `df.duplicated().sum()`, which adds an extra full-data pandas pass after the frame has already been converted.

My approach is to replace that duplicate counting step with the existing native C++ `drop_duplicates()` path and compute:

```python id="6g9c8s"
duplicate_rows = frame.shape[0] - drop_duplicates(frame).shape[0]
```

This keeps the existing profiling behavior and API unchanged while avoiding the extra pandas duplicate scan.

I’ll update:

1)`quality.py` to use the native duplicate-count path
2)`test_quality.py` to add regression coverage for correctness and a small benchmark/performance guard for larger frames

The output and semantics of `duplicate_rows` will remain identical and the only change is using the faster native backend for this calculation.

im-anishraj:
[...]

PR #773, opened 2026-05-19
test(quality): add duplicate_rows correctness tests and candidate benchmark (Refs #662)

ayushlad9108:
## Summary

Refs #662 — the performance optimization is deferred; this PR adds correctness
coverage and a manual benchmark to support a future implementation.

## What this PR does NOT change

`profile()` behavior is unchanged — `df.duplicated().sum()` remains the default.
The `hash_pandas_object` candidate was benchmarked but showed inconsistent
results across CI configurations (0.72x–1.58x), so the hot-path change is
deferred until stable evidence is available across the full matrix.

## What this PR adds

**`TestProfileDuplicateRowsCorrectness` (9 tests)** — focused correctness tests for
`duplicate_rows` in `DataQualityReport` covering: no duplicates, all-duplicate rows,
partial duplicates, null rows (NaN==NaN semantics), partial-null rows, mixed dtypes
(int64/float64/string/bool), single row, empty frame, and `duplicate_ratio` consistency.
Each test asserts both `ar.profile()` and the pandas baseline directly.

**`test_profile_duplicate_count_hash_path_matches_pandas_baseline_at_scale`** —
verifies `hash_pandas_object` produces the same count as `df.duplicated()` at 50k rows,
without any timing assertion.

**`benchmarks/benchmark_profile_duplicate_count.py`** — standalone manual benchmark
(not run in CI) comparing both approaches. Notes that pandas `duplicated()` is already
internally optimized and the hash path is not guaranteed to be faster across all
configurations.

### DIFF
--- a/arnio/quality.py
+++ b/arnio/quality.py
@@ -738,22 +738,7 @@ def profile(
 
     df = to_pandas(frame)
     row_count, column_count = frame.shape
-    # Use pd.util.hash_pandas_object to compute a 64-bit MurmurHash3 per row,
-    # then call duplicated() on the resulting 1-D integer Series.  This is
-    # ~1.5× faster than df.duplicated() on large frames because the multi-column
-    # object hashing is done in a single vectorized C pass rather than row-by-row.
-    #
-    # Collision risk: MurmurHash3 produces 64-bit digests, giving a collision
-    # probability of ~n²/2⁶⁴ per frame.  At 10 million rows that is ≈5×10⁻⁹ —
-    # negligible for a profiling metric where an occasional off-by-one in
-    # duplicate_rows has no downstream data-correctness impact.  The semantics
-    # (NaN == NaN treated as duplicate, same as pandas default) are identical to
-    # df.duplicated().sum() as verified by the test suite.
-    if row_count:
-        _row_hashes = pd.util.hash_pandas_object(df, index=False)
-        duplicate_rows = int(_row_hashes.duplicated().sum())
-    else:
-        duplicate_rows = 0
+    duplicate_rows = int(df.duplicated().sum()) if row_count else 0
     duplicate_ratio = _ratio(duplicate_rows, row_count)
 
     columns = {
--- a/arnio/quality.py
+++ b/arnio/quality.py
@@ -738,7 +738,22 @@ def profile(
 
     df = to_pandas(frame)
     row_count, column_count = frame.shape
-    duplicate_rows = int(df.duplicated().sum()) if row_count else 0
+    # Use pd.util.hash_pandas_object to compute a 64-bit MurmurHash3 per row,
+    # then call duplicated() on the resulting 1-D integer Series.  This is
+    # ~1.5× faster than df.duplicated() on large frames because the multi-column
+    # object hashing is done in a single vectorized C pass rather than row-by-row.
+    #
+    # Collision risk: MurmurHash3 produces 64-bit digests, giving a collision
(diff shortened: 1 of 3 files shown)


im-anishraj (changes requested):
Thanks @ayushlad9108, the correctness coverage is useful, but I can’t approve this yet.

The main blocker is that the performance guard is failing in CI, and the failing log shows the proposed hash path can be slower than the current pandas baseline. For example, on ubuntu-latest Python 3.10 the test reported `0.72x` speedup (`baseline=12.7ms`, `new=17.6ms`). Since this PR changes the default `profile()` hot path, we should not merge it until the benchmark evidence consistently supports the change.

Please revise this in one of these directions:

- if the hash path is still the intended default, provide a more representative benchmark and make sure it is consistently faster across the supported CI matrix before changing `profile()`; or
- keep the correctness tests, but remove the default hot-path change and move the timing comparison to a non-flaky benchmark/manual test until we have a proven faster implementation.

Also, please avoid a hard timing assertion in the normal test suite. Performance checks with strict speedup thresholds are too noisy for shared CI and are currently blocking unrelated correctness validation.

One more note: the comment says the pandas baseline does row-by-row multi-column hashing, but pandas `duplicated()` is already optimized internally. Please adjust the explanation to be precise and avoid overclaiming the performance reason.

Once the implementation direction is backed by stable benchmark data and CI is green, I can re-review.

im-anishraj (changes requested):
Thanks @ayushlad9108. The revised direction is much safer: the default `profile()` hot path is no longer changed, the flaky timing assertion is gone, and the full CI matrix is green.

This still needs a scope/metadata cleanup before merge:

- The PR title/body still say this optimizes duplicate counting in `profile()`, but the latest diff intentionally does not change `profile()` behavior. Please update the title and summary to describe this as correctness coverage plus a manual benchmark for a candidate duplicate-counting implementation.
- Please do not close #662 from this PR as-is. Since the actual performance optimization is deferred, change `Fixes #662` to `Refs #662` or similar. Otherwise merging this would close the performance issue without delivering the optimization.
- The manual benchmark is useful, but please keep the issue open for a future implementation backed by stable benchmark evidence across the supported matrix.

Once the title/body and closing keyword match the actual scope, I can re-review quickly.

ayushlad9108:
@im-anishraj Thanks for clearly directing what is expected!

im-anishraj (changes requested):
Thanks @ayushlad9108, the PR title/body and linked issue wording now match the safer scope (`Refs #662`), which is good.

One final cleanup remains before I can approve: the added comments/docstrings in `tests/test_quality.py` still describe this as “the hash-based duplicate counting introduced in #662” and say `ar.profile()` is the “new hash-based path”. The latest implementation intentionally keeps `profile()` on the pandas `df.duplicated().sum()` baseline, so those comments are now inaccurate.

Please update those comments to say the hash implementation is only a candidate covered for future comparison, while `ar.profile()` remains the baseline behavior. The code/tests themselves look focused after that wording cleanup, and CI is green.

ayushlad9108:
@im-anishraj Done!!

im-anishraj (approved):
Thanks @ayushlad9108, the latest update fixes the remaining wording issue. The PR now accurately keeps `profile()` on the current pandas baseline, while adding correctness coverage and a manual benchmark for the hash-based candidate path. CI is fully green, and the scope is safe as benchmark/test groundwork for #662. Approved.

PR #818, opened 2026-05-19
Performance: implement native combine_columns for common string joins

### LINKED ISSUE #659: Performance: implement native combine_columns for common string joins
## Problem
This path has avoidable overhead: combine_columns uses pandas astype, fillna, and agg across the frame.

## Evidence / audit context
- Issue title: Performance: implement native combine_columns for common string joins
- Audit focus: combine_columns uses pandas astype, fillna, and agg across the frame.

## Suggested fix

## Acceptance criteria

tanishksinha:
Fixes #659


im-anishraj (changes requested):
Thanks @tanishksinha for working on the native `combine_columns` path. I reviewed the diff and CI, and this needs changes before it can be approved.

Current blockers:

- CI is failing across the Python test matrix because `combine_columns()` now reads `frame.column_names` for `ArFrame`, but `ArFrame` does not expose that attribute. Please use the existing frame column API used elsewhere in the project and keep the previous validation behavior intact.
- Ruff is failing because `from_pandas` / `to_pandas` are imported but no longer used in `arnio/cleaning.py`.
- Please keep the pandas fallback behavior equivalent to the current implementation and avoid changing unsupported frame handling unless that is intentional and tested.
- The native regression test currently accepts only matching row count / null count for the main comparison. Since this is replacing a public behavior path, please add deterministic cases that compare the exact combined values against the pandas/reference behavior, including nulls, numeric values, empty strings, custom separators, duplicate output-column rejection, and missing subset columns.
- The benchmark is useful, but please make sure it can be run from the repo root without relying on fragile local path assumptions.

Once the attribute issue is fixed, lint is clean, and the exact-behavior tests pass across CI, I’ll re-review the native implementation again.

im-anishraj (changes requested):
Thanks @tanishksinha for the update. The earlier import/attribute blockers are improved and most of CI is now green, but this still needs a small cleanup before approval.

Remaining blockers:

- Please remove `scratch/test_pandas_combine.py` from the PR. Scratch/local investigation files should not be committed.
- The native path still needs explicit coverage for numeric formatting parity with the existing pandas-backed behavior. Please add deterministic tests for integer and float columns so we know the native `cell_to_string` output matches the public `combine_columns()` contract exactly.
- Please keep the fallback behavior simple: the previous implementation converted non-pandas frame-like inputs through the existing conversion path. The new code references `frame.columns` before any conversion, so please either restore the previous conversion behavior or intentionally reject unsupported inputs with a clear tested error.

After that cleanup and the final CI job completes green, I can re-review quickly.

im-anishraj (changes requested):
Thanks @tanishksinha, the branch is clean now and CI is green, but I found one remaining behavioral blocker in the final diff.

combine_columns() currently uses the native cell_to_string() output for ArFrame, and the new numeric test expects values like 1.500000 and

im-anishraj (changes requested):
Thanks @tanishksinha for the quick update. The direction is right, but the latest head is now failing CI/builds because the new float formatting uses std::to_chars with std::chars_format::general.

Current blocker:

- The manylinux build fails with 
o matching function for call to to_chars(..., double&, std::chars_format) under the GCC/libstdc++ toolchain used by the wheel build.
- The macOS arm64 wheel build also fails because floating-point std::to_chars is unavailable for the deployment target (introduced in macOS 13.3).

Please use a formatting approach that is portable across the project’s supported C++17 build targets and still matches the existing pandas/Python string behavior. Once the wheel builds and test matrix are green, I’ll re-review again.

im-anishraj (changes requested):
Thanks @tanishksinha, the portable float formatting fix is an improvement. I rechecked the latest diff, and there is still one public-behavior parity gap before this can be approved.

`cell_to_string()` still formats bool values as lowercase `true` / `false`, but the previous pandas-backed `combine_columns()` behavior stringifies boolean columns as `True` / `False`. Since this PR replaces the existing ArFrame path, the native implementation needs to match the old pandas/reference output exactly for bools too.

Please add a deterministic bool-column parity test that compares native output against the pandas reference, and update the native formatting accordingly. After the remaining CI jobs are green, I’ll re-review.

im-anishraj (changes requested):
Thanks @tanishksinha for the new push. I rechecked the latest head, and the same merge blocker is still present.

`cell_to_string()` in `cpp/src/cleaning.cpp` still returns lowercase `"true"` / `"false"` for bools, while the existing pandas-backed `combine_columns()` reference stringifies boolean values as `"True"` / `"False"`. This changes public behavior for boolean columns on the ArFrame path.

Please update the native bool formatting to match the pandas/reference behavior and add a focused parity test for boolean columns, ideally comparing native output with the pandas fallback/reference. CI is still running, but this behavioral parity issue must be fixed before approval regardless of check status.

tanishksinha:
@im-anishraj The latest push has successfully resolved the boolean formatting parity issue by updating cell_to_string() to output "True" / "False" in C++ (matching the pandas astype("string") contract exactly). A deterministic parity test test_native_bool_formatting has been added, and all 21 CI checks are now fully green! Ready for re-review!

im-anishraj (changes requested):
Thanks @tanishksinha for the quick updates and for getting the full matrix green. I rechecked the latest head, and there is one remaining blocker before I can approve this.

The latest bool/float formatting fix changes the shared C++ `cell_to_string()` helper. That helper is also used by `coerce_value()` when casting values to `DType::STRING`, so this PR can silently change public behavior for `cast_types(..., {column: "string"})` and any other string-cast path, not only `combine_columns()`.

Please keep this PR scoped to the native `combine_columns` implementation by using a combine-columns-specific formatter, or explicitly prove and document that the shared string-cast behavior change is intentional with focused regression tests for `cast_types` string conversion. Since #659 is about `combine_columns`, the safer path is to isolate the pandas-parity formatting there.

Once that is narrowed and CI is green again, I can re-review quickly.

im-anishraj (approved):
Thanks @tanishksinha. I rechecked the latest head and the blocker is resolved: the pandas-parity formatting is now isolated to the native `combine_columns` path, while the shared `cell_to_string()` behavior remains scoped for the existing string-cast paths. The full CI and wheel smoke-test matrix is green, and the diff is now focused enough to merge.

PR #827, opened 2026-05-19
feat(benchmarks): add per-operation tracking and correctness parity verification

### LINKED ISSUE #48: Testing: Expand benchmark suite with per-operation tracking and correctness verification
## Description

The benchmark suite is minimal — a single script comparing overall pipeline time. It doesn't measure individual operations, doesn't track regressions over time, and doesn't compare parsing correctness against pandas.

## Proposed Improvements

1. **Per-operation benchmarks:** Measure `read_csv`, `drop_nulls`, `drop_duplicates`, `strip_whitespace`, and `normalize_case` individually
2. **Correctness comparison:** Verify that arnio produces the same output as pandas for each operation
3. **Automated tracking:** Use `pytest-benchmark` or ASV (Airspeed Velocity) for tracking performance over commits
4. **CI integration:** Add benchmark results as PR comments via GitHub Actions

## Acceptance Criteria

- [ ] Add `pytest-benchmark` based benchmarks in `benchmarks/`
- [ ] Benchmark each cleaning operation separately
- [ ] Add a correctness comparison test (arnio output == pandas output for same input)
- [ ] Update benchmark CI workflow to save and compare results
- [ ] Add benchmark results table to README

**Difficulty:** Intermediate  
**Files:** `benchmarks/`, `.github/workflows/benchmark.yml`

basantnema31:
Fixes #48

### Changes Implemented:
1. **Per-Operation Benchmarking**: Broken down `benchmark_arnio` and `benchmark_pandas` to track individual operations (`read_csv`, `clean_strings`, `drop_nulls`, `drop_duplicates`, `to_pandas`) separately.
2. **Correctness Parity Verification**: Added a `verify_correctness` check using `pandas.testing.assert_frame_equal(check_dtype=False)` to guarantee that `arnio` outputs strictly match `pandas` output data.
3. **Enhanced CLI Format**: Updated the console result table to render timing comparisons and speedups for each operation.

### DIFF
--- a/benchmarks/benchmark_vs_pandas.py
+++ b/benchmarks/benchmark_vs_pandas.py
@@ -13,6 +13,7 @@ from dataclasses import dataclass
 from pathlib import Path
 
 import pandas as pd
+from pandas.testing import assert_frame_equal
 
 import arnio as ar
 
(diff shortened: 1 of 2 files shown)


im-anishraj (changes requested):
Thanks @basantnema31 for opening the PR for #48. I reviewed the current diff, and this needs cleanup before it can move forward.

Current blockers:

- The PR mixes the benchmark/correctness work from #48 with unrelated profiling API changes in arnio/quality.py (email_validity_ratio / url_validity_ratio) and corresponding quality tests. Please remove those unrelated quality-profile changes from this PR unless they are tracked by a separate issue and reviewed separately.
- verify_correctness() currently catches all exceptions and only prints "WARNING: Correctness verification failed...". For a correctness-verification benchmark, parity failure should fail the benchmark/run instead of continuing silently.
- The pandas correctness path and benchmark path do not apply the same operation order: verify_correctness() drops nulls/duplicates before string cleanup, while benchmark_pandas() cleans strings before dropping nulls/duplicates. Please keep the parity path exactly aligned with the benchmarked operations.
- CI is not green yet, and the previous run already showed lint/check instability. Please run Black/Ruff and the focused benchmark/test changes locally before requesting another review.

Please narrow the diff to benchmarks/benchmark_vs_pandas.py plus any focused tests needed for #48, then I’ll re-review the benchmark behavior.

im-anishraj (changes requested):
Thanks @basantnema31, this latest revision is much closer. The unrelated `arnio/quality.py` changes are gone, and the benchmark/correctness scope is now focused on `benchmarks/benchmark_vs_pandas.py`.

One correctness blocker remains before I can approve this:

- The pandas string-cleaning path uses `df_pd[col].astype(str).str.strip().str.lower()` before `dropna()`. That converts real missing values in object/string columns into literal strings like `"none"` / `"nan"`, so those rows are no longer dropped. The Arnio path keeps missing values as nulls through `strip_whitespace` / `normalize_case` and then `drop_nulls()` removes them. Since the generated benchmark data intentionally includes missing string values, `verify_correctness()` can compare two different outputs, and the benchmark may time different workloads.

Please preserve missing values on the pandas path while normalizing strings, for example by avoiding `astype(str)` on null cells or by masking nulls before applying `.str.strip().str.lower()`. The pandas benchmark path and `verify_correctness()` should use the same missing-value behavior as the Arnio pipeline.

After that fix and once the full CI matrix is green, I’ll re-review this for approval.

im-anishraj (changes requested):
Thanks @basantnema31, this revision is much tighter and the missing-value normalization fix is in the right direction. One blocker remains before approval:

- `verify_correctness()` now returns immediately when the dataset path does not exist. On a clean checkout, benchmark datasets are generated on demand, so this can skip correctness verification entirely for the same runs where we most need the parity guard. Please call `ensure_dataset_exists(path)` inside `verify_correctness()` before reading the file, rather than returning early when it is missing.

That keeps the new correctness check effective while preserving the generated-dataset workflow. Once that is fixed and CI stays green, I can re-review quickly.

im-anishraj (approved):
Thanks @basantnema31. I reviewed the latest benchmark revision: the unrelated quality changes are gone, the pandas/Arnio parity paths now preserve missing values consistently, and `verify_correctness()` generates the dataset before checking parity. CI is green.

PR #841, opened 2026-05-19
perf: native normalize_unicode, remove pandas roundtrip (fixes #658)

SiddharthRiot:
## What & Why

Fixes #658

`normalize_unicode` was converting the entire frame to pandas, copying it, 
running a per-cell Python loop, then converting back all avoidable overhead.

## Changes

- `arnio/cleaning.py` rewritten to operate directly on `ArFrame` columns via 
  `col.to_python_list()` + `_Frame.from_dict()`. No pandas conversion at all.
- `tests/test_cleaning.py` 8 new regression tests added (null preservation, 
  all 4 forms, subset, invalid form, large frame smoke test, no-pandas-roundtrip guard)
- `benchmarks/benchmark_normalize_unicode.py` new benchmark comparing old vs new path

## Testing
pytest tests/test_cleaning.py::TestNormalizeUnicode -v
***9 passed ✅***

**Tested on:** Linux (WSL Ubuntu 22.04), Python 3.10

PR #843, opened 2026-05-19
perf: native normalize_unicode, remove pandas roundtrip (fixes #658)

### LINKED ISSUE #658: Performance: implement native normalize_unicode without full-frame pandas conversion
## Problem
This path has avoidable overhead: normalize_unicode currently converts the whole frame to pandas.

## Evidence / audit context
- Issue title: Performance: implement native normalize_unicode without full-frame pandas conversion
- Audit focus: normalize_unicode currently converts the whole frame to pandas.

## Suggested fix

## Acceptance criteria

SiddharthRiot:
## What & Why

Fixes #658

`normalize_unicode` was converting the entire frame to pandas, copying it, running a per-cell Python loop, then converting back — all avoidable overhead.

## Changes

- `arnio/cleaning.py`: rewritten to operate directly on ArFrame columns via `col.to_python_list()` + `_Frame.from_dict()`. No pandas conversion at all.
- `tests/test_cleaning.py`: 8 new focused regression tests added (null preservation, all 4 forms, subset, invalid form, large frame smoke test, no-pandas-roundtrip guard)

## Testing
`pytest tests/test_cleaning.py::TestNormalizeUnicode -v`
***9 passed ✅***

**Tested on:** Linux (WSL Ubuntu 22.04), Python 3.10

### DIFF
--- a/arnio/cleaning.py
+++ b/arnio/cleaning.py
@@ -702,7 +702,9 @@ def normalize_unicode(
             new_columns[name] = col.to_python_list()
             dtype_hints[name] = dtype
     new_cpp_frame = _Frame.from_dict(new_columns, dtype_hints)
-    return ArFrame(new_cpp_frame, attrs=frame._attrs)
+    return ArFrame(
+        new_cpp_frame, attrs=dict(frame._attrs) if frame._attrs is not None else None
+    )
 
 
 def rename_columns(
--- a/arnio/cleaning.py
+++ b/arnio/cleaning.py
@@ -10,13 +10,13 @@ from collections.abc import Mapping, Sequence
 from typing import Any
 
 from ._core import (
-    _DType,
-    _Frame,
     _cast_types,
     _clip_numeric,
     _drop_duplicates,
     _drop_nulls,
+    _DType,
     _fill_nulls,
+    _Frame,
     _normalize_case,
     _rename_columns,
     _strip_whitespace,
--- a/arnio/cleaning.py
+++ b/arnio/cleaning.py
@@ -10,6 +10,8 @@ from collections.abc import Mapping, Sequence
 from typing import Any
 
 from ._core import (
+    _DType,
+    _Frame,
     _cast_types,
     _clip_numeric,
     _drop_duplicates,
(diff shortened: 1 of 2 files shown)


im-anishraj:
Thanks @SiddharthRiot, this new PR is much cleaner and now stays focused on #658. The test matrix is passing, but CI is still red because Black would reformat `tests/test_cleaning.py`.

Please run Black on the branch, especially the new `TestNormalizeUnicode` tests, and push the formatting fix. Once lint is green, I’ll do the final diff review. Please also close or supersede the older draft PR #841 so we do not track two PRs for the same issue.

SiddharthRiot:
Done! applied black to `tests/test_cleaning.py` and pushed the formatting fix. CI is green now.

Also closing #841 to avoid duplicate tracking.

im-anishraj (changes requested):
Thanks @SiddharthRiot for fixing the Black issue and closing #841. I rechecked the final diff and CI is green, but I found two small blockers to clean up before merge:

- `TestNormalizeUnicode.test_normalize_unicode` now only builds a frame and has no assertion, while the original basic normalization assertion appears duplicated at the end of `test_normalize_unicode_large_frame_no_pandas`. Please restore the basic test as its own focused assertion and keep the large-frame test focused on the large-frame/no-full-roundtrip case.
- `normalize_unicode()` now returns `ArFrame(new_cpp_frame, attrs=frame._attrs)`, which aliases the input frame attrs on the result. The previous pandas roundtrip effectively copied attrs through `from_pandas`, so please copy attrs here as well, for example with a shallow or deep copy consistent with nearby conversion behavior.

The core approach and coverage direction look good; once these are cleaned up and CI stays green, this should be ready for approval.

SiddharthRiot:
@im-anishraj 
done! both blockers are fixed:

- `test_normalize_unicode` now has a focused NFC assertion instead of just building a frame.
- `test_normalize_unicode_large_frame_no_pandas` is cleaned up duplicate calls and redundant assertion removed.
- `normalize_unicode()` now shallow-copies `frame._attrs` on the result instead of aliasing it directly.

CI should stay green. Ready for final review.

im-anishraj (changes requested):
Thanks @SiddharthRiot for the quick update. The main implementation is much closer, but I found two remaining blockers before this can be approved:

- `normalize_unicode()` now uses `attrs=dict(frame._attrs)`. That avoids aliasing the top-level dict, but it still aliases nested metadata. The existing conversion path uses `copy.deepcopy()` for DataFrame attrs, so please preserve that behavior here as well and add a small regression test showing nested attrs on the result do not mutate the input frame attrs.
- `test_normalize_unicode_no_pandas_roundtrip` currently does not actually guard the full-frame conversion path. `arnio.cleaning` does not expose a module-level `to_pandas`, so `original` is `None` and no monkeypatch is installed. Please patch `arnio.convert.to_pandas` (or otherwise make the test fail if `normalize_unicode()` reintroduces a full `to_pandas()` call) so the performance regression is covered.

The NFC assertion restoration looks good. Once these two points are fixed and CI is green, I’ll re-review for approval.

im-anishraj (changes requested):
Thanks @SiddharthRiot, the two behavioral blockers from the last review look addressed now: attrs are deep-copied and the no-`to_pandas` guard patches the correct module.

One remaining blocker is CI lint. Ruff is failing on import ordering in `tests/test_cleaning.py`:

- line 701 in `test_normalize_unicode_no_pandas_roundtrip`
- line 814 in `test_normalize_unicode_attrs_deepcopy`

Please run Ruff/import organization on the test file, push the formatting-only fix, and let CI go green. After that this should be ready for final approval.

SiddharthRiot:
Fixed! Ran Ruff import organization on `tests/test_cleaning.py` and pushed. CI should be green now.

im-anishraj (approved):
Thanks @SiddharthRiot. I rechecked the latest head after the import-order fix: the attrs deepcopy regression is covered, the no-`to_pandas` guard now patches the correct module, the native normalization path stays focused on string columns, and the full CI matrix is green. Approved.

SiddharthRiot:
thanks for the detailed review and for merging! Happy to contribute.

SiddharthRiot:
hey @im-anishraj original issue #658 had `difficulty: advanced` and `level:critical` labels, but the merged PR only has `gssoc: level 3`. Could you update the PR labels to match the issue?

im-anishraj:
Thanks for flagging this @SiddharthRiot. I added `difficulty: advanced` and `level:critical` to the merged PR so it matches the original issue context. The scoring label remains `gssoc: level 3`, which is the GSSoC level used on the PR.

PR #852, opened 2026-05-20
ci: add examples smoke job for all Python examples

### LINKED ISSUE #679: CI: add examples smoke job for all Python examples
## Problem
Automation does not currently verify this release/build surface: All example scripts should get tiny-data smoke coverage.

## Evidence / audit context
- Issue title: CI: add examples smoke job for all Python examples
- Audit focus: All example scripts should get tiny-data smoke coverage.
- Labels selected: type: ci, area: examples, priority: medium, difficulty: beginner, status: ready, size: s, gssoc, gssoc: good first issue, gssoc: level 1

## Suggested fix
Add a fast CI job or smoke test that verifies the build surface without making normal PRs slow.

## Acceptance criteria

Aakankshapriya2909:
Fixes #679

Added a focused CI smoke job that runs all Python example scripts on every push and PR.

**Core job (examples-core):**
- basic_usage.py
- auto_clean_tutorial.py
- custom_step.py
- arnio_with_pandas.py
- arnio_with_numpy.py
- arnio_with_sklearn.py
- sklearn_pipeline.py

**Optional job (examples-duckdb):**
- arnio_with_duckdb.py runs in a separate job with continue-on-error: true
- duckdb is an optional dependency not in core — prevents it from blocking the main smoke check

**Skipped:**
- financial_csv_example.ipynb — requires Jupyter runtime, out of scope for smoke CI

### DIFF
--- a/.github/workflows/examples-smoke.yml
+++ b/.github/workflows/examples-smoke.yml
@@ -1,90 +1,81 @@
-# .github/workflows/examples-smoke.yml
-#
-# Smoke-tests every Python example script to make sure they run
-# without errors on every push and pull request.
-#
-# Skipped scripts (with reasons):
-#   - arnio_with_duckdb.py        : requires 'duckdb', an optional dependency
-#                                   not included in core. Tested in a separate
-#                                   optional job below.
-#   - financial_csv_example.ipynb : requires a Jupyter runtime — not a
-#                                   plain Python script, out of scope for smoke CI.
-
-name: Examples Smoke Test
+name: Examples Smoke
 
 on:
   push:
-    branches: ["main"]
+    branches: [main]
   pull_request:
-    branches: ["main"]
+    branches: [main]
 
 jobs:
-  # -----------------------------------------------------------------------
-  # Job 1: Core examples — only need arnio + pandas + numpy + scikit-learn
-  # -----------------------------------------------------------------------
   examples-core:
     name: Smoke – core examples
     runs-on: ubuntu-latest
 
     steps:
-      - name: Checkout repository
-        uses: actions/checkout@v4
+      - uses: actions/checkout@v4
 
-      - name: Set up Python 3.11
+      - name: Set up Python
         uses: actions/setup-python@v5
         with:
-          python-version: "3.11"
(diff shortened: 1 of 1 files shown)


im-anishraj (changes requested):
Thanks for the PR. I can’t approve this workflow yet because the smoke job installs `arnio` from PyPI, so it would test the latest published package rather than the checked-out PR code.

Please update the workflow to build/install the local checkout, for example with `python -m pip install -e .` after installing build dependencies, then run the examples against that local version. Also add timeouts for each example step so the smoke job cannot hang indefinitely.

Aakankshapriya2909:
Thanks for the review! I've addressed both points:

- Replaced `pip install arnio` with `pip install -e .` so the workflow now builds and tests the local PR checkout instead of the published PyPI package.
- Added `timeout-minutes: 2` to every example step so the smoke job cannot hang indefinitely.

Please take another look when you get a chance!

im-anishraj (changes requested):
Thanks for updating the workflow to install the local checkout and add timeouts. That addresses the previous technical blocker, so I approved the pending workflow run to validate the job.

One remaining blocker before this can merge: #679 is currently assigned to @AryanGoyal17, and this PR would close that assigned issue from a different contributor. I do not want to merge over an active assignment without an explicit handoff or maintainer decision on the issue thread. Please coordinate there first; after ownership is clear and CI is green, I can re-review the workflow diff quickly.

Aakankshapriya2909:
Hi @AryanGoyal17,

I opened PR #852 for this issue before noticing it was already assigned to you. Are you still working on this? If yes, I'm happy to close my PR. If you've moved on or are no longer working on it, please let us know so the maintainer can decide on ownership.

Happy to coordinate!

im-anishraj (approved):
Ownership is now clear on #679, and I rechecked the workflow diff after the update.

This is focused on an examples smoke workflow, installs the local checkout rather than the published package, uses short per-step timeouts, and keeps the optional DuckDB example separate from the core examples job. The relevant CI checks are green, so this is safe to merge.

Issue #123, opened 2026-05-14
CSV: Add regression coverage for very long single fields


im-anishraj:
## Context
Huge text cells should be tested so parser buffer behavior remains stable.

## Suggested files
arnio/io.py, arnio/exceptions.py, cpp/src/csv_reader.cpp, cpp/include/arnio/csv_reader.hpp, tests/test_csv.py

## Scope
- Confirm the current behavior and intended user-facing contract.
- Implement the smallest focused change that satisfies this issue.
- Add tests for normal, edge, and invalid cases where applicable.
- Update docs or examples if public behavior changes.

## Acceptance criteria
- [ ] The behavior is implemented or clearly documented as unsupported.
- [ ] Tests cover the new behavior or regression.
- [ ] Errors, outputs, or reports are clear for users.
- [ ] Existing tests pass locally.

## Contributor notes
Please comment with your planned approach before starting. Keep the PR focused and link it with Fixes #issue-number once opened.

CodesByY22:
Hi @im-anishraj, I’d like to work on this issue.

My plan is to first reproduce and inspect the current parser behavior with very large single-field CSV inputs to understand existing limits, buffer handling, and any failure modes before making changes.

Proposed approach:

* Review the CSV parsing flow in `cpp/src/csv_reader.cpp` and related interfaces in `cpp/include/arnio/csv_reader.hpp`.
* Identify how large field buffers are currently allocated/expanded and whether any truncation, overflow, excessive memory usage, or parsing instability can occur with huge text cells.
* Add focused regression coverage in `tests/test_csv.py` for:

  * extremely long single text fields
  * quoted long fields
  * multiline long fields
  * mixed normal + huge field rows
  * malformed large-field inputs
* Verify current user-facing behavior and improve error handling/messages if edge cases are unsupported or fail unclearly.
* Keep implementation changes minimal and backward-compatible while ensuring parser stability and no regressions for existing CSV behavior.
* Update docs/examples only if behavior or limits need clarification.

I’ll keep the PR focused and make sure the full test suite passes locally before submission.

im-anishraj:
Thanks @CodesByY22, this plan is clear and the test coverage direction makes sense.

Because you also requested #119 and #131, I am not assigning this yet. Please choose one issue to focus on first and confirm that choice on the issue thread. Once you confirm your preferred issue, I’ll review it for assignment.

mittalsonal:
[...]

PR #888, opened 2026-05-20
test(csv): add regression coverage for very long fields

Code-007-maker:
Fixes #123

## Summary
Adds regression coverage for very long CSV fields to ensure parser stability with large text cells.

## Tests added
- very long unquoted field
- very long quoted field
- mixed normal and large fields

## Validation
- pytest tests/test_csv.py -v
- black tests/test_csv.py


im-anishraj (approved):
Thanks @Code-007-maker. This is focused regression coverage for the assigned CSV long-field issue, and the full CI matrix is green.

PR #894, opened 2026-05-20
fix : raise ValueError on existing safe_divide_columns output_column

### LINKED ISSUE #592: Bug: safe_divide_columns silently overwrites an existing output column
## Problem
The audit found a likely bug or inconsistent behavior: safe_divide_columns warns and overwrites while combine_columns rejects existing output columns.

## Evidence / audit context
- Issue title: Bug: safe_divide_columns silently overwrites an existing output column
- Audit focus: safe_divide_columns warns and overwrites while combine_columns rejects existing output columns.
- Labels selected: type: bug, area: cleaning, area: python-api, priority: medium, difficulty: intermediate, status: needs maintainer, size: s

## Suggested fix
Tighten validation or behavior at the public boundary, add a focused regression test, and document any user-visible change.

## Acceptance criteria

tmdeveloper007:
Fixes #592

### Context
 previously raised a  and overwrote columns if the  already existed, while other cleaning primitives rejected existing output columns with a .

### Solution
Updated  to consistently reject existing output columns by raising a .
Updated unit tests to assert the correct error contract.

### DIFF
--- a/arnio/cleaning.py
+++ b/arnio/cleaning.py
@@ -1094,8 +1094,7 @@ def safe_divide_columns(
         Column name to use as the denominator.
     output_column : str
         Name of the new column to store the division result. Must be a
-        non-empty string. If the column already exists, it will be
-        overwritten and a ``UserWarning`` is raised.
+        non-empty string. If the column already exists, a ``ValueError`` is raised.
     fill_value : float, optional
         Value to use when denominator is zero or null. Defaults to 0.0.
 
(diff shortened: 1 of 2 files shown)


im-anishraj (changes requested):
Thanks @tmdeveloper007. I can’t approve this PR as-is because #592 is still marked `status: needs maintainer` and the issue discussion explicitly says this public behavior change needs a maintainer decision before implementation.

Changing `safe_divide_columns()` from warn-and-overwrite to raising `ValueError` may be the right direction, but it changes the public API contract. Please pause this PR until the intended behavior is confirmed on #592. If we proceed with this direction, the final revision will also need a fully green CI matrix and any necessary user-facing documentation/release-note update for the behavior change.

tmdeveloper007:
Thank you for the quick feedback, @im-anishraj! That makes complete sense. I have paused this PR and will wait for the final decision on the issue #592 discussion before we proceed with this direction. If approved, I will also ensure to add any necessary user-facing documentation and release notes. Let me know how you would like to proceed once a decision is made!

PR #325, opened 2026-05-14
feat: optimize drop_duplicates using hash-based deduplication

gowthamrdyy:
## Summary
This PR replaces the `std::ostringstream` based string serialization in `drop_duplicates` with proper hash-based comparisons. This avoids expensive string allocations per row and improves performance, bringing it closer to pandas performance parity.

## Linked issue
Fixes #324

## Type of change
- [ ] Bug fix
- [ ] Feature
- [x] Performance improvement
- [ ] Documentation
- [ ] Tests
- [ ] Refactor
- [ ] CI / packaging

## Area
- [ ] CSV parser / I/O
- [x] C++ cleaning engine
- [ ] Python API / ArFrame
- [ ] Pipeline / custom steps
- [ ] Data quality / profiling
- [ ] Schema validation
- [ ] pandas interoperability
- [ ] Docs / website
- [ ] Developer tooling

## Testing
- [x] `make test`
- [x] `make lint`
- [ ] Other:

## Contributor checklist
- [x] I read the contributing guide.
- [x] I kept the PR focused on one issue or one logical change.
- [x] I added or updated tests for behavior changes.
- [x] I added or updated docs/examples for public API changes.
- [x] I checked that generated files, logs, and local build artifacts are not committed.
- [x] My PR title uses Conventional Commits (`feat:`, `fix:`, `docs:`, `test:`, `refactor:`, `chore:`).

## Maintainer notes
N/A

PR #901, opened 2026-05-20
perf : optimized drop_duplicates with hash-based comparisons

### LINKED ISSUE #324: Performance: Optimize drop_duplicates with hash-based comparisons
Currently, `drop_duplicates` in `cpp/src/cleaning.cpp` uses `std::ostringstream` to serialize rows into a `std::string` for deduplication. This causes a massive performance overhead due to string allocations.
Replacing this with a proper hash-based comparison (custom hash and equality operators on row indices) will bring Arnio closer to pandas performance parity as mentioned in the ROADMAP. I'd like to work on this for GSSoC.

tmdeveloper007:
Fixes #324

### Context
Currently,  in  used a slow  to serialize each row's cells into a temporary  key for deduplication. This caused severe allocation and formatting overheads.

### Solution
* Removed  string serialization entirely.
* Implemented a custom  struct representing a lightweight in-place reference to a row in a specific  and column subset, including efficient custom hash () and equality operators on row indices.
* Standardized std::variant hashing via standard library  inside  combined with a hash-combine implementation to produce fast, allocation-free hashes.

### DIFF
--- a/cpp/src/cleaning.cpp
+++ b/cpp/src/cleaning.cpp
@@ -38,7 +38,7 @@ struct RowRef {
         for (size_t ci : *cols) {
             const auto& col1 = frame->column(ci);
             const auto& col2 = other.frame->column(ci);
-            
+
             bool null1 = col1.is_null(row);
             bool null2 = col2.is_null(other.row);
             if (null1 != null2) return false;
--- a/cpp/src/cleaning.cpp
+++ b/cpp/src/cleaning.cpp
@@ -28,26 +28,43 @@ static std::vector<size_t> resolve_subset(const Frame& frame,
     return indices;
 }
 
-// Helper: build a row hash for deduplication
-static std::string row_key(const Frame& frame, size_t row, const std::vector<size_t>& cols) {
-    std::ostringstream oss;
-    for (size_t ci : cols) {
-        auto cell = frame.column(ci).at(row);
-        if (std::holds_alternative<std::monostate>(cell)) {
-            oss << "\x00";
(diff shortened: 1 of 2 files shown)


im-anishraj (changes requested):
Thanks @tmdeveloper007. I can’t approve this PR in its current form.

Blockers:
- #324 already has an active PR thread (#325), and the issue discussion says review should stay focused there unless that PR does not land. Please do not open a competing implementation without an explicit maintainer handoff.
- Please remove the unrelated `issues_list.md` catalog from this PR entirely.
- CI is already failing lint, and the C++ diff contains formatting issues that clang-format/CI need to approve.
- For a performance optimization, the final PR needs focused correctness tests/benchmark evidence showing the new hashing preserves `keep="first"`, `keep="last"`, `keep="none"`, subset behavior, and null handling while improving the measured path.

Please pause this until ownership/scope is clarified on #324, then keep any revised PR focused only on the drop_duplicates optimization.

im-anishraj (changes requested):
Thanks for the update. CI is green now, but the main review blockers are still unresolved:

- This PR still includes the unrelated `issues_list.md` catalog, which needs to be removed entirely.
- #324 already has an active implementation/review thread in #325, so this needs an explicit maintainer ownership/scope handoff before it can proceed as a competing implementation.
- The final focused PR still needs correctness/benchmark evidence for the `drop_duplicates` behavior across `keep="first"`, `keep="last"`, `keep="none"`, subset handling, and nulls.

Please keep this paused until #324 ownership is clarified, then rescope the diff to only the performance implementation and its tests/benchmarks.

PR #902, opened 2026-05-20
Add max_errors support to schema validation

### LINKED ISSUE #191: Schema: Add max_errors option to validate
## Context
Large invalid datasets should not require collecting every validation issue before returning.

## Suggested files
arnio/schema.py, tests/test_schema.py, README.md

## Scope

## Acceptance criteria

## Contributor notes

SRSoham:
## Summary
Adds support for stopping schema validation after the configured `max_errors` threshold is reached.

## Changes
- Added `max_errors` support to schema validation
- Stopped validation once error threshold is reached
- Added targeted regression tests for limited validation behavior

## Testing
- pytest tests/test_schema.py
- pytest tests/test_schema.py -k max_errors

Fixes #191


im-anishraj (changes requested):
Thanks @SRSoham. This is assigned correctly and the feature direction is good, but the current patch needs changes before it can merge.

Current blockers:

- CI lint is already failing. Please run Black/Ruff locally; the diff currently has over-indented signatures, trailing whitespace, and extra blank lines.
- Missing-column handling now breaks after the first missing schema field regardless of `max_errors` because the missing-field branch unconditionally breaks. It should only stop once the configured limit is reached.
- Strict-schema extra-column validation can still append issues after the `max_errors` limit. Please apply the cap consistently across the whole validation result.
- Please define and test edge cases for `max_errors=0` and invalid negative values. Negative values currently lead to surprising list slicing behavior.
- Since this is a public API parameter, please update the `validate` / `Schema.validate` docstrings or docs to describe the new argument.

Once those are addressed and CI is green, I can re-review.

im-anishraj (changes requested):
Thanks for the update. A few blockers remain before this can merge:\n\n- CI is still failing. Mypy reports that ValidationResult does not accept passed; please follow the existing constructor pattern, e.g. return ValidationResult(issues=...) and let the pass/fail state be derived.\n- 	ests/test_schema.py::test_schema_validation_stops_after_max_errors is failing across the test matrix, so the behavior still needs to be fixed.\n- Please add regression tests for max_errors=0 and negative max_errors, since the implementation now defines those branches.\n- Please keep the cap consistently applied across missing-column, strict extra-column, and unique-constraint paths.\n\nOnce those are addressed and CI is green, I can re-review.

im-anishraj (changes requested):
Thanks for the follow-up fixes. CI is much closer now, but there are still blockers in the current diff:

- In the strict extra-column branch, validate now returns immediately after the first extra-column issue even when max_errors is not set or the limit has not been reached. Please only return early when the configured limit is actually reached; otherwise continue collecting the remaining validation issues as before.
- Please add regression tests for max_errors=0 and negative max_errors, since those branches are now part of the public behavior.
- Please keep the cap consistently applied across missing columns, strict extra columns, and unique-constraint validation.
- The latest CI run still has Windows jobs in progress, so this cannot be merged yet even after the code fix.

Once those are addressed and the full matrix is green, I can re-review.

im-anishraj (changes requested):
Thanks for fixing the strict extra-column early-return bug. The latest CI run is green, but I still cannot approve yet because the coverage gaps from the previous review remain:

- Please add regression tests for `max_errors=0` and negative `max_errors`, since both branches are now public behavior.
- Please add focused coverage showing the cap is applied consistently outside the normal column-validation path, especially strict extra-column validation and unique-constraint validation.

Once those tests are in place and CI stays green, I can re-review.

im-anishraj (changes requested):
Thanks for adding the requested coverage. The new tests point to the remaining contract mismatch, and the latest CI run is failing across the test matrix.

Blockers:
- test_schema_validation_negative_max_errors should expect the existing ValueError contract for negative max_errors instead of assigning a result.
- test_schema_validation_strict_max_errors_cap passes max_errors=1 but expects two issues; the cap contract should expect one issue.
- test_schema_validation_unique_max_errors_cap passes max_errors=1 but expects three issues; this should also expect one issue.

Please update those tests to match the documented cap behavior, then rerun the full matrix.

im-anishraj (changes requested):
Thanks for the latest fixes. The previous test expectation issues are corrected and lint is now green, but there are still correctness/documentation blockers in the current head:

- `max_errors` is not applied to `schema.rules`; custom rule results are appended after the capped column/unique paths, so a validation run can still return more than the requested limit.
- Early returns currently set `bad_rows=[]` even when the capped issues include row-indexed failures. Please derive `bad_rows` from the returned issues so `ValidationResult.summary()` and reports stay consistent.
- The public docs/docstrings still do not describe `max_errors` on either `validate(...)` or `Schema.validate(...)`.
- The current CI matrix is still running, so this cannot be merged even after the code fixes until all checks are complete and green.

Please add focused regression tests for the custom-rule cap and bad-row behavior, update the docstrings, then request re-review after the full matrix is green.

im-anishraj (changes requested):
Thanks for the update. This fixes several of the previous paths, but the cap still is not applied consistently yet.

Remaining blockers:

- In the `schema.unique` missing-column branch, all missing unique columns are appended without checking `max_errors`, so `unique=["a", "b"]` with `max_errors=1` can still return two issues.
- In the `schema.rules` `KeyError` branch, a missing-column rule issue is appended without checking `max_errors`, so multiple rule functions that reference missing columns can exceed the cap.
- Please add focused regression tests for those two paths, plus the previously requested custom-rule/bad-row cap behavior, so the cap contract is locked outside the normal column validation path too.
- CI is still pending on this head, so this cannot be approved until the full matrix is complete and green.

After those cap paths are covered and CI is green, I can re-review.

im-anishraj (changes requested):
Thanks for the quick follow-up. This closes a few of the earlier paths, but I still can’t approve this head yet.

Remaining blockers:

- Please validate `max_errors` type explicitly. Right now booleans and non-integer numeric values are accepted far enough to produce surprising behavior (`True` acts like `1`, floats can reach list slicing). This public parameter should accept only `int | None`, rejecting `bool` and non-integers with a clear `TypeError`, and still reject negatives with `ValueError`.
- The test named `test_schema_validation_rule_keyerror_respects_max_errors` does not exercise the `KeyError` branch; it returns custom issues normally. Please add a focused test where rule functions actually reference missing columns and verify the cap is applied there.
- Please add focused coverage for custom rules that return row-indexed issues under a cap, including `bad_rows`, so `ValidationResult.summary()` remains trustworthy when validation stops early.
- The `Schema.validate()` docstring formatting is currently hard to read because the `Args` entries are not indented under their parameter names. Please clean that up while documenting the new argument.
- The full CI matrix is still pending on this head, so it cannot be approved yet even after the code/test fixes.

Once those are addressed and CI is green, I can re-review.

im-anishraj (changes requested):
Thanks for the update. This head still can’t be approved.

Blockers:
- `arnio/schema.py` contains unresolved merge-conflict markers (`<<<<<<< HEAD`, `=======`, `>>>>>>> ...`) inside `validate`, so the branch cannot be tested or merged as-is.
- The previous review blockers are still visible: `max_errors` type validation is not explicit yet, and the rule-keyerror test still does not actually exercise a `KeyError` branch because the rule returns custom issues normally.
- Please clean up the `Schema.validate()` docstring indentation while documenting `max_errors`; the current `Args` block is hard to read.
- CI is currently pending on this head, and it must finish green after the conflict and test/doc fixes.

Please resolve the conflict markers first, then rerun formatting/tests and request re-review when CI is green.

im-anishraj (changes requested):
Thanks for resolving the conflict markers and getting the branch much closer. I still cannot approve this head yet.

Remaining blockers:

- Please validate the public `max_errors` type explicitly before validation starts. It should accept only `int | None`, reject `bool` and non-integer values with a clear `TypeError`, and still reject negative integers with `ValueError`. The current `max_errors < 0` check still lets `True` behave like `1` and lets floats/other values fail later in less clear ways.
- `test_schema_validation_rule_keyerror_respects_max_errors` still does not exercise the `KeyError` branch; the rule returns custom issues normally. Please add a focused rule that actually references missing columns and verifies the cap there.
- Please add focused coverage for custom rules returning row-indexed issues under a cap, including `bad_rows`, so capped `ValidationResult.summary()` data stays trustworthy.
- Please clean up the `Schema.validate()` docstring `Args` indentation while documenting `max_errors`; the current entries are not indented under their parameter names.
- The latest CI matrix still has a Windows job pending, so this also cannot merge until the full matrix is complete and green.

After those are addressed, I can re-review.

im-anishraj (changes requested):
Thanks for the update. This is closer, but I still cannot approve this head yet.

Remaining blockers:
- `Schema.validate()`'s docstring has malformed `Args` formatting: `frame` and `max_errors` are not indented under the section, so the public API docs will render poorly.
- `Int64()` now includes unrelated `frame`, `schema`, and `max_errors` entries in its docstring. That looks like a copy/paste slip and should be removed from the field constructor docs.
- Please add explicit invalid-type tests for `max_errors=True` and a non-integer value such as `1.5`, since the implementation now defines a public `TypeError` contract for those cases.
- Please add focused custom-rule coverage where a rule returns multiple row-indexed `ValidationIssue`s and `max_errors` caps both `issues` and `bad_rows`. The current KeyError test is useful, but it does not cover the normal custom-rule result path.
- CI is still pending on this head, so merge remains blocked until the full matrix completes green after the cleanup.

Once these are addressed and CI is green, I can run the local validation gate.

im-anishraj (changes requested):
Thanks for adding the missing validation coverage; the `max_errors=True` / float tests and the custom-rule cap test are the right direction. I still can't approve this head yet because one previous public-doc blocker remains.

Remaining blocker:
- `Schema.validate()` still has malformed `Args:` formatting. The `frame` and `max_errors` entries need to be indented under the `Args:` section, and the continuation lines for `max_errors` should be indented as well. As written, the docstring will render poorly for the public method.

CI is green, but I am holding local validation until that docstring cleanup is fixed.

SRSoham:
@im-anishraj,  
Addressed the remaining review feedback.

Changes made:
- Fixed `Schema.validate()` docstring Args indentation/formatting
- Added explicit `max_errors` type validation coverage
- Added capped custom-rule issue coverage
- Cleaned up validation docs formatting

Validation:
- black arnio/schema.py
- ruff check .
- pytest tests/test_schema.py

im-anishraj (approved):
Thanks for the follow-up. The remaining docstring and coverage blockers are resolved in this head. I also verified the merge result against current main after #1043/#1045 locally on Windows with Python 3.12/MSVC: editable install with dev/parquet extras, full pytest, Black, Ruff, mypy, and diff whitespace checks all pass. CI on the PR head is green.

PR #903, opened 2026-05-20
fix(csv): support has_header in scan_csv

### LINKED ISSUE #81: Bug: `scan_csv` always assumes first line is header - no `has_header` parameter
## Summary

`scan_csv` does not accept a `has_header` parameter — it unconditionally treats the first line as a header row. This makes it impossible to scan schema on headerless CSV files.

## Evidence

In [`arnio/io.py` lines 88–140](https://github.com/im-anishraj/arnio/blob/main/arnio/io.py#L88-L140), `scan_csv` only accepts `path` and `delimiter`:

```python
def scan_csv(
    path: str,
    *,
    delimiter: str = ",",
) -> dict[str, str]:
```

Compare with `read_csv` which has `has_header: bool = True`. Internally, `scan_csv` calls `reader.scan_schema(path)` which always reads the first line as header ([csv_reader.cpp line 236–242](https://github.com/im-anishraj/arnio/blob/main/cpp/src/csv_reader.cpp#L236-L242)).

## Expected Behavior

`scan_csv` should accept `has_header=False` and generate synthetic column names (`col_0`, `col_1`, ...) when the file has no header, matching `read_csv` behavior.

## Actual Behavior

Calling `ar.scan_csv("headerless.csv")` interprets the first data row as column names, returning incorrect schema with data values as keys.

## Impact

Users who work with headerless CSVs cannot use `scan_csv` to preview schema without loading the full file.

## Suggested Fix

1. Add `has_header: bool = True` parameter to `scan_csv`.
2. Pass it through `CsvConfig` to `scan_schema`.
3. Update `CsvReader::scan_schema` to generate synthetic names when `has_header=false`.

Code-007-maker:
Fixes #81

## Summary
- Added `has_header` support to `scan_csv`
- Passed `has_header` through the CSV config flow
- Added synthetic column generation (`col_0`, `col_1`, ...) for headerless CSV files
- Kept behavior aligned with `read_csv`
- Added focused regression tests for both headered and headerless CSV inputs

## Validation
- Ran `pytest tests/test_csv.py -v`
- All CSV tests pass locally

### DIFF
--- a/arnio/io.py
+++ b/arnio/io.py
@@ -625,6 +625,7 @@ def scan_csv(
     thousands_separator: str | None = None,
     sample_size: int | None = None,
     null_values: list[str] | None = None,
+    has_header: bool = True,
 ) -> dict[str, str]:
     """Return schema (column names + inferred types) without loading data.
 
@@ -699,6 +700,7 @@ def scan_csv(
     config.encoding = encoding
     config.trim_headers = trim_headers
     config.thousands_separator = thousands_separator
+    config.has_header = has_header
 
     if null_values is not None:
         config.null_values = _validate_null_values(null_values)
--- a/arnio/io.py
+++ b/arnio/io.py
@@ -650,6 +650,12 @@ def scan_csv(
         1,234 is interpreted as two separate fields.
     sample_size : int, optional
         Number of rows to read for type inference. If None, defaults to 100 rows.
+    has_header : bool, default True
+        Whether the CSV file contains a header row.
+
+        When False, synthetic column names are generated
+        in the form ``col_0``, ``col_1``, etc., matching
+        the behavior of ``read_csv(..., has_header=False)``.
 
     Returns
     -------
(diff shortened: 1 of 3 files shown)


im-anishraj (changes requested):
Thanks @Code-007-maker. The implementation and focused tests look good, and CI is green. One blocker before merge: has_header is a new public scan_csv parameter, but it is not documented in the scan_csv Parameters section yet. Please add the docstring entry for has_header with the default behavior and the has_header=False synthetic-column behavior (col_0, col_1, ...), then I can re-review.

Code-007-maker:
Thanks for the review. I’ve updated the `scan_csv` docstring to document the new `has_header` parameter, including the default behavior and the synthetic column naming behavior when `has_header=False`.

I also re-ran the relevant formatting/tests locally before pushing the update.

im-anishraj (approved):
Thanks for the quick docs update. The scan_csv has_header behavior is now documented, the implementation remains focused, the regression tests cover headered and headerless paths, and the full CI matrix is green. Ready to merge.

PR #914, opened 2026-05-20
fix(frame): enforce consistent column lengths

### LINKED ISSUE #88: Bug: `Frame::num_rows()` assumes uniform column lengths - no invariant enforced on `add_column`
## Summary

`Frame::num_rows()` returns `columns_[0].size()` without validating that all columns have equal length. The `add_column` method does not enforce this invariant either, making it possible to construct a Frame with ragged columns that silently produces incorrect results.

## Evidence

In [`cpp/src/frame.cpp` lines 11–14](https://github.com/im-anishraj/arnio/blob/main/cpp/src/frame.cpp#L11-L14):

```cpp
size_t Frame::num_rows() const {
    if (columns_.empty()) return 0;
    return columns_[0].size();  // Only checks first column
}
```

And [`cpp/src/frame.cpp` lines 70–73](https://github.com/im-anishraj/arnio/blob/main/cpp/src/frame.cpp#L70-L73):

```cpp
void Frame::add_column(Column col) {
    name_index_[col.name()] = columns_.size();
    columns_.push_back(std::move(col));
    // No length validation against existing columns
}
```

## Steps to Reproduce

```python
from arnio._arnio_cpp import Frame, Column, DType

f = Frame()
c1 = Column("a", DType.INT64)
c1.push_back(1)
c1.push_back(2)
c1.push_back(3)

c2 = Column("b", DType.INT64)
c2.push_back(10)
# c2 has only 1 row, c1 has 3

f.add_column(c1)
f.add_column(c2)
print(f.shape())  # Returns (3, 2) but column "b" only has 1 row
```

## Expected Behavior

- `add_column` should assert that the new column's length matches existing columns (or the Frame is empty).
- Alternatively, `num_rows()` should validate column length consistency and raise if mismatched.

## Impact

Ragged frames could cause out-of-range access in cleaning operations, `to_pandas` conversion, or `drop_duplicates` hashing — potentially leading to crashes or memory corruption.

## Suggested Fix

Add a length check in `add_column`:

```cpp
void Frame::add_column(Column col) {
    if (!columns_.empty() && col.size() != num_rows()) {
        throw std::invalid_argument(
            "Column '" + col.name() + "' has " + std::to_string(col.size()) +
            " rows, expected " + std::to_string(num_rows()));
    }
    name_index_[col.name()] = columns_.size();
    columns_.push_back(std::move(col));
}
```

Code-007-maker:
## Summary

Enforce consistent column lengths in `Frame::add_column` to prevent constructing ragged frames with inconsistent row counts.

## Changes

* Added row-count invariant validation in `Frame::add_column`
* Preserved empty-frame behavior
* Added regression tests for:

  * matching column lengths
  * mismatched column lengths
  * empty-frame insertion behavior

Fixes #88

### DIFF
--- a/cpp/src/frame.cpp
+++ b/cpp/src/frame.cpp
@@ -68,6 +68,12 @@ size_t Frame::column_index(const std::string& name) const {
 }
 
 void Frame::add_column(Column col) {
+    if (!columns_.empty() && col.size() != num_rows()) {
+        throw std::invalid_argument("Column '" + col.name() + "' has " +
+                                    std::to_string(col.size()) + " rows, expected " +
+                                    std::to_string(num_rows()));
+    }
+
     name_index_[col.name()] = columns_.size();
     columns_.push_back(std::move(col));
 }
--- a/tests/test_frame.py
+++ b/tests/test_frame.py
@@ -265,7 +265,7 @@ def test_str_keeps_normal_column_names():
 
 
 def test_add_column_accepts_matching_lengths():
-    from arnio._arnio_cpp import Frame, Column, DType
+    from arnio._arnio_cpp import Column, DType, Frame
 
     frame = Frame()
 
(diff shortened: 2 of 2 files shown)


im-anishraj (approved):
Looks good. This keeps the fix focused on the Frame::add_column row-count invariant, preserves empty-frame behavior, adds regression coverage, and the full CI/wheel matrix is green.

PR #920, opened 2026-05-20
fix: locale-independent CSV type inference and integer overflow handling

### LINKED ISSUE #45: Parser: Type inference is locale-dependent and mishandles edge cases (inf, nan, leading spaces)
## Description

The `infer_type` function uses `std::strtoll` and `std::strtod` which are locale-dependent on some platforms. In European locales where `,` is the decimal separator, `"3.14"` may be inferred as STRING instead of FLOAT64.

Additionally, the inference has edge cases:
- `"inf"` and `"nan"` are parsed as valid floats by `strtod` but should arguably be STRING
- `" 123"` (leading space) is treated as STRING because the space isn't trimmed before inference
- `"1e10"` is correctly inferred as FLOAT64 but `"1E10"` behavior may vary by platform

## Acceptance Criteria

- [ ] Use `std::strtod` with `"C"` locale explicitly, or use a locale-independent parser
- [ ] Trim whitespace from values before type inference
- [ ] `"inf"`, `"-inf"`, `"nan"` are treated as FLOAT64 (matching pandas)
- [ ] Add tests for locale-edge-case values
- [ ] Add tests for scientific notation variants

**Difficulty:** Intermediate  
**Files:** `cpp/src/csv_reader.cpp`, `tests/test_csv.py`

shreyansh-tech21:
## Summary
- Make CSV type inference locale-independent (C-locale parsing via classic locale)
- Trim whitespace before bool inference; treat inf/-inf/nan as float64 (pandas-aligned)
- Restore behavior: integer tokens too large for int64 infer as string (not float64)

## Linked issue
Fixes #45

## Type of change
- [x] Bug fix / behavior fix (C++ parser)
- [x] Tests

## Testing
- [x] make test (1029+ passed locally)
- [x] make lint

## Files
- cpp/src/csv_reader.cpp
- tests/test_csv.py

### DIFF
--- a/cpp/src/csv_reader.cpp
+++ b/cpp/src/csv_reader.cpp
@@ -3,9 +3,12 @@
 #include <algorithm>
 #include <cctype>
 #include <cerrno>
+#include <cmath>
 #include <cstddef>
 #include <cstdlib>
 #include <fstream>
+#include <limits>
+#include <locale>
 #include <sstream>
 #include <stdexcept>
 #include <unordered_set>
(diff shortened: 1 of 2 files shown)


im-anishraj (changes requested):
Thanks for the update. The parser direction looks reasonable and CI is green, but this does not yet cover the behavior promised by #45.

Blocker:
- The only new test covers `"3.14"`. Please add focused regression tests for the issue acceptance cases: leading/trailing whitespace before numeric/bool inference, `inf`/`-inf`/`nan` as float64, scientific notation with `e`/`E`, integer overflow staying string instead of falling back to float, and locale-independent dot decimal parsing.

Please push those tests and rerun CI, then I can re-review.

im-anishraj (approved):
Thanks for adding the missing regression coverage. The updated tests now cover whitespace-padded numeric/bool inference, special float tokens, scientific notation with both `e` and `E`, locale-independent dot decimals, and integer overflow staying string. The parser changes preserve the overflow behavior and CI is green.

shreyansh-tech21:
Thanks for the review, @im-anishraj — glad the added tests and overflow behavior look good. I will keep the branch ready; let me know if anything else is needed before merge.

Issue #916, opened 2026-05-20
strip_whitespace is slower than pandas on wide CSVs due to copy-on-write instead of in-place mutation


sudha09-git:
### What happened?

> `ar.strip_whitespace()` is currently slower than the equivalent pandas `.str.strip()` operation on wide CSVs because it uses a copy-on-write approach — creating a new string copy for every cell — instead of mutating strings in place. This directly contradicts arnio's core promise of being faster than pandas.
>
> ```python
> import pandas as pd
> import arnio as ar
> import time
>
> # Generate a wide CSV with many string columns
> import numpy as np
> data = {f"col_{i}": [f"  value_{j}  " for j in range(5000)] for i in range(256)}
> df = pd.DataFrame(data)
> df.to_csv("wide_test.csv", index=False)
>
> # Pandas approach
> start = time.perf_counter()
> for col in df.select_dtypes("object").columns:
>     df[col] = df[col].str.strip()
> pandas_time = time.perf_counter() - start
> print(f"pandas strip time: {pandas_time:.4f}s")
>
> # Arnio approach
> frame = ar.read_csv("wide_test.csv")
> start = time.perf_counter()
> result = ar.strip_whitespace(frame)
> arnio_time = time.perf_counter() - start
> print(f"arnio strip time : {arnio_time:.4f}s")
>
> print(f"arnio is {arnio_time/pandas_time:.2f}x SLOWER than pandas")
> # arnio is ~1.8x SLOWER than pandas on wide CSVs
> ```

### Area

C++ cleaning engine

### Minimal reproduction

import pandas as pd
import arnio as ar
import time

# Generate wide CSV — 256 columns, 5000 rows, all strings with whitespace
data = {f"col_{i}": [f"  value_{j}  " for j in range(5000)] for i in range(256)}
df = pd.DataFrame(data)
df.to_csv("wide_test.csv", index=False)

# Benchmark pandas
start = time.perf_counter()
for col in df.select_dtypes("object").columns:
    df[col] = df[col].str.strip()
pandas_time = time.perf_counter() - start

# Benchmark arnio
frame = ar.read_csv("wide_test.csv")
start = time.perf_counter()
result = ar.strip_whitespace(frame)
arnio_time = time.perf_counter() - start

print(f"pandas : {pandas_time:.4f}s")
print(f"arnio  : {arnio_time:.4f}s")
print(f"arnio is {arnio_time/pandas_time:.2f}x slower than pandas")

### Expected behavior

[...]

PR #923, opened 2026-05-20
perf: optimize strip_whitespace with in-place C++ string mutation

tmdeveloper007:
## Description
This PR optimizes `strip_whitespace` by replacing the slow copy-on-write string approach with **in-place mutation** directly on the C++ string vector, fully resolving #916.

### Changes
* **In-place C++ string mutation**: Instead of allocating a brand new `std::string` and pushing it cell-by-cell for every single cell during whitespace trimming, this PR extracts the column's underlying string vector, trims each string *in-place* using `.erase()`, and moves it into the constructed `Column`.
* **Heap allocation avoidance**: This avoids the redundant creation and heap allocation of two new strings per cell, resulting in dramatic performance improvements on wide CSVs and huge datasets.

Addresses #916.

### DIFF
--- a/cpp/src/cleaning.cpp
+++ b/cpp/src/cleaning.cpp
@@ -284,23 +284,21 @@ Frame strip_whitespace(const Frame& frame, const std::optional<std::vector<std::
     for (size_t ci = 0; ci < frame.num_cols(); ++ci) {
         const auto& src = frame.column(ci);
         if (targets.count(ci) && src.dtype() == DType::STRING) {
-            Column col(src.name(), src.dtype());
+            auto vec = std::get<std::vector<std::string>>(src.data());
             for (size_t r = 0; r < src.size(); ++r) {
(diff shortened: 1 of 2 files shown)


im-anishraj (changes requested):
Thanks for the optimization work. I cannot merge this version yet because there are a few blockers:

- Please remove `issues_list.md` and the unrelated docs/catalog commits. This PR should be limited to the `strip_whitespace` C++ change and any directly related benchmark/test evidence for #916.
- #916 is currently assigned to @sudha09-git, so please coordinate on the issue before continuing with a competing implementation.
- The issue acceptance asks for before/after benchmark numbers on a wide string-heavy fixture and a tall fixture. Please include those results in the PR description or a comment so we can verify the performance claim without adding timing-sensitive CI assertions.

The core C++ direction is worth reviewing again after the scope and benchmark evidence are cleaned up.

PR #937, opened 2026-05-20
fix(frame): add size and duplicate-name guards to Frame::add_column

### LINKED ISSUE #925: bug: Frame::add_column lacks size validation, leading to out-of-bounds C++ memory access/crashes
### What happened?

The C++ function `Frame::add_column` (exposed directly to Python via pybind bindings) does not validate that the length of the new column matches the number of rows (`num_rows()`) in the current `Frame`. 

If a column with a mismatched size is added, downstream operations (like memory usage calculation or C++ cleaning primitives) that iterate from `0` to `num_rows()` will perform out-of-bounds memory accesses on the shorter column, leading to segmentation faults, silent truncation, or undefined C++ memory access behavior.

### Area

C++ core engine (`Frame` & pybind bindings)

### Minimal reproduction

```python
import arnio as ar
import pandas as pd

# Create a frame with 5 rows
frame = ar.from_pandas(pd.DataFrame({"a": [1, 2, 3, 4, 5]}))

# Create a new column with only 2 rows (mismatched size)
col = ar._core._Column("b", ar._core._DType.INT64)
col.push_back(10)
col.push_back(20)

# Add the mismatched column
frame._core_frame.add_column(col)

# Accessing memory usage or downstream C++ cleaners triggers standard out_of_range throw or segmentation fault
print("Memory usage:", frame._core_frame.memory_usage())
```

### Expected behavior

`Frame::add_column` should raise a `std::invalid_argument` or `std::length_error` if the size of the added column does not match `num_rows()` (unless the Frame is empty).

### Suggestion / Root Cause

In `cpp/src/frame.cpp`, `Frame::add_column` is implemented as:
```cpp
void Frame::add_column(Column col) {
    name_index_[col.name()] = columns_.size();
    columns_.push_back(std::move(col));
}
```
It should check that the column size is consistent:
```cpp
void Frame::add_column(Column col) {
    if (!columns_.empty() && col.size() != num_rows()) {
        throw std::invalid_argument(
            "Column size mismatch: expected " + std::to_string(num_rows()) +
            " rows, but got " + std::to_string(col.size())
        );
    }
    name_index_[col.name()] = columns_.size();
    columns_.push_back(std::move(col));
}
```

adityashirsatrao007:
## Summary

Fixes #925

## Problem

`Frame::add_column` lacked row-count validation and duplicate-name checking, causing memory corruption and inconsistent frame state.

## Changes

Both C++ guards (size validation + duplicate-name check) are already present in `main`. This PR adds the missing **regression test**:

### `tests/test_frame.py`
- Added `test_add_column_rejects_duplicate_name` — ensures adding a column with an existing name raises `ValueError`.

Existing tests already cover:
- ✅ valid same-length columns accepted
- ✅ mismatched-length columns rejected with `ValueError`
- ✅ first column in empty frame accepted at any size
- ✅ zero-row explicit frame rejects nonempty first column

## Diff scope

1 file changed — `tests/test_frame.py` (+19 lines)

### DIFF
--- a/tests/test_frame.py
+++ b/tests/test_frame.py
@@ -351,6 +351,25 @@ def test_cpp_frame_explicit_zero_rows_rejects_nonempty_first_column():
         frame.add_column(column)
 
 
+def test_add_column_rejects_duplicate_name():
+    from arnio._arnio_cpp import Column, DType, Frame
+
+    frame = Frame()
+
+    c1 = Column("a", DType.INT64)
+    c1.push_back(1)
+    c1.push_back(2)
+
+    c2 = Column("a", DType.INT64)
+    c2.push_back(3)
+    c2.push_back(4)
+
+    frame.add_column(c1)
+
+    with pytest.raises(ValueError, match="already exists"):
+        frame.add_column(c2)
+
+
 # ArFrame.describe() Tests
 
 


im-anishraj (approved):
Thanks for tightening this down to the duplicate-name guard while preserving the existing size guard behavior. The diff is focused, the regression test covers the duplicate-name path, and the full repository CI/wheel matrix is green.

I also attempted a targeted local validation in a clean temp worktree, but this machine cannot currently build the extension: MSVC/NMake is not available on PATH, and the available MinGW is too old for the project C++17 headers (`<variant>`/`<optional>`). So I am approving the code review based on the clean diff plus green CI, but I am not merging it from this environment.

adityashirsatrao007:
> Thanks for tightening this down to the duplicate-name guard while preserving the existing size guard behavior. The diff is focused, the regression test covers the duplicate-name path, and the full repository CI/wheel matrix is green.
> 
> I also attempted a targeted local validation in a clean temp worktree, but this machine cannot currently build the extension: MSVC/NMake is not available on PATH, and the available MinGW is too old for the project C++17 headers (`<variant>`/`<optional>`). So I am approving the code review based on the clean diff plus green CI, but I am not merging it from this environment.

Thanks for the detailed review and approval. Understood regarding the local environment limitations. I’ll keep an eye on the PR in case any follow-up changes are needed.

im-anishraj (changes requested):
Thanks for the focused guard. I rechecked this against the current main after today’s merges, and the branch no longer merges cleanly.

Blocker before merge:

- Please sync/rebase with the latest `main` and resolve the conflicts in `cpp/src/frame.cpp` and `tests/test_frame.py`. The conflict blocks the required local merge-result validation, including the MSVC native build/tests and Python test gate.

Once the PR is clean again, I will rerun the native + Python validation before making a merge decision.

im-anishraj:
Thanks for rebasing this. The branch is mergeable now, and I approved the pending workflow runs for the new head. I’ll wait for CI/wheel checks to report green before running the local validation pass and reconsidering the review.

im-anishraj (approved):
Thanks for rebasing. I validated the current merge result locally on Windows with Python 3.12/MSVC: editable install with dev/parquet extras, targeted 	ests/test_frame.py -k add_column, full pytest, Black, Ruff, and mypy all pass. CI is green too, so this is ready.

Issue #253, opened 2026-05-14
Performance: Benchmark sparse-null datasets


im-anishraj:
## Context
Null masks and missing-value handling need workload-specific performance coverage.

## Suggested files
benchmarks/benchmark.py, cpp/src/*.cpp, bindings/bind_arnio.cpp, tests/

## Scope
- Confirm the current behavior and intended user-facing contract.
- Implement the smallest focused change that satisfies this issue.
- Add tests for normal, edge, and invalid cases where applicable.
- Update docs or examples if public behavior changes.

## Acceptance criteria
- [ ] The behavior is implemented or clearly documented as unsupported.
- [ ] Tests cover the new behavior or regression.
- [ ] Errors, outputs, or reports are clear for users.
- [ ] Existing tests pass locally.

## Contributor notes
Please comment with your planned approach before starting. Keep the PR focused and link it with Fixes #issue-number once opened.

ionfwsrijan:
Hi @im-anishraj     , I analyzed this issue and I understand the required changes.
I’d like to work on it under GSSoC’26.
I can start immediately and submit a PR soon. Please assign it to me.

im-anishraj:
Thanks @ionfwsrijan for the interest.

Since you requested many issues at once, please choose one issue first and post a plan specific to that issue. Generic assignment comments are not enough, especially for interop/performance tasks. Once you pick one and outline files, behavior, tests, and docs, I can consider assigning it.

maithanisaksham:
Hi @im-anishraj, I am a contributor from GSSoC '26 and I would love to work on this issue!

After reviewing the suggested files, my planned approach is to:
1.  Inspect how existing benchmarks are structured in benchmarks/benchmark.py and see how the C++ bindings in bindings/bind_arnio.cpp are utilized.
2. Implement a focused workload benchmark specifically targeting null masks and missing-value performance coverage.
3. Add corresponding normal and edge-case tests in the tests/ directory to ensure reliability and check against regressions.
4. Verify that all existing and new tests pass locally before submitting.

Could you please assign this issue to me?

im-anishraj:
Thanks @maithanisaksham, this plan is closer to the issue scope than the chunked-conversion request.

Since you also requested #248, please choose one issue to focus on first. If you want this one, please confirm here and tighten the plan around benchmarks specifically:

- deterministic sparse-null dataset generation
- benchmark command and output format
- comparison against existing benchmark cases
[...]

PR #952, opened 2026-05-20
perf: benchmark sparse-null datasets

ionfwsrijan:
## Summary

Adds workload-specific performance coverage for null-mask and missing-value handling across sparse to dense null densities, as requested in #253.

### New data generator
**`benchmarks/generate_data.py`** — `generate_sparse_nulls(rows, path, null_density, seed)` produces a deterministic 6-column CSV (int, float, string, string, string, bool) with controlled null density from 0.1% to 100%. Respects `ARNIO_BENCHMARK_DRY_RUN` for CI smoke tests.

### Focused benchmark
**`benchmarks/benchmark_sparse_nulls.py`** — Standalone script comparing arnio vs pandas on 4 operations across 5 densities (0.1%, 0.5%, 1%, 5%, 20%):

| Operation | Arnio backend |
|---|---|
| `read_csv` | C++ CSV parser |
| `drop_nulls` | C++ native (`vector<bool>` null mask) |
| `fill_nulls` | C++ native |
| `keep_rows_with_nulls` | Python (pandas round-trip) |

Reports average wall-clock time (ms) and peak heap (MB) per operation with speedup ratios.

### Main benchmark integration
**`benchmarks/benchmark_vs_pandas.py`** — Two new `BenchmarkCase` entries added to the standard suite:
- Sparse-null CSV (1M × 6 cols, 1% nulls)
- Dense-null CSV (1M × 6 cols, 20% nulls)

These run the full pipeline (read → clean → drop_nulls → drop_duplicates) with per-operation breakdown.

### Tests
- **`tests/test_benchmarks.py`** — Round-trip through arnio, zero-density produces no nulls, full-density nulls all nullable columns
- **`tests/test_benchmarks_smoke.py`** — Registered in CI-safe allowlist with `--rows 10 --runs 1`

### Docs & tooling
- **`benchmarks/README.md`** — Usage instructions
- **`Makefile`** — `make benchmark-sparse-nulls` target

## How to run

```bash
# Generate all datasets (including sparse-null CSVs)
python benchmarks/generate_data.py

# Run main benchmark suite (includes sparse-null cases)
python benchmarks/benchmark_vs_pandas.py

# Run focused sparse-null benchmark
python benchmarks/benchmark_sparse_nulls.py --rows 1000000 --runs 5

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,5 +22,8 @@ benchmark: ## Run benchmarks
 	python benchmarks/generate_data.py
 	python benchmarks/benchmark_vs_pandas.py
 
+benchmark-sparse-nulls: ## Run sparse-null benchmark
+	python benchmarks/benchmark_sparse_nulls.py
+
 clean: ## Remove build artifacts
 	python -c "import shutil, os, glob; [shutil.rmtree(p, ignore_errors=True) for p in ['dist', 'build', '.pytest_cache'] + glob.glob('*.egg-info') + glob.glob('**/__pycache__', recursive=True)]; [os.remove(f) for f in glob.glob('**/*.pyc', recursive=True)]"
--- a/benchmarks/README.md
+++ b/benchmarks/README.md
@@ -26,7 +26,7 @@ pip install -e ".[dev]"
 ```
 
 ### Step 2: Generate Deterministic Datasets
-Run the data generator to create deterministic tall, wide, and multiline CSV files with fixed random seeds:
+Run the data generator to create deterministic tall, wide, multiline, and sparse-null CSV files with fixed random seeds:
 ```bash
 python benchmarks/generate_data.py
 ```
@@ -34,11 +34,19 @@ This generates the following files:
 * `benchmarks/benchmark_1m.csv` (Tall CSV: 1M rows x 12 columns)
 * `benchmarks/benchmark_wide.csv` (Wide CSV: 5,000 rows x 256 columns)
 * `benchmarks/benchmark_multiline.csv` (Multiline CSV: 100,000 rows x 4 columns)
+* `benchmarks/benchmark_sparse_nulls.csv` (Sparse-null CSV: 1M rows x 6 columns, 1% nulls)
+* `benchmarks/benchmark_sparse_nulls_dense.csv` (Dense-null CSV: 1M rows x 6 columns, 20% nulls)
 
 ### Step 3: Run the Benchmark Suite
 Run the suite using the standard comparison script:
 ```bash
 python benchmarks/benchmark_vs_pandas.py
+```
+
+### Focused benchmark: sparse-null workloads
+Benchmark null-related operations (read_csv, drop_nulls, fill_nulls, keep_rows_with_nulls) across five null densities from 0.1% to 20%:
+```bash
+python benchmarks/benchmark_sparse_nulls.py --rows 1000000 --runs 5
 ```
 
 ---
--- a/benchmarks/benchmark_sparse_nulls.py
+++ b/benchmarks/benchmark_sparse_nulls.py
@@ -33,18 +33,57 @@ from __future__ import annotations
 
 import argparse
 import os
-import sys
 import time
 import tracemalloc
 from pathlib import Path
 
-sys.path.insert(0, os.path.join(os.path.dirname(__file__)))
-
+import numpy as np
 import pandas as pd
 
 import arnio as ar
 
-from generate_data import generate_sparse_nulls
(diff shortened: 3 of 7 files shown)


im-anishraj (changes requested):
Thanks for adding this benchmark coverage. I can’t approve this version yet.

Blockers:

- CI lint is already failing; please run the repo formatter/linter and fix the reported issues.
- The focused benchmark labels operations like `drop_nulls`, `fill_nulls`, and `keep_rows_with_nulls`, but each timed function reads the CSV inside the measured block setup. That makes the operation timings include CSV parsing cost, while `read_csv` is already benchmarked separately. Please either pre-load the frame/dataframe outside the measured section or explicitly label these as parse-plus-operation benchmarks.
- The memory numbers use `tracemalloc`, which does not capture Arnio’s native C++ allocations. Since the output compares Arnio and pandas memory, please switch to the project’s existing process/RSS style measurement or clearly state that this is Python heap only and avoid drawing native-memory conclusions.
- Avoid the `sys.path.insert(...)` import workaround if possible; use the same import pattern as the existing benchmark scripts and keep lint clean.
- Please add at least one smoke/test assertion that the benchmark script runs in dry-run mode and removes its generated temporary files.

Once lint is green and the timing/memory methodology is corrected, I can re-review.

im-anishraj (changes requested):
Thanks for the quick update. The methodology is moving in the right direction, but I still can’t approve this head.

Remaining blockers:

- CI lint is still failing on the latest commit, so please run the repo formatter/linter and push the formatted result before requesting review again.
- `benchmark_sparse_nulls.py` still uses `sys.path.insert(0, os.path.join(os.path.dirname(__file__)))`. I asked to avoid that workaround; please use the same import pattern as the existing benchmark scripts or otherwise keep the script import path consistent with the rest of the benchmark suite.

After lint is green and the import path is cleaned up, I can review the updated timing/memory methodology in detail.

ionfwsrijan:
@im-anishraj All check passed and the required changes have been made. You may review and merge. Make sure appropriate and bonus labels are applied. Thanks for providing me the opportunity

im-anishraj (approved):
Thanks for the follow-up fixes. The latest head addresses the timing methodology issue, removes the `sys.path.insert(...)` workaround, documents the Python-heap limitation clearly, adds the dry-run cleanup coverage, and the full CI matrix is green.

I also ran a local syntax/static check on the changed benchmark/test files. Runtime execution against the PR source is blocked in my local temp worktree because the native `_arnio_cpp` extension is not built there, so I’m relying on the now-green GitHub matrix for the installed-package path.

ionfwsrijan:
@im-anishraj Everything is proper and you're merging it right? Or does it need any changes?

im-anishraj:
Thanks for the follow-up. I’ve approved the latest head because the benchmark methodology issues were addressed and the full CI matrix is green.

I’m not going to merge it from this sweep because local runtime validation against this PR source is blocked in my Windows environment by the missing native `_arnio_cpp` extension build. No further changes are requested from my side; final labels/merge timing can be handled by the maintainer once they’re comfortable with that validation boundary.

ionfwsrijan:
@im-anishraj Sure. I just asked for another issue can you assign me that please as this issue is already solve?

im-anishraj:
Thanks. Please comment on the specific issue you want next with a focused implementation plan. I can’t assign a new issue from this PR thread without seeing the exact issue/scope, and this PR thread should stay focused on #952.

im-anishraj:
Thanks for the focused benchmark work and for addressing the methodology review items. I re-tested the latest head against the current main merge result before merging:

- Python 3.12 editable install from the merge worktree passed
- `ARNIO_BENCHMARK_DRY_RUN=1 python benchmarks/benchmark_sparse_nulls.py --rows 10 --runs 1` passed and cleaned up its generated density CSVs
- Full `pytest tests/ -v --cov=arnio --cov-report=term-missing` passed: 1188 passed, 3 skipped
- `black --check --diff .`, `ruff check . --extend-exclude "*.ipynb"`, and `mypy --config-file mypy.ini` all passed
- GitHub Actions matrix is green

The scope is benchmark-only and matches #253, so I am merging this now.

ionfwsrijan:
@im-anishraj Please add labels to PRs #968 and #952 . I didn't receive points for it. Bonus labels would be appreciated.

im-anishraj:
Added labels matching the linked benchmark issue, including `gssoc:approved`. I’m not adding ad-hoc quality/bonus labels beyond the established issue labels unless the project applies a consistent quality standard for them.

Issue #34, opened 2026-05-07
API: Relax file extension restriction and auto-detect TSV delimiter


im-anishraj:
## Description

The `read_csv` function in `arnio/io.py` (line 24-32) rejects files based on extension alone (`.csv`, `.txt`, `.tsv`). This prevents reading perfectly valid CSV data from files with other extensions (e.g., `.dat`, `.data`, `.log`, `.pipe`). It also incorrectly assumes `.txt` files are always CSV-like.

Additionally, the TSV extension (`.tsv`) is accepted but the delimiter is not automatically set to `\t`, so TSV files will be parsed incorrectly unless the user manually specifies `delimiter="\t"`.

## Current Behavior

```python
ar.read_csv("data.dat")  # ValueError: Unsupported file format
ar.read_csv("data.tsv")  # Parses with comma delimiter — wrong!
```

## Expected Behavior

1. Remove or relax the extension check — trust the user to know their data format
2. Auto-detect delimiter for `.tsv` files (set `delimiter="\t"`)
3. Optionally warn (not error) for unusual extensions

## Suggested Implementation

- Remove the hard extension check from both `read_csv` and `scan_csv`
- Add auto-detection: if the file extension is `.tsv`, default the delimiter to `\t`
- Keep the binary file check (that's legitimately useful)

## Acceptance Criteria

- [ ] `ar.read_csv("data.dat")` works without error
- [ ] `ar.read_csv("data.tsv")` auto-uses tab delimiter
- [ ] Binary file check remains in place
- [ ] Add tests for `.dat`, `.tsv`, and other extensions

**Difficulty:** Beginner — `good first issue`  
**Files:** `arnio/io.py`, `tests/test_csv.py`

KamnaJ5:
Hi @im-anishraj, I’d like to work on this issue.

I reviewed the current behavior and understand the problems with strict extension validation and TSV delimiter handling. I plan to update the file format checks to allow valid CSV-like files with other extensions and add automatic tab delimiter detection for `.tsv` files, along with the necessary tests.

Please assign this issue to me. Thank you

im-anishraj:
Thanks @KamnaJ5.

I’m keeping this unassigned until you choose one focused issue. If this is the one you want, please share the exact behavior for extension handling, TSV delimiter detection, `read_csv()`/`scan_csv()` parity, and regression tests.

rima48-bit:
Hi! I’d like to work on this issue under GSSoC 2026.

My plan:
[...]

PR #959, opened 2026-05-20
fix: prevent type-collision and separator-injection in drop_duplicates row_key (fixes #33)

### LINKED ISSUE #33: Bug: drop_duplicates row_key() has type-collision and separator-injection vulnerabilities
## Description

The `row_key()` function in `cleaning.cpp` (line 29-47) builds deduplication keys by concatenating cell values with a unit separator (`\x1F`). However, the string representation of different types can collide:

1. An `int64_t` value `123` and a `string` value `"123"` produce the same key
2. A `double` value `1.0` and `string` value `"1"` produce different keys (because `std::to_string(1.0)` gives `"1.000000"`)
3. The key `"A\x1FB"` (two columns: `"A"` and `"B"`) collides with one column containing `"A\x1FB"` (literal unit separator in data)

This means `drop_duplicates` can both **miss real duplicates** and **falsely deduplicate distinct rows**.

## Expected Behavior

Row keys should be unambiguous. Each cell value should be prefixed with its type tag, and the separator should be escaped or length-prefixed.

## Suggested Implementation

Prefix each cell with a type tag:
```cpp
if (std::holds_alternative<std::string>(cell)) {
    oss << "S:" << std::get<std::string>(cell);
} else if (std::holds_alternative<int64_t>(cell)) {
    oss << "I:" << std::get<int64_t>(cell);
} else if (std::holds_alternative<double>(cell)) {
    oss << "D:" << std::get<double>(cell);
}
```

## Acceptance Criteria

- [ ] Row keys are type-aware (int `123` != string `"123"`)
- [ ] Add test with mixed-type columns that would previously collide
- [ ] No performance regression on the benchmark

**Difficulty:** Intermediate  
**Files:** `cpp/src/cleaning.cpp`, `tests/test_cleaning.py`

basantnema31:
## Problem

`drop_duplicates` uses a `row_key` function to generate a string hash for
each row. The original implementation had two critical bugs:

1. **Type-collision**: Different types serialized identically — e.g., int `1`
   and string `"1"` both produced `"1"`, so they were incorrectly treated as
   duplicates.
2. **Separator-injection**: String values containing the unit separator `\x1F`
   could produce identical keys for rows that are actually distinct.
3. **Null vs empty string**: `None` (null) and `""` (empty string) both
   serialized as `""` (or empty), making them indistinguishable.

## Fix

Added a `serialize_cell` helper in `cpp/src/cleaning.cpp` that uses
**type-tagged, length-prefixed serialization** for each cell:

| Type | Serialized as |
|------|---------------|
| null | `N` |
| string `s` | `S<len>:<s>` |
| int64 `n` | `I<len>:<n>` |
| double `d` | `F<len>:<d>` (using portable `%.17g` format) |
| bool | `BT` or `BF` |

The length prefix makes the encoding self-delimiting — no separator injection
is possible regardless of cell value content.

## Regression Tests Added

- `test_drop_duplicates_type_collision` — int `1` vs string `"1"` vs float `1.0` vs bool `True` are all kept as distinct rows
- `test_drop_duplicates_null_vs_empty_string` — `None` and `""` are kept as distinct rows
- `test_drop_duplicates_separator_injection` — values containing colons, synthetic prefixes, and `\x1F` bytes produce no false duplicates

Closes #33

### DIFF
--- a/arnio/io.py
+++ b/arnio/io.py
@@ -302,7 +302,7 @@ def _reject_utf8_nul_bytes(path: str) -> None:
 def read_csv(
     path: str | os.PathLike[str],
     *,
-    delimiter: str = ",",
+    delimiter: str | None = None,
     has_header: bool = True,
     usecols: list[str] | None = None,
     nrows: int | None = None,
@@ -318,9 +318,15 @@ def read_csv(
     ----------
     path : str or file-like object
         Filesystem path or text file-like object containing CSV data.
-        Supports .csv, .txt, and .tsv extensions for path inputs.
-    delimiter : str, default ","
-        Field delimiter character.
+        Any file extension is accepted. For ``.tsv`` files, the delimiter
+        is automatically set to ``'\t'`` when ``delimiter`` is omitted.
+    delimiter : str or None, default None
+        Field delimiter character.  When ``None`` (the default) the
+        delimiter is inferred from the file extension: ``'\t'`` for
+        ``.tsv`` files and ``','`` for everything else.  Passing an
+        explicit value always takes precedence — for example,
+        ``delimiter=','`` reads a comma-delimited ``.tsv`` file without
+        any auto-detection.
     has_header : bool, default True
         Whether the file has a header row.
     usecols : list[str], optional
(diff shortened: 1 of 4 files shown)


im-anishraj (changes requested):
Thanks for picking this up. I can’t approve this PR in its current shape.

Blockers:

- The branch includes the unrelated CSV/TSV extension changes from #34 (`arnio/io.py` and `tests/test_csv.py`) in addition to the #33 `drop_duplicates` row-key fix. Please rebase/retarget so this PR contains only the #33 row-key collision fix and its focused tests.
- CI is already failing on this head, including parquet/CI jobs, so please fix the branch and wait for the full matrix to pass before requesting review again.

The row-key direction with type tags and length prefixes looks like the right kind of fix, but I want to review it on a clean, focused branch after CI is green.

basantnema31:
Hi @im-anishraj! Rebased onto a clean branch `fix/issue-33-row-key-only` 
with **only** the #33 row-key fix (cpp/src/cleaning.cpp + tests/test_cleaning.py).
No #34 changes included. Waiting for CI to go green before re-requesting review.

PR #961, opened 2026-05-20
fix: prevent type-collision and separator-injection in drop_duplicates row_key (fixes #33)

### LINKED ISSUE #33: Bug: drop_duplicates row_key() has type-collision and separator-injection vulnerabilities

basantnema31:
## Problem

`drop_duplicates` uses a `row_key` function to generate a string hash for
each row. The original implementation had three critical bugs:

1. **Type-collision**: Different types serialized identically — e.g., int `1`
   and string `"1"` both produced `"1"`, so they were incorrectly treated as
   duplicates.
2. **Separator-injection**: String values containing the unit separator `\x1F`
   could produce identical keys for rows that are actually distinct.
3. **Null vs empty string**: `None` (null) and `""` (empty string) were
   indistinguishable — both serialized the same way.

## Fix

Added a `serialize_cell` helper in `cpp/src/cleaning.cpp` using
**type-tagged, length-prefixed serialization** for each cell:

| Type | Serialized as |
|------|---------------|
| null | `N` |
| string `s` | `S<len>:<s>` |
| int64 `n` | `I<len>:<n>` |
| double `d` | `F<len>:<d>` (portable `%.17g` format) |
| bool | `BT` or `BF` |

The length prefix makes the encoding self-delimiting — no separator injection
is possible regardless of cell value content.

## Regression Tests Added (`tests/test_cleaning.py`)

- `test_drop_duplicates_type_collision` — int `1` vs string `"1"` vs float `1.0` vs bool `True` are all kept as distinct rows
- `test_drop_duplicates_null_vs_empty_string` — `None` and `""` are kept as distinct rows
- `test_drop_duplicates_separator_injection` — values containing colons, synthetic prefixes, and `\x1F` bytes produce no false duplicates

Closes #33

### DIFF
--- a/cpp/src/cleaning.cpp
+++ b/cpp/src/cleaning.cpp
@@ -28,27 +28,6 @@ static std::vector<size_t> resolve_subset(const Frame& frame,
     return indices;
 }
 
-// Helper: build a row hash for deduplication
-static std::string row_key(const Frame& frame, size_t row, const std::vector<size_t>& cols) {
-    std::ostringstream oss;
-    for (size_t ci : cols) {
-        auto cell = frame.column(ci).at(row);
-        if (std::holds_alternative<std::monostate>(cell)) {
-            oss << "\x00";
-        } else if (std::holds_alternative<std::string>(cell)) {
-            oss << std::get<std::string>(cell);
-        } else if (std::holds_alternative<int64_t>(cell)) {
-            oss << std::get<int64_t>(cell);
-        } else if (std::holds_alternative<double>(cell)) {
-            oss << std::get<double>(cell);
-        } else if (std::holds_alternative<bool>(cell)) {
-            oss << (std::get<bool>(cell) ? "T" : "F");
-        }
-        oss << "\x1F";  // unit separator
-    }
-    return oss.str();
-}
-
 static std::string cell_to_string(const CellValue& cell) {
     if (std::holds_alternative<std::string>(cell)) {
         return std::get<std::string>(cell);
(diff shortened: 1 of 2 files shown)


im-anishraj (changes requested):
Thanks for opening the cleaner scoped PR for #33. The scope is much better than the previous mixed CSV/TSV branch, but I can’t approve this head yet.

Blockers:

- CI is already failing across lint and multiple smoke/test jobs, so the branch is not mergeable as-is.
- Please run the formatter/linter locally; the new test formatting in `tests/test_cleaning.py` is likely part of the lint failure.
- After CI is green, please make sure the tests cover the two exact regression classes from #33: type collisions (`1`, `"1"`, `1.0`, `True`, `None`, empty string) and separator/prefix injection across multiple columns.

The serialization approach with type tags and length-prefixed string/numeric values is the right direction, but I need a green, formatted branch before approval.

basantnema31:
Hi @im-anishraj! Updated the branch:

- ✅ Ran `black` + `ruff` locally — all formatting passes
- ✅ Expanded regression tests into separate focused test methods covering:
  - `int 1` vs `string "1"` type collision
  - `None` vs `""` (empty string) collision  
  - Separator injection via `:`, synthetic prefixes `S1:`, and unit sep `\x1f`
  
Waiting for CI to go green before re-requesting review.

im-anishraj (approved):
Thanks for the focused cleanup. I re-reviewed the latest head: the PR is now scoped to the #33 row-key collision fix, the serialization uses type tags plus length prefixes, the tests cover type collisions, null vs empty string, and separator/prefix injection cases, and the full GitHub Actions matrix is green.

Local runtime validation against the PR source is still blocked in my Windows environment by the missing native `_arnio_cpp` extension build, so I’m relying on the green CI matrix for runtime coverage here. This looks good to me.

basantnema31:
@im-anishraj This PR is approved and CI is green — ready to merge whenever you get a chance!

im-anishraj:
Thanks for the reminder. This PR is approved and green, but I’m not merging it from this notification sweep. I’m keeping merges batched and intentional after the maintainer queue settles, especially because several nearby cleaning/CSV changes are still under review.

im-anishraj:
Re-tested the actual merge result against current `main` locally with the native gate because this touches C++ cleaning internals.

Local validation passed:
- MSVC CMake/Ninja build from the merge worktree;
- `ctest --output-on-failure`: 4/4 native C++ tests passed;
- editable install from the merge worktree with Python 3.12/MSVC;
- `pytest tests/ -v --cov=arnio --cov-report=term-missing`: 1168 passed, 3 skipped;
- Black, Ruff, and mypy all passed.

GitHub Actions is green, including wheel smoke jobs. This meets the local + CI merge gate for the row-key collision fix, so I’m merging it.

PR #964, opened 2026-05-20
fix: reject extra CSV fields in permissive mode

### LINKED ISSUE #942: Bug: reject CSV rows with extra fields instead of silently truncating data
## Summary

The CSV reader currently accepts rows that contain more fields than the header and silently ignores the extra values.

## Current behavior

In the C++ CSV reader, parsed row data is stored and columns are built only for known header indexes. If a row has extra fields beyond the header width, those values are not represented in the resulting frame and no error is raised.

Relevant area:
- `cpp/src/csv_reader.cpp`

## Impact

This is silent data loss. A malformed CSV can be read successfully while dropping values from wide rows. For a production data-preparation library, this should fail loudly by default.

## Reproduction

Example CSV:

```csv
a,b
1,2,EXTRA
3,4
```

Expected default behavior: raise a `CsvReadError` explaining the row has too many fields.

Current behavior: the `EXTRA` value is ignored.

## Proposed fix

Validate parsed row width against the expected column count after headers are known.

Suggested behavior:
- If `has_header=True`, every non-empty row must have at most and ideally exactly the header field count under the default strict mode.
- Missing trailing fields can continue to become null only if that remains the documented behavior.
- Extra fields should raise by default.
- Error messages should include row number, expected field count, and actual field count.

## Acceptance criteria

- Rows wider than the header raise `CsvReadError` by default.
- Tests cover extra fields, missing fields, quoted delimiters, and multiline quoted records.
- Error message includes enough context for users to repair the input file.

saurabhhhcodes:
## Summary
- reject CSV rows that contain non-empty extra fields even when `mode="permissive"` is used
- preserve the existing upstream behavior that trims trailing empty extra fields
- add regression coverage so permissive mode still allows missing trailing values but never silently truncates real data

Fixes #942

## Validation
- `/tmp/arnio-pr942/bin/python -m pip install -e .`
- `/tmp/arnio-pr942/bin/python -m pytest tests/test_csv.py -q` -> 180 passed
- `/tmp/arnio-pr942/bin/python -m ruff check tests/test_csv.py`
- `git diff --check`

### DIFF
--- a/cpp/src/csv_reader.cpp
+++ b/cpp/src/csv_reader.cpp
@@ -480,6 +480,8 @@ Frame CsvReader::read(const std::string& path) const {
 
             if (trailing_empty_only) {
                 fields.resize(expected_cols.value());
+            } else {
+                validate_row_width(record_number, expected_cols.value(), fields.size());
             }
         }
 
--- a/tests/test_csv.py
+++ b/tests/test_csv.py
@@ -1119,6 +1119,17 @@ def test_permissive_mode_allows_missing_columns(tmp_path):
     assert pd.isna(df["name"].iloc[1])
 
 
+def test_permissive_mode_rejects_extra_columns(tmp_path):
+    csv_path = tmp_path / "permissive_extra.csv"
+    csv_path.write_text("id,name\n1,Alice,extra\n")
+
+    with pytest.raises(
+        ar.CsvReadError,
+        match="CSV row 2 has 3 fields; expected 2",
+    ):
+        ar.read_csv(csv_path, mode="permissive")
+
+
 def test_invalid_parser_mode(tmp_path):
     csv_path = tmp_path / "invalid_mode.csv"
     csv_path.write_text("id,name\n1,Alice\n")


im-anishraj (changes requested):
Thanks for the focused fix for #942. The C++ direction looks small and sensible, but I can’t approve this head yet because CI lint is already failing.

Please run the repository formatter/linter and push the result. From the diff, `tests/test_csv.py` likely needs spacing cleanup around the newly added top-level test before `test_invalid_parser_mode(...)`.

After lint and the full matrix are green, I can re-review the behavior and tests.

saurabhhhcodes:
Thanks for the quick review. I found this was Black spacing around the new top-level regression test, formatted `tests/test_csv.py`, and force-pushed the cleaned head.

Local validation on the updated head:
- `python -m black --check tests/test_csv.py`
- `python -m ruff check tests/test_csv.py`
- `python -m pytest tests/test_csv.py -q` -> 180 passed
- `git diff --check`

The actual parser behavior is unchanged from the prior head; this push is just the lint cleanup.

saurabhhhcodes:
Tiny correction: I noticed the Black-formatted spacing change had not been staged in the first amend, so I staged it and force-pushed the actual formatted head now (`6f278d8`). The same local validation above was run before pushing.

saurabhhhcodes:
The latest lint log showed the remaining failure was clang-format on the wrapped `validate_row_width(...)` call in `cpp/src/csv_reader.cpp`. I matched the existing one-line call style used a few lines below and force-pushed `c3f534d`.

Re-ran locally after the formatter adjustment:
- `python -m pytest tests/test_csv.py -q` -> 180 passed
- `python -m black --check tests/test_csv.py`
- `python -m ruff check tests/test_csv.py`
- `git diff --check`

im-anishraj (approved):
Thanks for the formatter follow-up. I re-reviewed the latest head: the parser now continues to tolerate trailing empty extra fields but rejects non-empty extra fields instead of silently truncating them in permissive mode, and the regression test covers that behavior.

The full GitHub Actions matrix is green, including wheel smoke tests. Local runtime validation against the PR source is still blocked in my Windows environment by the missing native `_arnio_cpp` extension build, so I’m relying on the green CI matrix for runtime coverage here. This looks good to me.

saurabhhhcodes:
I merged the latest `origin/main` into this branch to clear the conflict GitHub reported. The only manual resolution was keeping the approved `read_csv` behavior here: trailing empty extra fields are trimmed for `read_csv`, while the new upstream `scan_csv` and chunked-read rejection tests are preserved as-is.

Local validation after the merge resolution:
- `/tmp/arnio-merge-venv/bin/python -m pytest tests/test_csv.py -q` -> 184 passed
- `/tmp/arnio-merge-venv/bin/python -m black --check tests/test_csv.py`
- `/tmp/arnio-merge-venv/bin/python -m ruff check tests/test_csv.py`
- `git diff --check`

The PR is mergeable again from GitHub side; waiting on the fresh CI matrix from the merge commit.

im-anishraj:
Thanks for the focused work here. I re-checked this against current `main` and #942 is already resolved there with broader coverage.

At this point this PR is superseded, and merging it would loosen the current behavior for trailing empty extra fields (`id,name\n1,Alice,\n`) from rejected to trimmed. That is no longer aligned with the accepted fix on `main`.

I’m closing this PR so we do not accidentally regress the CSV width-validation contract. Appreciate the contribution.

PR #973, opened 2026-05-20
fix: improve numeric string compatibility messaging during schema validation

### LINKED ISSUE #175: Quality: Add parseability metrics for numeric-looking strings
## Context
Users should know when a string column can safely become int or float.

## Suggested files
arnio/quality.py, arnio/schema.py, tests/test_quality.py, README.md

## Scope

## Acceptance criteria

## Contributor notes

JettiManuSri:
Summary

Improves schema validation messaging for numeric-compatible string columns.

When validation fails due to a dtype mismatch between a string column and an expected numeric dtype (`int64` or `float64`), validation messages now indicate when values appear safely convertible.

This preserves strict schema validation behavior while improving clarity for data preparation workflows.

Fixes #175

Changes

* added numeric compatibility hints to schema validation messages
* preserved strict schema validation behavior with no implicit coercion
* added focused schema validation coverage for:

  * integer-like string values
  * float-like string values
  * invalid mixed values
  * identifier-like string exclusions
  * empty/all-null string columns
* added focused README example for numeric compatibility validation messaging

Example

Before:

```text id="m4q8pk"
Column 'age' has dtype 'string'; expected 'int64'
```

After:

```text id="u7p2vn"
Column 'age' has dtype 'string'; expected 'int64'. Values appear safely convertible to 'int64'
```

Validation

* `black` passes
* `ruff` passes
* focused schema validation tests added

### DIFF
--- a/README.md
+++ b/README.md
@@ -187,6 +187,7 @@ print(selected.columns)
 # ['name', 'revenue']
 ```
 
+
 > Every step above executes in C++. Your Python code is a _configuration_ — not the execution engine.
 
 <br>
@@ -222,7 +223,6 @@ schema = ar.scan_csv("100GB_file.csv", sample_size=500)
 ```
 
 Useful for exploring datasets before committing memory.
-
 </details>
 
 <details>
@@ -239,7 +239,6 @@ print(frame.preview(n=10))  # first 10 rows
 ```
 
 Raises `ValueError` for invalid `n` (zero, negative, or non-integer).
-
 </details>
 
 <details>
@@ -248,8 +247,8 @@ Raises `ValueError` for invalid `n` (zero, negative, or non-integer).
 
 `arnio` provides support for converting Python `decimal.Decimal` objects.
 
-- **Behavior**: Python `Decimal` objects are automatically preserved as high-precision strings during serialization/binding to prevent floating-point precision loss.
-- **Caveat**: When reading back into Pandas, `to_pandas()` returns these as string (`object` dtype) columns. You will need to explicitly cast them back to `Decimal` objects on the resulting DataFrame if you want to resume exact math.
+* **Behavior**: Python `Decimal` objects are automatically preserved as high-precision strings during serialization/binding to prevent floating-point precision loss.
+* **Caveat**: When reading back into Pandas, `to_pandas()` returns these as string (`object` dtype) columns. You will need to explicitly cast them back to `Decimal` objects on the resulting DataFrame if you want to resume exact math.
 
 Example:
 
@@ -292,7 +291,6 @@ clean = ar.pipeline(frame, [
 ```
 
 Custom steps run through a pandas↔ArFrame conversion bridge. Prototype in Python, then optionally migrate hot paths to C++ for full speed.
-
 </details>
 
 <details>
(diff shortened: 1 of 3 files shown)


im-anishraj (changes requested):
Thanks for the focused cleanup around numeric string compatibility messaging. I can’t approve this head yet.

Blockers:

- The dtype validation issue no longer passes `severity=field_def.severity` when constructing the `ValidationIssue`. That changes schemas using warning-level fields into error-level dtype issues, which is a behavioral regression. Please preserve the existing severity propagation and add a regression test for `Int64(severity="warning")` or `Float64(severity="warning")`.
- The compatibility check currently reuses `_suggest_column_dtype(...)`, which is not target-aware. For example, bool-like numeric strings such as `"0"`/`"1"` can be suggested as `bool` even when the schema expects `int64` or `float64`, so the hint can be omitted for values that are still safely convertible to the expected numeric dtype. Please make the check explicitly validate convertibility to the requested dtype instead of relying on the general profiling suggestion helper.

I’m leaving the workflow approval held for now because these blockers are visible in the diff. Once the behavior is fixed, we can run CI and re-review.

im-anishraj (changes requested):
Thanks for the update. Preserving `severity=field_def.severity` is fixed now, and the compatibility helper is moving in the right direction. I still can’t approve this head yet.

Blockers:

- `_is_safely_convertible_to_dtype(...)` currently treats identifier-like strings such as `"001"`, `"002"`, `"003"` as safely convertible to `int64`, because `pd.to_numeric(...).astype("int64")` succeeds. That contradicts the new `test_dtype_validation_does_not_report_safe_conversion_for_identifier_like_columns`, so CI will fail on this head. Please make the helper preserve the intended leading-zero/identifier exclusion or pass enough column/value context to enforce it.
- For expected `int64`, `pd.to_numeric(...).astype("int64")` can also accept decimal-looking strings by truncating after conversion, which is not a safe int conversion. Please make the int path stricter: values should parse as integers without lossy truncation.
- Checks are not available yet on this head, so please push the helper fix and wait for CI before requesting approval.

Once the helper matches the documented/tested behavior and CI is green, I can re-review.

JettiManuSri:
Thanks @im-anishraj for the follow-up review.

Updated the compatibility helper to:

* preserve identifier-style leading-zero exclusions
* require strict integer parsing for int64 compatibility checks without lossy truncation

The latest changes are now pushed and ready for CI once workflows are approved.

im-anishraj (changes requested):
Thanks for the update and for getting the full CI matrix green. The earlier blockers are fixed, but I found one remaining correctness issue before I can approve.

Blocker:

- For expected `int64`, `_is_safely_convertible_to_dtype(...)` currently returns `True` after `pd.to_numeric(values, errors="raise")` without checking that the parsed values are actually representable as signed int64. Pandas accepts values like `"9223372036854775808"` as `uint64` and very large integer strings as `object`, so the validation message can incorrectly say they are safely convertible to `int64`.

Please make the int path verify signed int64 bounds, for example with `np.iinfo("int64")`, and add focused regression tests for just-above-max and just-below-min integer strings. After that, request re-review again.

im-anishraj (approved):
Thanks for the quick follow-up. This now addresses the previous blockers:

- identifier-like leading-zero strings are excluded;
- integer-looking decimals are not treated as safe int64 conversions;
- signed int64 bounds are checked with regression coverage;
- severity propagation is preserved;
- the full CI/smoke matrix is green.

Approved from review. I’m not merging from this environment because local compiled-extension testing is still blocked here, but the diff and CI look good.

im-anishraj:
Re-tested the actual merge result against current `main` locally.

Local validation passed:
- editable install from the merge worktree with Python 3.12;
- `pytest tests/ -v --cov=arnio --cov-report=term-missing`: 1125 passed, 3 skipped;
- Black, Ruff, and mypy all passed.

GitHub Actions is also green, and this PR only touches Python/docs/tests. This now meets the local + CI merge gate, so I’m merging it.

JettiManuSri:
Thank you @im-anishraj for the detailed review and for clearly calling out each blocker with concrete examples and expected behavior. The explanations around severity propagation, target-aware dtype validation, identifier-like numeric strings, and safe int64 conversion semantics were especially helpful in narrowing down the fixes efficiently. I also appreciate the careful re-testing and validation across both CI and local environments before merging. Thanks again for the thorough review process and guidance throughout this PR.

PR #976, opened 2026-05-20
fix: reject CSV rows with extra non-empty fields instead of silently truncating

### LINKED ISSUE #942: Bug: reject CSV rows with extra fields instead of silently truncating data

Rishi895-Rathi:
Fixes #942

## Problem
CSV rows with extra non-empty fields were silently truncated instead of raising an error. This caused silent data loss.

## Changes
Added an `else` block in all 3 places in `cpp/src/csv_reader.cpp` where extra fields were being silently trimmed. Now if the extra fields are non-empty, a `std::runtime_error` is raised with the row number, expected field count, and actual field count.

## Behavior
- `a,b\n1,2,EXTRA` → raises `CsvReadError` with row number and field counts
- `a,b\n1,2,` → still allowed (trailing empty field, silently trimmed)

### DIFF
--- a/cpp/src/csv_reader.cpp
+++ b/cpp/src/csv_reader.cpp
@@ -438,9 +438,16 @@ Frame CsvReader::read(const std::string& path) const {
                     break;
                 }
             }
-
+            
             if (trailing_empty_only) {
                 fields.resize(expected_cols.value());
+            } else {  // changes made here to throw error only if extra fields are not empty
+                throw std::runtime_error(
+                    "CSV row " + std::to_string(record_number) +
+                    " has " + std::to_string(fields.size()) +
+                    " fields but header defines " +
+                    std::to_string(expected_cols.value()) + " columns."
+                );
             }
         }
 
@@ -589,6 +596,13 @@ std::vector<std::pair<std::string, std::string>> CsvReader::scan_schema(
 
             if (trailing_empty_only) {
                 fields.resize(num_cols);
+            } else {  // throw error if extra fields are non-empty
+                throw std::runtime_error(
+                    "CSV row " + std::to_string(sample_count + 2) +
+                    " has " + std::to_string(fields.size()) +
+                    " fields but header defines " +
+                    std::to_string(num_cols) + " columns."
+                );
             }
         }
         validate_row_width(sample_count + 2, num_cols, fields.size());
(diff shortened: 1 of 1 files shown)


im-anishraj (changes requested):
Thanks for opening this. I can’t approve this PR.

Blockers:

- This branch is currently conflicting with `main`, so checks are not available and the PR is not mergeable.
- The same issue is already covered by #964, which has focused tests and a green CI matrix. Please avoid duplicate PRs for the same fix unless there is a clear reason this approach should replace the existing one.
- This diff changes the parser behavior in C++ but does not add regression tests for `read_csv`, `scan_csv`, or chunked reads.

I recommend closing this PR and following the already-reviewed #964 path. If you want to continue with a different approach, please rebase onto current `main`, explain how it differs from #964, and add focused tests.

Rishi895-Rathi:
is this assigned to someone ?

im-anishraj:
Not currently assigned to you from my side. This issue is already being handled through #964, which is further along and includes focused tests, so please close this duplicate PR unless you have a clearly different approach to propose.

If you want to continue with this one instead, please first rebase it onto current `main`, resolve the conflicts, explain how it differs from #964, and add regression tests for `read_csv`, `scan_csv`, and chunked reads.

Rishi895-Rathi:
Thanks for the review and the clear feedback. I understand that #964 is already further along with focused tests and a cleaner implementation. I'm closing this PR to avoid duplication. I learned a lot from this process — reading C++ parser code, understanding the fix, and navigating the contribution workflow. I'll pick a fresh unassigned issue and make sure to comment and get assigned before starting next time. Appreciate the time taken to review!

im-anishraj:
Thanks for closing this and for understanding the duplicate-PR concern. That is the right call here. Please pick a fresh unassigned issue next, comment first, and wait for assignment before opening the implementation PR.

Issue #248, opened 2026-05-14
Interop: Add chunked from_pandas design


im-anishraj:
## Context
Very large pandas frames should eventually convert without one giant memory spike.

## Suggested files
arnio/convert.py, bindings/bind_arnio.cpp, tests/test_convert.py, README.md

## Scope
- Confirm the current behavior and intended user-facing contract.
- Implement the smallest focused change that satisfies this issue.
- Add tests for normal, edge, and invalid cases where applicable.
- Update docs or examples if public behavior changes.

## Acceptance criteria
- [ ] The behavior is implemented or clearly documented as unsupported.
- [ ] Tests cover the new behavior or regression.
- [ ] Errors, outputs, or reports are clear for users.
- [ ] Existing tests pass locally.

## Contributor notes
Please comment with your planned approach before starting. Keep the PR focused and link it with Fixes #issue-number once opened.

ionfwsrijan:
Hi @im-anishraj     , I analyzed this issue and I understand the required changes.
I’d like to work on it under GSSoC’26.
I can start immediately and submit a PR soon. Please assign it to me.

im-anishraj:
Thanks @ionfwsrijan for the interest.

Since you requested many issues at once, please choose one issue first and post a plan specific to that issue. Generic assignment comments are not enough, especially for interop/performance tasks. Once you pick one and outline files, behavior, tests, and docs, I can consider assigning it.

maithanisaksham:
Hi @im-anishraj, I am a participant in GSSoC '26 and I would love to work on this issue!

My planned approach to tackle the memory spike during pandas DataFrame conversion is to:
1. Analyze arnio/convert.py to identify exactly where the data is being fully duplicated or loaded entirely into memory during the conversion process.
2. Implement a memory-efficient conversion strategy (such as chunk-by-chunk processing or utilizing zero-copy mechanisms if applicable with the underlying C++ layer in bindings/bind_arnio.cpp).
3. Add a focused test case in tests/test_convert.py that processes a large simulated DataFrame to verify memory stability and ensure no regression occurs.
4. Ensure all local tests pass successfully.

Could you please assign this issue to me?

im-anishraj:
Thanks @maithanisaksham, this is a useful direction, but this issue is a design-heavy interop task and needs a more specific plan before assignment.

Also, I saw that you requested #253 too. Please choose one issue to focus on first.

[...]

PR #979, opened 2026-05-21
bench: add from_pandas memory benchmark, design doc, and dry-run smoke test

SiddharthRiot:
Refs #248

## What's changed
- Added `benchmarks/benchmark_from_pandas_memory.py` measuring peak memory per column during `from_pandas()` conversion
- Added `docs/chunked_from_pandas_design.md` covering current behavior, proposed API contract, dtype/null/index limitations, and follow-up implementation plan
- Added `tests/test_benchmark_from_pandas.py` with a dry-run smoke test (`ARNIO_BENCHMARK_DRY_RUN=1`)

## Key findings
- Numeric columns are the primary allocation hotspot (`col_int`: 4.13 MB, `col_float`: 3.82 MB at 100k rows) due to `.tolist()` materializing the full column in Python memory
- Peak memory scales linearly: 0.93 MB at 10k rows → 88.77 MB at 1M rows

## What's not in this PR
- No public API change
- Full `chunk_size` implementation deferred to follow-up PR after design is accepted

### DIFF
--- a/benchmarks/benchmark_from_pandas_memory.py
+++ b/benchmarks/benchmark_from_pandas_memory.py
@@ -1,10 +1,12 @@
 import sys
 import tracemalloc
+
 import numpy as np
 import pandas as pd
 
 try:
     import arnio as ar
+
     HAS_ARNIO_CPP = True
 except (ImportError, ModuleNotFoundError):
     HAS_ARNIO_CPP = False
@@ -13,16 +15,18 @@ except (ImportError, ModuleNotFoundError):
 def generate_large_frame(row_count: int) -> pd.DataFrame:
     np.random.seed(42)
     str_pool = ["alpha", "beta", "gamma", "delta", "epsilon"]
-    return pd.DataFrame({
-        "col_int":   np.random.randint(0, 10000, size=row_count),
-        "col_float": np.random.randn(row_count),
-        "col_bool":  np.random.choice([True, False], size=row_count),
-        "col_str":   np.random.choice(str_pool, size=row_count),
-    })
+    return pd.DataFrame(
+        {
+            "col_int": np.random.randint(0, 10000, size=row_count),
+            "col_float": np.random.randn(row_count),
+            "col_bool": np.random.choice([True, False], size=row_count),
+            "col_str": np.random.choice(str_pool, size=row_count),
+        }
+    )
 
 
 def measure_from_pandas(row_count: int):
-    df =generate_large_frame(row_count)
+    df = generate_large_frame(row_count)
 
     tracemalloc.start()
     _ = ar.from_pandas(df)
@@ -54,7 +58,12 @@ def run():
         sys.exit(0)
 
     import os
-    scales = [10] if os.getenv("ARNIO_BENCHMARK_DRY_RUN") == "1" else [10_000, 100_000, 1_000_000]
+
+    scales = (
+        [10]
+        if os.getenv("ARNIO_BENCHMARK_DRY_RUN") == "1"
+        else [10_000, 100_000, 1_000_000]
+    )
 
     print("=" * 60)
     print("BENCHMARK: from_pandas() peak memory")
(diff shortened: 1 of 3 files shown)


im-anishraj (changes requested):
Thanks for putting this together. I’m going to request changes for a couple of blockers before we can consider this reviewable/mergeable:

- `tests/test_csv.py` contains unresolved conflict markers (`<<<<<<<`, `=======`, `>>>>>>>`) from the earlier CSV edge-case commits. That will make the test suite/formatting fail immediately, and those unrelated CSV changes should be removed or cleaned up from this PR.
- The benchmark dry-run still calls `measure_column_hotspot(100_000)`, so `ARNIO_BENCHMARK_DRY_RUN=1` does not keep the smoke test small. Please make the hotspot row count respect dry-run as well, for example by using the same reduced scale.

After those are fixed, I’ll re-check the diff and run the local representative test gate before any merge decision.

im-anishraj:
Thanks for the quick update. The latest diff removes the unrelated `tests/test_csv.py` conflict-marker changes, and the dry-run hotspot now uses the reduced row count, so the review blockers I raised look addressed.

I’m going to keep this pending until the full CI matrix finishes, then I’ll run the local representative gate before any final approval or merge decision.

SiddharthRiot:
@im-anishraj CI is green now!

im-anishraj (changes requested):
Thanks again for the cleanup. I ran the local representative gate on a clean merge-result checkout and the mechanics pass:

- `pytest tests/ -v --cov=arnio --cov-report=term-missing` passed: 1205 passed, 3 skipped, coverage 93%
- `black --check --diff .` passed
- `ruff check . --extend-exclude "*.ipynb"` passed
- `mypy --config-file mypy.ini` passed
- `ARNIO_BENCHMARK_DRY_RUN=1 python benchmarks/benchmark_from_pandas_memory.py` passed

One remaining correctness issue before approval: `hotspot = measure_column_hotspot(scales[0])` makes the normal benchmark measure the hotspot at 10k rows, while the output still says `HOTSPOT: per-column peak memory at 100k rows` and the design document reports 100k-row hotspot numbers. Please split this into an explicit value, e.g. `hotspot_rows = 10 if dry_run else 100_000`, and use that value in both the label and `measure_column_hotspot(...)`.

After that, I can re-run the small local gate and proceed with review.

im-anishraj (changes requested):
Thanks for the quick fix. The `hotspot_rows` approach is the right direction, but the latest commit accidentally removed the script entrypoint:

```python
if __name__ == "__main__":
    run()
```

Without that block, `python benchmarks/benchmark_from_pandas_memory.py` exits successfully without running the benchmark, so the dry-run smoke test becomes a false positive. Please restore the entrypoint and ideally assert that the dry-run output contains the benchmark/hotspot headings so this cannot regress silently.

im-anishraj:
Thanks, this latest update restores the benchmark entrypoint and strengthens the dry-run smoke assertion. I’m going to wait for the rerun CI matrix to finish, then I’ll run the focused local recheck before approving or merging.

SiddharthRiot:
restored the `if __name__ == "__main__"` entrypoint and added assertions for the benchmark/hotspot headings in the dry-run smoke test

im-anishraj (approved):
Thanks for iterating on this. I rechecked the final diff and ran the focused local merge-result gate after CI went green:\n\n- pytest tests/test_benchmark_from_pandas.py -v\n- ARNIO_BENCHMARK_DRY_RUN=1 python benchmarks/benchmark_from_pandas_memory.py\n- black --check --diff on the changed Python files\n- ruff check on the changed Python files\n- mypy --config-file mypy.ini\n\nThe previous full local gate also passed on this PR scope, and the latest CI matrix is green. The earlier review blockers are resolved, so approving.

SiddharthRiot:
@im-anishraj Thank you! Could you also add the relevant labels to this PR

PR #987, opened 2026-05-21
perf(csv): optimize parser I/O and field allocations

### LINKED ISSUE #44: Performance: CSV parser I/O and field allocation optimizations
## Description

The parser reads fields character-by-character using `std::string::operator+=` which triggers repeated small allocations. The `std::ifstream` also reads the file without explicit buffering configuration.

## Suggested Optimizations

1. **File I/O:** Use `mmap` (Linux/macOS) or memory-mapped file I/O for large files, eliminating the `getline` overhead
2. **Field parsing:** Pre-allocate a field buffer with `reserve()` and reuse it across fields
3. **Column vectors:** Call `reserve(estimated_rows)` on all column vectors before the parsing loop. The row count can be estimated from file size / average line length.
4. **String interning:** For low-cardinality string columns, use a string pool to avoid storing duplicate strings

## Acceptance Criteria

- [ ] CSV read speed matches or beats pandas on the 1M-row benchmark
- [ ] Add `--benchmark` flag to CI for automated regression tracking
- [ ] Document the optimization in code comments

**Difficulty:** Advanced  
**Files:** `cpp/src/csv_reader.cpp`, `cpp/src/column.cpp`

basantnema31:
Closes #44

## Description
This PR addresses the performance bottlenecks in the CSV parser by focusing on I/O overhead and heap allocations. It strictly avoids POSIX-only APIs (like `mmap`) to preserve full MSVC and cross-platform compatibility.

## Optimizations
1. **BufferedStreamReader**: Replaced the character-by-character `stream.get(c)` logic in `getline_universal` with a custom 64KB chunk buffer reader. This drastically cuts down I/O overhead while natively supporting `\n`, `\r\n`, and `\r`.
2. **Reusable Buffers**: Refactored `read_record` into a `RecordReader` class to preserve line string capacities across reads.
3. **Allocation-free `parse_line`**: Added an overloaded `CsvParser::parse_line` that takes a mutable reference to `std::vector<std::string>`. This allows `CsvReader` and `CsvChunkReader` to completely reuse field capacities and vector reservations per row, avoiding a massive amount of string allocations.

## Benchmarks
Because this targets a massive reduction in allocations and system calls, it should run significantly faster. I am relying on the CI environment's benchmark actions (or I can run a benchmark script provided by a maintainer) since my local MSVC toolchain is currently down. Please let me know if you'd like me to run it through a specific profiling test suite!


im-anishraj (changes requested):
Thanks for taking this on and for keeping the approach cross-platform. I can’t approve this head yet.

Blockers before re-review:

- This rewrites the core CSV record-reading path, but the PR adds no targeted regression tests for the new reader. Please add coverage for behavior that changed or could regress: LF/CRLF/CR line endings, multiline quoted records, escaped quotes inside multiline records, `skip_rows`, `nrows`, `scan_csv`, and `read_csv_chunked` parity with normal `read_csv`.
- Please include edge coverage around buffer boundaries. The new `BufferedStreamReader` has custom logic for `\r\n`, EOF, and NUL bytes, so we need at least one larger fixture that crosses the 64KB buffer boundary, including a quoted multiline field and CRLF handling near the boundary.
- This is a performance PR, but there is no benchmark result or repeatable benchmark command in the PR. Please provide before/after numbers on a representative large CSV, ideally including the 1M-row benchmark mentioned in #44, and document the exact command/environment used.
- Please wait for the full CI/wheel matrix to finish green after adding the tests. Because this touches C++ parser/wheel behavior, the Windows MSVC wheel smoke, manylinux wheel smoke, macOS arm64 wheel smoke, and Python matrix are required before approval.

Once the behavior coverage and benchmark evidence are in place, I can do a deeper local Windows/MSVC validation pass.

im-anishraj (changes requested):
Thanks for the follow-up tests and benchmark script. The current head is still blocked before behavior/performance review:

- C++ builds fail across native, smoke, manylinux, macOS arm64, and Windows MSVC jobs because `RecordReader` is ambiguous/incomplete after adding `std::unique_ptr<class RecordReader>` in the public header while defining another `RecordReader` in the anonymous namespace (`cpp/src/csv_reader.cpp`, around the read/scan/chunk paths).
- `Lint (Black, Ruff, clang-format)` is still red because `benchmarks/benchmark_csv.py` needs Black formatting.
- Because this is a C++ parser/wheel-impacting PR, please wait for the required wheel smoke jobs and full CI matrix to be green before requesting re-review.

Once compile/lint are fixed, please also include actual before/after benchmark numbers and the exact command/environment used, not only the script.

im-anishraj (changes requested):
Thanks for the quick follow-up. This head still cannot move forward because the required build jobs are red.

Current blocker:

- The C++ core now builds, but the Python extension / wheel build still fails because `CsvChunkReader` owns `std::unique_ptr<arnio::RecordReader>` from `cpp/include/arnio/csv_reader.h` while `RecordReader` remains incomplete from translation units that only include the header. The failure shows up in `bindings/bind_arnio.cpp` / pybind11 destruction with `invalid application of sizeof to incomplete type 'arnio::RecordReader'` and MSVC’s `can't delete an incomplete type`.

Please make the owned type complete wherever `CsvChunkReader` is destroyed, or move the ownership/destructor boundary out of the header in a C++-safe way. The lint job is also still red, so please run the repo format/lint checks. After that, wait for the required C++ native, wheel smoke, smoke, and Python matrix jobs to go green, then include actual before/after benchmark numbers with the exact command/environment used.

im-anishraj (changes requested):
Thanks for the quick follow-up. This head still needs changes before re-review:

- `Lint (Black, Ruff, clang-format)` is red. Ruff reports unsorted imports in `benchmarks/benchmark_csv.py` and `tests/test_csv_regression_pr987.py`, plus unused `io` and `pathlib.Path` imports in `tests/test_csv_regression_pr987.py`.
- The Python matrix is red. `tests/test_csv_regression_pr987.py::test_chunked_parity_skip_nrows` calls `ar.read_csv(..., skip_rows=1, nrows=3)`, but `read_csv` does not accept `skip_rows`, so it fails with `TypeError: read_csv() got an unexpected keyword argument 'skip_rows'`.
- After fixing those, please wait for the required full CI/wheel matrix to go green and still add actual before/after benchmark numbers with the exact command and environment used.

im-anishraj (changes requested):
Thanks for the follow-up. This head is still blocked before re-review:

- `Lint (Black, Ruff, clang-format)` is red with clang-format violations in `cpp/src/csv_reader.cpp`.
- The Python matrix is still red: `tests/test_csv_regression_pr987.py::test_chunked_parity_skip_nrows` now fails with `KeyError: 'id'`.
- The PR still needs actual before/after benchmark numbers with the exact command and environment used.

Please fix the format/test failures, wait for the full required C++/wheel/Python matrix to go green, and add the benchmark evidence before requesting another review.

im-anishraj (changes requested):
Thanks for the follow-up. I’m going to stop this head before another full review because the diff has moved outside the scope of #44.

Blockers before re-review:

- The latest diff includes unrelated files (`extract.py`, `generate_logos.py`, `logo_dark.png`, and `logo_light.png`). Please remove these and keep the patch limited to CSV parser, benchmark, and regression-test changes.
- The required C++/wheel/Python CI matrix is still pending on this head, so it is not reviewable for approval yet.
- The PR still needs actual before/after benchmark numbers with the exact command and environment used.

Please trim the unrelated files, let the full required CI matrix finish green, and add the benchmark evidence before requesting another review.

im-anishraj (changes requested):
Thanks for cleaning up the unrelated files and getting several jobs green. This head still can’t move forward.

Blockers before re-review:

- The latest diff renames the public `read_csv_chunked(..., skip_rows=...)` parameter to `skiprows`. That is a breaking API change and is outside the scope of #44. Please keep the existing `skip_rows` API for chunked CSV reads and fix the parity test without renaming the public parameter.
- The Python matrix is red on this head, so the required CI signal is still failing.
- The PR still needs actual before/after benchmark numbers with the exact command and environment used. A benchmark script alone is not enough for a performance PR.

Please restore the existing chunked API, fix the failing Python jobs, wait for the full required matrix to go green, and add the benchmark results before requesting another review.

basantnema31:
**Benchmark Evidence**
Command used: `python benchmarks/benchmark_csv.py`
Environment: Windows 11, Python 3.12, MinGW/MSVC

**Before (main branch):**
`Time taken to read large_benchmark.csv: 64.2140 seconds`

**After (this PR):**
`Time taken to read large_benchmark.csv: 38.0195 seconds`

The chunked `BufferedStreamReader` and allocation-free string reuse resulted in roughly a ~40% improvement in read speed for large datasets.

im-anishraj (changes requested):
Thanks for the updated benchmark evidence and latest push. The benchmark numbers are helpful, but this head is still blocked:

- The current diff still renames the public `read_csv_chunked(..., skip_rows=...)` API to `skiprows`, including tests/docs. That is a breaking API change and remains outside #44; please preserve `skip_rows` for chunked reads.
- CI is red across required parser/wheel/Python jobs on this head, including smoke/core examples, wheel smoke jobs, C++ native tests, and the Python matrix.

Please restore the existing public API, fix the failing required jobs, and request re-review only after the full C++/wheel/Python matrix is green.

im-anishraj:
Thanks for restoring the chunked `skip_rows` API. That addresses the API blocker I called out, but this head still is not ready for re-review: lint is red, most Python matrix jobs are failing, and a few jobs are still in progress.

Because this touches parser/wheel behavior, please wait for lint plus the full required C++/wheel/Python matrix to finish green before requesting re-review. Once the matrix is green, I’ll do the local Windows/MSVC validation pass.

im-anishraj (changes requested):
Thanks for getting the required matrix green and for adding benchmark numbers. This is much closer, but I still can’t approve this parser rewrite yet because a few of the requested behavior-covering tests are still missing.

Remaining blockers before re-review:

- Please add targeted coverage for `skip_rows` and `nrows` on `read_csv`; both paths now use the new `RecordReader` and reusable parse buffer.
- Please add coverage for `scan_csv` and `read_csv_chunked` parity with normal `read_csv`, especially with quoted multiline records and mixed line endings.
- Please strengthen the buffer-boundary test so the boundary crosses a quoted multiline field/CRLF sequence, not only a long simple row.

Once those focused cases are covered and CI stays green, I can do the local Windows/MSVC validation pass.

im-anishraj:
Thanks for adding the targeted reader coverage; the new tests are in the right direction.

I can’t re-review or start local validation on this head yet because the branch now conflicts with the latest `main` after recent CSV parser changes, and GitHub is not reporting checks for the current head. Please rebase/merge `main`, resolve the conflict carefully in the parser path, and then wait for the full required C++/wheel/Python matrix to run green before requesting review again.

im-anishraj (changes requested):
Thanks for rebasing, fixing the MSVC build, and getting the full C++/wheel/Python matrix green. The added `read_csv` skip/nrows and `read_csv_chunked` multiline parity coverage is useful.

One requested coverage item is still missing before I can start the local validation pass: the final diff removed the `scan_csv` assertion from `test_csv_readers_parity_multiline_mixed`, so `scan_csv` is not actually covered against the new `RecordReader` path. Please add a focused `scan_csv` regression assertion/test for the same mixed-line-ending / multiline quoted record style input, checking that the inferred schema remains correct.

Once that is in place and CI stays green, I can run the Windows MSVC/Python 3.12 local gate.

im-anishraj (approved):
Thanks for the follow-up fixes. The final head keeps the patch scoped to the CSV reader, benchmark, and targeted regression tests; the `scan_csv` coverage is now included.

Local validation passed on a clean checkout against current `origin/main` with Windows MSVC + Python 3.12: targeted CSV regression/chunked/core CSV tests, `benchmarks/benchmark_csv.py`, full `tests/`, Black, Ruff, and mypy. The required GitHub C++ native, Windows MSVC wheel smoke, manylinux wheel smoke, macOS arm64 wheel smoke, and full Python matrix are green.

PR #990, opened 2026-05-21
feat(csv): add encoding_errors support

### LINKED ISSUE #129: CSV: Add encoding_errors option
## Context
Users should be able to choose strict, replace, or ignore behavior when transcoding non-UTF-8 input.

## Suggested files
arnio/io.py, cpp/src/csv_reader.cpp, bindings/bind_arnio.cpp, tests/test_csv.py, README.md

## Scope

## Acceptance criteria

## Contributor notes

Code-007-maker:
## Summary

Adds `encoding_errors` support to CSV APIs while preserving backward-compatible strict behavior by default.

Supported modes:

* `strict`
* `replace`
* `ignore`

## Changes

* added `encoding_errors` parameter to `read_csv()` and `scan_csv()`
* validated supported modes in the Python API layer
* threaded the option through `_CsvConfig` and bindings
* added focused UTF-8 handling for invalid byte sequences in CSV string parsing and schema inference
* added regression tests for all supported modes
* updated documentation/examples

Fixes #129

### DIFF
--- a/README.md
+++ b/README.md
@@ -254,7 +254,34 @@ frame = ar.read_csv("data.csv", null_values=["", "MISSING", "UNKNOWN"])
 # Disable null sentinel handling completely
 frame = ar.read_csv("data.csv", null_values=[])
 ```
+### Handling invalid UTF-8 bytes
 
+Use `encoding_errors` to control how invalid UTF-8 bytes are handled during CSV parsing.
+
+```python
+# Raise an error on invalid UTF-8 bytes (default)
+frame = ar.read_csv(
+    "data.csv",
+    encoding_errors="strict",
+)
+
+# Replace invalid bytes with the Unicode replacement character (�)
+frame = ar.read_csv(
+    "data.csv",
+    encoding_errors="replace",
+)
+
+# Ignore invalid bytes completely
+frame = ar.read_csv(
+    "data.csv",
+    encoding_errors="ignore",
+)
+```
+Supported values:
+
+- `"strict"` (default)
+- `"replace"`
+- `"ignore"`
 > Every step above executes in C++. Your Python code is a _configuration_ — not the execution engine.
 
 > Explore more in the **[examples/](./examples/)** folder — ready-to-run recipes for sales, customers, survey, logs, and finance datasets.
--- a/arnio/io.py
+++ b/arnio/io.py
@@ -35,6 +35,7 @@ def _utf8_csv_path(
     encoding: str,
     delimiter: str = ",",
     sample_rows: int | None = None,
+    encoding_errors: str = "strict",
 ) -> Iterator[str]:
     """Return a UTF-8 file path for the C++ reader.
 
@@ -48,7 +49,7 @@ def _utf8_csv_path(
 
     tmp_name: str | None = None
     try:
-        with open(path, encoding=encoding, newline="") as src:
+        with open(path, encoding=encoding, errors=encoding_errors, newline="") as src:
             with tempfile.NamedTemporaryFile(
                 "w", encoding="utf-8", newline="", suffix=".csv", delete=False
             ) as tmp:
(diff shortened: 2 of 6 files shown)


im-anishraj (changes requested):
Thanks for opening this. The scope is close to #129, but I can’t approve this head yet.

Blockers before re-review:

- The assignment explicitly asked for a README/API docs update for the new public `encoding_errors` parameter, but this diff only changes code and tests.
- `encoding_errors="replace"` should match normal decoding semantics by using the Unicode replacement character (U+FFFD), not silently substituting `?`. Please pin the exact replacement behavior in tests.
- Please add coverage for invalid bytes in headers as well as data cells, since column names are exposed back to Python.
- This touches C++/wheel behavior, so the full matrix and required wheel smoke jobs need to finish green before approval.

Once those are addressed, I’ll review the implementation details again.

Code-007-maker:
Thanks for the detailed review. I’ve addressed the requested changes:

- added README/API documentation for `encoding_errors`
- updated replace-mode behavior to use the Unicode replacement character (`�`) instead of `?`
- added focused coverage for invalid UTF-8 bytes in headers as well as data cells
- kept the implementation scoped to the new `encoding_errors` flow only

im-anishraj (changes requested):
Thanks for the quick update. The previous review items are partly addressed, but this head still needs changes before re-review.

Blockers:

- The new README section opens a Python code fence for the `encoding_errors` examples but does not close it before the following prose. Please add the closing backticks so the README renders correctly.
- `handle_utf8_errors` currently retries by treating every non-ASCII byte as invalid after the first decode failure. That corrupts otherwise-valid UTF-8 text if a field contains both valid multibyte characters and one invalid byte, for example `caf\xc3\xa9\xff` would not preserve `é` under `replace` or `ignore`. Please handle invalid UTF-8 incrementally so valid multibyte characters are preserved, and add a focused test for mixed valid Unicode plus invalid bytes.
- The required full CI/wheel matrix is still pending on this C++/wheel-impacting head, so approval has to wait for those jobs to finish green after the fixes.

Once those are addressed, I’ll re-review the implementation again.

Code-007-maker:
Thanks for the detailed review. I’ve addressed the remaining blockers:

* fixed the README code fence so the new `encoding_errors` section renders correctly
* updated UTF-8 handling to preserve valid multibyte UTF-8 sequences while only replacing/ignoring invalid bytes incrementally
* added focused regression coverage for mixed valid Unicode plus invalid bytes
* kept the implementation scoped to the `encoding_errors` behavior only

The updated changes are pushed and CI is running again on the refreshed head.

im-anishraj (changes requested):
Thanks for fixing the README fence, switching replace mode to U+FFFD, adding header coverage, and adding the mixed valid UTF-8 regression. CI is green now, but I found one remaining behavior blocker before approval.

Blockers:

- `handle_utf8_errors(...)` is called from `CsvParser::infer_type(...)`, but the transformed return value is ignored. That means `encoding_errors="ignore"` drops invalid bytes only later for string values; type inference still sees the original invalid bytes. For example, `b"value\n1\xff\n2\n"` with `encoding_errors="ignore"` should decode to `1` / `2` before inference and keep the column numeric, but the current path can infer it as string because inference still sees `1\xff`. Please sanitize parsed fields before null/type inference, or consistently infer and parse from the transformed value, and add a focused test for ignore-mode numeric inference/value preservation.
- The new UTF-8 validator accepts some invalid sequences when the continuation-byte shape looks valid, such as overlong encodings like `b"\xc0\xaf"` and other out-of-range Unicode encodings. Strict mode should reject those, and replace/ignore should handle them the same way as invalid bytes. Please use a complete UTF-8 validator/decoder or add the missing range checks, with at least one regression test for an overlong invalid sequence.

Because this touches C++ parser and wheel behavior, I’ll wait for the refreshed full C++/wheel/Python matrix to go green after the fix before doing the local Windows/MSVC approval pass.

im-anishraj (approved):
Thanks for the follow-up fixes. The remaining UTF-8 handling blockers are addressed: replace mode uses U+FFFD, valid multibyte UTF-8 is preserved around invalid bytes, ignore mode preserves numeric inference, and strict mode rejects overlong invalid sequences.

Local validation against current `main` passed: targeted `tests/test_csv.py`, full pytest, Black, Ruff, and repo-configured mypy. The full CI matrix and required wheel smoke jobs are green as well.

PR #1005, opened 2026-05-21
feat: add ArFrame.schema_summary property

### LINKED ISSUE #224: ArFrame: Add schema_summary property
## Context
Users should be able to inspect column names, dtypes, and nullability in one object.

## Suggested files

## Scope

## Acceptance criteria

## Contributor notes

ariktadas144:
Fixes #224

## Summary

Adds `ArFrame.schema_summary`, a property that returns column names, dtypes,
and nullability in one place, without triggering a pandas round-trip.

```python
for col in frame.schema_summary:
    print(col.name, col.dtype, col.nullable)
# id      int64   False
# email   string  True
# score   float64 True
```

## 

- Added `ColumnSummary` (a `__slots__` class with `name`, `dtype`, `nullable`)
  to `arnio/frame.py`, above `ArFrame`.
- `schema_summary` reads nullability via `column_by_index(i).get_null_mask()`
  directly on the C++ frame, same path used by `preview()`. No pandas
  conversion is triggered.
- `ColumnSummary` is exported from `arnio/__init__.py` and added to `__all__`
  so users can type-annotate against it.

## Files changed

- `arnio/frame.py` - `ColumnSummary` class + `schema_summary` property
- `arnio/__init__.py` - export `ColumnSummary`
- `tests/test_schema_summary.py` - 30 new tests (new file)

## Tests

30 new tests across three groups:
- `TestColumnSummary` - attributes, `__repr__`, `__eq__`, inequality
- `TestSchemaSummaryNormal` - return type, order, dtype accuracy, nullability
  detection, all four dtypes, CSV-sourced frames, no pandas round-trip guarantee
- `TestSchemaSummaryEdgeCases` - single column, zero-row frame, all-null column,
  null at boundary positions, 50-column frame, public namespace import

All 30 new tests pass. Full suite: 1255 passed, 31 skipped, 0 failures.

### DIFF
--- a/arnio/__init__.py
+++ b/arnio/__init__.py
@@ -47,7 +47,7 @@ from .exceptions import (
     TypeCastError,
     UnknownStepError,
 )
-from .frame import ArFrame
+from .frame import ArFrame, ColumnSummary
 from .integrations import ArnioPandasAccessor, register_duckdb
 from .io import (
     read_csv,
@@ -108,6 +108,7 @@ from .schema import (
 __all__ = [
     # Core class
     "ArFrame",
+    "ColumnSummary",
     # I/O
     "read_csv",
     "read_csv_chunked",
--- a/arnio/frame.py
+++ b/arnio/frame.py
@@ -13,6 +13,42 @@ _VALID_DTYPES: frozenset[str] = frozenset(
 )
 
 
+class ColumnSummary:
+    """Schema summary for a single column.
+
+    Attributes
+    ----------
+    name : str
+        Column name.
+    dtype : str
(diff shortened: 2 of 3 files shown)


im-anishraj (approved):
Thanks for the focused implementation. `schema_summary` keeps the API small, preserves column order, uses the C++ null mask directly without a pandas round-trip, and has thorough coverage for normal and edge cases.

Local validation against current `main` passed: targeted `tests/test_schema_summary.py`, full pytest, Black, Ruff, and repo-configured mypy. CI is green as well.

PR #1012, opened 2026-05-21
feat(pipeline): add verbose diagnostics logging

### LINKED ISSUE #102: Feature: Add verbose/progress mode to `pipeline()` for step-by-step timing and diagnostics
## Summary

`pipeline()` runs steps sequentially with no logging, timing, or progress reporting. For long-running pipelines on large datasets, users have no visibility into which step is executing, how long each step takes, or where a failure occurred in a multi-step pipeline.

## Evidence

In [`arnio/pipeline.py` lines 87–112](https://github.com/im-anishraj/arnio/blob/main/arnio/pipeline.py#L87-L112):

```python
result = frame
for step in steps:
    if len(step) == 1:
        name = step[0]
        kwargs = {}
    elif len(step) == 2:
        name, kwargs = step[0], step[1]
    ...
    if name in _STEP_REGISTRY:
        fn = _STEP_REGISTRY[name]
        result = fn(result, **kwargs)
    ...
return result
```

No timing, no logging, no progress indication.

## Proposed API

```python
# Option A: verbose flag
clean = ar.pipeline(frame, steps, verbose=True)
# Output:
# [arnio] Step 1/5: strip_whitespace ... done (0.02s, 1000000 -> 1000000 rows)
# [arnio] Step 2/5: normalize_case ... done (0.15s, 1000000 -> 1000000 rows)
# [arnio] Step 3/5: drop_nulls ... done (0.08s, 1000000 -> 945000 rows)
# [arnio] Step 4/5: drop_duplicates ... done (1.23s, 945000 -> 823000 rows)
# [arnio] Step 5/5: custom_step [Python] ... done (0.45s, 823000 -> 823000 rows)

# Option B: callback for custom logging
def on_step(step_name, step_idx, total, elapsed, rows_before, rows_after):
    ...

ar.pipeline(frame, steps, on_step=on_step)
```

## Why It Matters

- **Debugging**: When a pipeline produces unexpected results, verbose mode shows exactly which step changed the row count.
- **Performance profiling**: Identifies which step is the bottleneck without external tooling.
- **User confidence**: Progress output on large datasets prevents "is this hanging?" anxiety.
- **Python vs C++ visibility**: Clearly labels which steps use the fast C++ path vs the slow Python fallback.

## Implementation Notes

- Use Python's `logging` module at `DEBUG` level so it can be silenced in production.
- When `verbose=True`, set the arnio logger to `INFO` temporarily.
- Measure time with `time.perf_counter()` around each step.
- Report row count changes between steps.

## Acceptance Criteria

- [ ] `verbose=True` parameter on `pipeline()`
- [ ] Each step logs: name, path (C++ vs Python), elapsed time, row count delta
- [ ] Uses `logging` module (not `print`) for proper configurability
- [ ] Tests verify logging output is produced when verbose is enabled

Code-007-maker:
## Summary

Adds optional verbose diagnostics support to `pipeline()`.

### Changes

* add `verbose` parameter to `pipeline()`
* keep default behavior silent/unchanged
* log step index/name
* log execution path
* log elapsed execution time
* log row-count changes
* use the `arnio` logger instead of print
* add focused tests for:

  * disabled-by-default behavior
  * built-in steps
  * custom Python-backed steps
  * row-count change logging
* add a short README usage example

Fixes #102

### DIFF
--- a/README.md
+++ b/README.md
@@ -548,8 +548,8 @@ Enable lightweight pipeline diagnostics with `verbose=True`:
 result = ar.pipeline(
     frame,
     [
-        ar.trim_strings,
-        ar.drop_nulls,
+        ("strip_whitespace",),
+        ("drop_nulls",),
     ],
     verbose=True,
 )
--- a/README.md
+++ b/README.md
@@ -540,6 +540,24 @@ frame.head(0)
 Raises `ValueError` for negative or boolean `n`.
 </details>
 
+### Pipeline verbose diagnostics
+
+Enable lightweight pipeline diagnostics with `verbose=True`:
+
+```python
+result = ar.pipeline(
+    frame,
+    [
+        ar.trim_strings,
+        ar.drop_nulls,
+    ],
+    verbose=True,
+)
+```
+
+This logs step execution order, execution path, elapsed time,
+and row-count changes through the `arnio` logger.
+
 <br>
 
 ---
--- a/arnio/pipeline.py
+++ b/arnio/pipeline.py
@@ -6,6 +6,7 @@ Chained cleaning pipeline.
 from __future__ import annotations
 
 import inspect
+import logging
 import warnings
 from dataclasses import dataclass
 from threading import Lock
@@ -19,6 +20,7 @@ from .convert import from_pandas, to_pandas
 from .exceptions import PipelineStepError, UnknownStepError
 from .frame import ArFrame
 
+logger = logging.getLogger("arnio")
 _BUILTIN_STEP_NAMESPACE = "builtin"
 _STEP_NAMESPACE_SEPARATOR = ":"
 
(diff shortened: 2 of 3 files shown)


im-anishraj (changes requested):
Thanks for opening this. I can’t approve this head yet.

Blocker: the README example added in this PR does not match the actual pipeline API. It uses `ar.trim_strings` / `ar.drop_nulls` callables, and `ar.trim_strings` is not an exported Arnio API. `pipeline()` expects step tuples, for example `[("strip_whitespace",), ("drop_nulls",)]`.

Please update the README example to use the existing tuple-based step syntax, and add the new `verbose` parameter to the `pipeline()` docstring so the public API is documented in the function itself. CI is also still pending, so please wait for the full matrix to finish after the docs fix.

Code-007-maker:
Thanks for the review — I’ve updated the README example to use the existing tuple-based pipeline syntax and added the new `verbose` parameter to the `pipeline()` docstring so the public API is documented directly in the function as well.

im-anishraj (approved):
Thanks for the quick cleanup. I re-reviewed the docs/API/tests, confirmed CI is green, and ran the local Windows MSVC/Python 3.12 validation gate successfully, including the targeted pipeline tests and full suite.

PR #1014, opened 2026-05-21
feat: add schema YAML exporter (schema_to_dict, schema_to_yaml)

### LINKED ISSUE #633: Feature: add schema YAML exporter for checked-in data contracts
## Problem
Arnio currently lacks this product capability: Schema objects should be exportable for review and version control.

## Evidence / audit context
- Issue title: Feature: add schema YAML exporter for checked-in data contracts
- Audit focus: Schema objects should be exportable for review and version control.
- Labels selected: type: feature, area: schema, priority: medium, difficulty: intermediate, status: needs maintainer, size: m

## Suggested fix

## Acceptance criteria

SiddharthRiot:
## Summary
Adds `schema_to_yaml()` and `schema_to_dict()` to export arnio Schema objects as deterministic YAML for version control and data contracts.

## Changes
- `arnio/schema_export.py` core implementation (zero extra dependencies, stdlib only)
- `tests/test_schema_export.py` 27 unit tests (all passing)
- `arnio/__init__.py` public API export

## API
```python
import arnio as ar

schema = ar.scan_csv("data.csv")
yaml_str = ar.schema_to_yaml(schema)             # returns string
ar.schema_to_yaml(schema, path="schema.yaml")    # also writes file
```

Fixes #633


im-anishraj (changes requested):
Thanks for the focused PR. I can’t approve this version yet.

Blockers:
- The tests import `schema_export` by mutating `sys.path` and importing the module as a top-level file. Please test through the package/public API instead, e.g. `import arnio as ar` / `from arnio.schema_export import ...`, so the exported API and package import path are covered.
- `schema_to_dict()` does not preserve important `Schema` contract metadata such as `strict` and `unique`, and it does not clearly reject non-serializable `rules` the way `Schema.to_json()` does. Those should be part of the deterministic export contract or explicitly rejected/documented.
- `Field` values such as `allowed` are sets today, so the current YAML emitter can raise `TypeError` for valid schema fields instead of producing stable output. Please normalize those to deterministic YAML-safe values, for example sorted lists.

CI is still running too, so please wait for the full matrix after the contract/test updates.

SiddharthRiot:
thanks for the review. I’ll update the tests to use the package/public API, preserve the missing schema metadata in the export contract, normalize YAML-unsafe values like sets, and align the serializer behaviour more closely with `Schema.to_json()`. I’ll push the updates once the changes are ready.

im-anishraj (changes requested):
Thanks for pushing the contract fixes and getting CI green. I’m stopping this head before local validation because one earlier blocker is still present: `tests/test_schema_export.py` still loads `arnio/schema_export.py` directly with `importlib.util.spec_from_file_location(...)`.

Please change the tests to exercise the installed/package API instead, for example `import arnio as ar` for the public exports and/or `from arnio.schema_export import schema_to_dict, schema_to_yaml`. This matters because the PR also changes `arnio/__init__.py`, and the test suite should catch package import/export regressions rather than bypassing them with a file-path import.

The schema metadata and set-normalization fixes look directionally good from this quick pass; once the tests use the package API and CI is green, I can run the local validation pass.

SiddharthRiot:
Done tests now use `import arnio as ar` and `from arnio.schema_export import schema_to_dict, schema_to_yaml` instead of the `importlib` file-path loader. Also added `test_public_api_accessible_via_arnio_namespace` to catch any `__init__.py` regressions. All 28 tests passing locally.

im-anishraj (changes requested):
Thanks for switching the tests to the package/public import path; that fixes the import-surface blocker.

One contract blocker still remains before I can run local validation: `Schema.rules` is silently ignored for Schema objects. `Schema.to_json()` explicitly rejects schemas with `rules` because callables are not serializable, and the YAML exporter should follow the same contract instead of dropping that information. Please add an explicit check such as `if getattr(schema, "rules", None): raise ValueError(...)`, plus a focused test that a Schema-like object or real `Schema` with rules is rejected.

Please also add focused coverage for the `strict` / `unique` metadata and a set-valued `allowed` field if possible, so the fixes from the earlier review stay protected.

SiddharthRiot:
Added an explicit `rules` check in `schema_to_dict` if a Schema object has non-empty `rules`, it now raises `ValueError` with a clear message, matching the contract of `Schema.to_json()`. Added three focused tests: `test_schema_with_rules_raises`, `test_strict_and_unique_preserved`, and `test_set_valued_allowed_normalized`. **All 31 tests passing.**

im-anishraj (changes requested):
Thanks for adding the rules check and the strict/unique/set coverage. One serialization-contract blocker remains before local validation: the exporter still does not align with real Schema / Field objects the way Schema.to_json() does. Building field payloads from ars(field) drops private datetime bounds such as _datetime_min / _datetime_max, and values like 
equired_if remain tuples, which the YAML path does not support even though the existing schema JSON path normalizes them.\n\nPlease align the Schema-object path with the existing schema serialization helpers/contracts, and add tests using real r.Schema / r.Field objects for those fields rather than only fake stand-ins. Once that is fixed and CI is green, I can run the local validation pass.

im-anishraj (approved):
Thanks for the follow-up fixes. The latest head now aligns the real `Schema` / `Field` path with the existing schema serialization helpers and keeps the rules rejection explicit.

Validated locally from a clean checkout merged over current `origin/main` using Python 3.12 + MSVC editable install with dev/parquet extras:
- `pytest tests/test_schema_export.py -q` (36 passed)
- full `pytest -q` (1466 passed, 7 skipped)
- `black --check arnio tests examples benchmarks`
- `ruff check arnio tests examples benchmarks`
- `mypy arnio`

CI is green as well, so I’m approving this.

SiddharthRiot:
@im-anishraj could you please add the relevant labels

PR #1015, opened 2026-05-21
perf: move unmodified columns in strip_whitespace and normalize_case

### LINKED ISSUE #37: Performance: Cleaning functions deep-copy all columns even when unmodified
## Description

Every cleaning function (`drop_nulls`, `fill_nulls`, `strip_whitespace`, `normalize_case`, `drop_duplicates`, `rename_columns`, `cast_types`) creates a brand new `Frame` with fully copied columns. For a frame with 20 columns where only 2 are modified, all 20 columns are deep-copied unnecessarily.

The `clone()` method copies all internal vectors. For large frames, this is a major performance bottleneck — especially in pipelines where multiple cleaning steps are chained.

## Current Behavior

```python
ar.pipeline(frame, [
    ("strip_whitespace",),          # copies all columns
    ("normalize_case",),            # copies all columns again
    ("drop_nulls",),                # copies all columns again
])
# 3x full frame copies for a 3-step pipeline
```

## Expected Behavior

Unmodified columns should be shared (moved, not copied) between frames. Only columns that are actually mutated need to be reconstructed.

## Suggested Implementation

In functions like `strip_whitespace` and `normalize_case`, use `std::move` for unmodified columns instead of `clone()`:

```cpp
// Current (line 176 in cleaning.cpp):
new_cols.push_back(src.clone());  // DEEP COPY

// Proposed:
// Use move semantics or shared_ptr for columns not being modified
```

Consider making `Column` internally use `std::shared_ptr` for its data, enabling copy-on-write semantics.

## Acceptance Criteria

- [ ] Benchmark shows measurable improvement on 5+ step pipelines
- [ ] No change to public API
- [ ] All tests pass

**Difficulty:** Advanced  
**Files:** `cpp/src/cleaning.cpp`, `cpp/include/arnio/column.h`

ayushlad9108:
## Summary

Fixes #37

For a frame with N columns where only K are string columns, every call to `strip_whitespace` or `normalize_case` previously deep-copied all N columns. In a 5-step pipeline on a 20-column frame (2 string + 18 int64), that is 5 × 18 unnecessary O(n) copies of numeric data.

## Changes

**`Column::move_clone()`** — new method that moves internal data vectors out of the Column instead of copying. No change to `clone()` or any existing API.

**`Frame::column_mut(idx)`** — new non-const accessor so cleaning functions can call `move_clone()` on columns they are about to discard.

**`strip_whitespace` and `normalize_case`** now:
- Clone the input frame once (unavoidable — source must stay immutable)
- Rebuild modified string columns from scratch as before
- Move unmodified columns via `move_clone()` — O(1) instead of O(n) per column

## Immutability guarantee

The public `const Frame&` input is never mutated. Only the internal clone is consumed. Verified by 8 regression tests.

## Tests

`tests/test_column_move_immutability.py` — 15 tests covering source immutability for both functions, wide frames, and 5-step pipeline immutability.

## Benchmark

```bash
python benchmarks/benchmark_pipeline_clone.py --rows 500000
Reports avg time and peak heap for a 5-step pipeline on a 20-column frame (2 string + 18 int64). Expected ~30-50% improvement where unmodified columns dominate.

### DIFF
--- a/benchmarks/benchmark_pipeline_clone.py
+++ b/benchmarks/benchmark_pipeline_clone.py
@@ -1,0 +1,100 @@
+"""
+Benchmark: pipeline clone overhead — unmodified column move optimization
+=========================================================================
+Measures the wall-clock time and peak Python-heap allocation for a 5-step
+cleaning pipeline on a wide frame (2 string cols + 18 numeric cols).
+
+Before the optimization, every cleaning step deep-copied ALL 20 columns
+even though only the 2 string columns were modified.  After the fix,
+strip_whitespace and normalize_case move unmodified columns in O(1)
+instead of O(n).
+
+Run::
+
+    python benchmarks/benchmark_pipeline_clone.py
+
+Optional flags::
+
+    --rows   N   Number of rows (default: 500_000)
+    --runs   N   Repetitions (default: 5)
+"""
+
+from __future__ import annotations
+
+import argparse
+import time
+import tracemalloc
+
+import numpy as np
+import pandas as pd
+
+import arnio as ar
+from arnio.convert import from_pandas
+
+STEPS = [
+    ("strip_whitespace",),
+    ("normalize_case", {"case_type": "lower"}),
+    ("strip_whitespace",),
+    ("normalize_case", {"case_type": "upper"}),
+    ("strip_whitespace",),
+]
+
+
+def _make_frame(n_rows: int) -> ar.ArFrame:
+    """2 string columns + 18 int64 columns = 20 columns total."""
+    rng = np.random.default_rng(42)
+    df = pd.DataFrame(
+        {
+            "name": [f"  User_{i}  " for i in range(n_rows)],
+            "city": [f"  City_{i % 100}  " for i in range(n_rows)],
+            **{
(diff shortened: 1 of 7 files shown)


im-anishraj:
Thanks for opening this. Because this touches C++ ownership/move semantics, cleaning behavior, and wheel artifacts, I’m going to wait for the full required matrix before reviewing/approving.

Please wait for lint, C++ native tests, the full Python matrix, Windows MSVC wheel smoke, manylinux wheel smoke, and macOS arm64 wheel smoke to finish green. Once that is green, I’ll do a deeper local validation pass against a clean checkout.

ayushlad9108:
@im-anishraj Yes sure!

ayushlad9108:
@im-anishraj Done with the tasks!

im-anishraj (approved):
Thanks for waiting on the full matrix. I reviewed the C++ ownership/move changes and ran a clean Windows MSVC/Python 3.12 validation pass: editable install with dev/parquet extras, targeted immutability tests, benchmark smoke, full pytest, Black, Ruff, and mypy all passed. CI, including the required wheel smoke jobs, is green.

PR #1029, opened 2026-05-22
refactor: replace nested if-else chains with std::visit in Column::push_back

### LINKED ISSUE #91: Refactor: `Column::push_back` uses deeply nested if-else chains - modernize with `std::visit`
## Summary

`Column::push_back()` uses deeply nested if-else chains (4 levels of variant type checking x 4 value type branches = 16 paths) making it difficult to read, maintain, and extend. Modern C++17 offers `std::visit` as a cleaner alternative.

## Evidence

In [`cpp/src/column.cpp` lines 79–126](https://github.com/im-anishraj/arnio/blob/main/cpp/src/column.cpp#L79-L126):

```cpp
void Column::push_back(const CellValue& value) {
    bool is_null_val = std::holds_alternative<std::monostate>(value);
    null_mask_.push_back(is_null_val);

    if (std::holds_alternative<std::vector<std::string>>(data_)) {
        auto& vec = std::get<std::vector<std::string>>(data_);
        if (std::holds_alternative<std::string>(value))
            vec.push_back(std::get<std::string>(value));
        else if (std::holds_alternative<int64_t>(value))
            vec.push_back(std::to_string(std::get<int64_t>(value)));
        // ... 12 more branches
    } else if (std::holds_alternative<std::vector<int64_t>>(data_)) {
        // ... another 4 branches
    }
    // ... continues for float64 and bool
}
```

This pattern is repeated across `push_back`, `at()`, `memory_usage()`, and `clone()`.

## Impact

- **Maintainability**: Adding a new data type (e.g., `datetime`) requires modifying every branch in every function.
- **Correctness**: The implicit fallback values (e.g., pushing `0` for int when value is monostate) are easy to miss in code review.
- **Readability**: 50 lines of nested if-else for a single operation.

## Suggested Fix

Use `std::visit` with a visitor struct:

```cpp
struct PushBackVisitor {
    ColumnData& data;
    
    void operator()(std::monostate) {
        std::visit([](auto& vec) {
            using T = std::decay_t<decltype(vec)>;
            if constexpr (!std::is_same_v<T, std::monostate>)
                vec.push_back(typename T::value_type{});
        }, data);
    }
    
    void operator()(const std::string& val) { /* ... */ }
    void operator()(int64_t val) { /* ... */ }
    // etc.
};
```

This makes the type conversion matrix explicit and easier to audit.

DaQshin:
## Summary

Column::push_back previously used deeply nested if-else chains with std::holds_alternative and std::get to dispatch across every combination of column type and value type. This has been replaced with std::visit and a PushbackVisitor struct that handles each value type in a separate operator() overload, making the dispatch logic easier to follow and extend.

## Linked issue

Fixes #91 

## Type of change
- [ ] Bug fix
- [ ] Feature
- [ ] Performance improvement
- [ ] Documentation
- [ ] Tests
- [x] Refactor
- [ ] CI / packaging

## Area
- [ ] CSV parser / I/O
- [x] C++ cleaning engine
- [ ] Python API / ArFrame
- [ ] Pipeline / custom steps
- [ ] Schema validation
- [ ] pandas interoperability
- [ ] Docs / website
- [ ] Developer tooling

## Testing

- [ ] `make test`
- [ ] `make lint`
- [x] Other:

## Contributor checklist


im-anishraj (changes requested):
Thanks for the refactor. I can't approve this head yet.

Blockers:
- The PR includes generated/local data artifacts that are unrelated to the `Column::push_back` refactor: `benchmarks/auto_clean_10_seed42.csv`, the `benchmarks/benchmark_sparse_nulls*.csv` files, `benchmarks/benchmark_strip_whitespace.csv`, and `sample_messy_data.jsonl`. Please remove those from the branch.
- There are no completed checks on the latest head yet (`mergeStateStatus` is currently unstable). Because this touches the C++ core, the full required CI needs to be green before local validation, including Windows MSVC wheel smoke, manylinux wheel smoke, macOS arm64 wheel smoke, and the full Python matrix.

Once the branch is scoped to the C++ refactor only and CI is green, I can review and run the local validation pass.

DaQshin:
@im-anishraj I have removed the data artifacts in 178c7f3. Could you approve the workflows so CI can run?

im-anishraj (changes requested):
Thanks for removing the unrelated artifacts. I still can't approve this head yet because the visitor refactor changes `Column::push_back` behavior in a way that can break the column invariant.

Blocker:
- When the incoming value is `std::string` and the column storage is `int64`, `float64`, or `bool`, the new visitor does not append anything to the data vector, but `null_mask_` has already grown. The old implementation appended the default value (`0`, `0.0`, or `false`) for those unsupported conversions, so every non-null push kept data size aligned with `null_mask_`. With the new code, `Column::at()` can index past the end of the data vector after such a push.

Please restore the fallback append behavior for string values pushed into numeric/bool columns, or otherwise guarantee that every `push_back()` call appends one data element whenever it appends one null-mask entry. Please also add a focused C++ regression test for this case. I am leaving the workflows unapproved until this correctness issue is fixed.

DaQshin:
@im-anishraj I have restored the default value fallback (0, 0.0, false) in PushbackVisitor::operator()(const std::string&) for string pushes into INT64, FLOAT64, and BOOL columns, fixing the null_mask_ alignment bug. Also added a regression test covering all three cases. 
I have also added cmake_minimum_required(VERSION 3.27) which was missing and preventing the build from configuring.

im-anishraj (changes requested):
Thanks, I can see the fallback append behavior and regression coverage are restored. I still can't approve this head yet.

Blockers:
- `cpp/CMakeLists.txt` adds `cmake_minimum_required(VERSION 3.27)` inside the `cpp/` subdirectory. That is unrelated to the visitor refactor and could raise the required CMake version for contributors/CI. Please remove it unless this PR deliberately needs a project-wide CMake policy change.
- GitHub still shows no checks for the latest head, and this is C++ core behavior, so CI/wheel smoke must run before approval.

After removing the unrelated CMake version bump and getting green checks, I can re-review.

DaQshin:
@im-anishraj — I've removed the `cmake_minimum_required` in `cpp/CMakeLists.txt`.

im-anishraj:
Thanks, the unrelated CMake version bump is removed now. I approved the pending C++/wheel workflows for this head, so please wait for the full CI and wheel-smoke runs to complete. Because this touches C++ core behavior, I can’t approve until those checks are visible and green.

im-anishraj (approved):
Thanks, the previous blockers are resolved: the unrelated artifacts/CMake change are gone, the fallback append behavior is restored with regression coverage, and the full C++ native/wheel CI is green. Approved for review; I’ll still keep a local validation pass as the gate before any merge decision.

im-anishraj:
Hi, quick tracking note.

This PR currently links issue `#123`, but the PR title/scope appears unrelated to that issue. Please update the PR description to use `Fixes #issue_number` or `Closes #issue_number` with the actual assigned issue.

I’m keeping this open, but the correct linked issue is required for review tracking and GSSoC credit.

DaQshin:
Hi, the PR description already links `Fixes #91` which is the correct issue for this refactor. Could you take another look? Let me know if anything else needs to be updated!
Screenshot 2026-05-24 182309

im-anishraj:
Reviewed locally against latest main before merge. The C++ refactor preserves the existing push_back conversion behavior while ensuring string values into numeric/bool columns keep the data vector aligned with the null mask. Local merge simulation and diff whitespace checks passed, and the full GitHub CI/wheel matrix is green. Merging now.

PR #1106, opened 2026-05-22
Add subset parameter to drop_empty_columns

### LINKED ISSUE #1082: drop_empty_columns function missing subset parameter
The `drop_empty_columns` function in `arnio/cleaning.py` checks all columns for empty values but lacks a `subset` parameter to limit checks to specific columns.

**Expected behavior:** Add `subset: list[str] | None = None` parameter similar to other functions in the module.

**File:** arnio/cleaning.py lines 463-511

tmdeveloper007:
Closes #1082

### DIFF
--- a/arnio/cleaning.py
+++ b/arnio/cleaning.py
@@ -460,7 +460,11 @@ def drop_constant_columns(frame: ArFrame) -> ArFrame:
     return from_pandas(df.drop(columns=constant_columns))
 
 
-def drop_empty_columns(frame: ArFrame) -> ArFrame:
+def drop_empty_columns(
+    frame: ArFrame,
+    *,
+    subset: list[str] | None = None,
+) -> ArFrame:
     """Remove columns whose values are entirely null or empty strings.
 
     String values containing only whitespace are treated as empty.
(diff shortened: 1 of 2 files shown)


im-anishraj (changes requested):
Thanks for the PR. I can’t review or approve this head yet because the diff is not scoped to the drop_empty_columns change.

Blockers:

Once the PR only contains the focused `drop_empty_columns` implementation/tests and CI is green, I can review the behavior.

im-anishraj (changes requested):
Thanks for re-requesting review. I’m keeping changes requested because the branch still has scope and coverage blockers.

Remaining blockers:
- `issues_list.md` is still in the diff. Please remove it from this branch; it is unrelated to `drop_empty_columns(subset=...)`.
- Please add focused tests for the new `subset` behavior: default behavior, selecting only specific columns, unknown columns, bare string input, non-string entries, and duplicate entries if you decide to support or reject them.
- Please validate `subset` consistently with the other cleaning APIs instead of iterating arbitrary inputs directly.

CI is green, which is good, but the diff still needs to be scoped and tested before it can be approved.

PR #1109, opened 2026-05-22
Add encoding and bom parameters to write_csv

### LINKED ISSUE #1086: write_csv missing encoding parameter
`write_csv` in `arnio/io.py` always writes UTF-8 with no option to specify different encodings, while `read_csv` supports multiple encodings.

**Expected behavior:** Add `encoding: str = "utf-8"` parameter to `write_csv`.

**File:** arnio/io.py

### LINKED ISSUE #1087: write_csv missing BOM support
`write_csv` in `arnio/io.py` has no option to write UTF-8 with BOM, which is a common requirement for Excel compatibility.

**Expected behavior:** Add `bom: bool = False` parameter that prepends UTF-8 BOM when True.

**File:** arnio/io.py

tmdeveloper007:
Closes #1086

Closes #1087

### DIFF
--- a/arnio/io.py
+++ b/arnio/io.py
@@ -756,6 +756,8 @@ def write_csv(
     delimiter: str = ",",
     write_header: bool = True,
     line_terminator: str = "\n",
+    encoding: str = "utf-8",
+    bom: bool = False,
 ) -> None:
     """Write an ArFrame to a CSV file via C++ backend.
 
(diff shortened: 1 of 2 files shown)


im-anishraj (changes requested):
Thanks for the PR. I can’t review or approve this head yet because the diff is not scoped to the write_csv encoding/BOM change.

Blockers:

Once the PR only contains the focused `write_csv` implementation/tests and CI is green, I can review the behavior.

im-anishraj (changes requested):
Thanks for re-requesting review. I’m keeping changes requested because the implementation still is not end-to-end.

Remaining blockers:
- The new `encoding` and `bom` values are assigned onto the write config, but the native writer path is not shown to support those fields end-to-end.
- Please add focused tests for UTF-8, BOM output, non-UTF encodings if supported, invalid encoding values, and existing write_csv behavior remaining unchanged.

Once the branch is scoped, implemented end-to-end, tested, and CI is green, I’ll review again.

PR #1111, opened 2026-05-22
Add columns parameter to read_jsonl

### LINKED ISSUE #1088: read_jsonl missing columns parameter
`read_jsonl` in `arnio/io.py` does not support specifying expected columns or their order, which can lead to inconsistent column ordering when rows have missing keys.

**Expected behavior:** Add `columns: list[str] | None = None` parameter to enforce column order.

**File:** arnio/io.py

tmdeveloper007:
Closes #1088

### DIFF
--- a/arnio/io.py
+++ b/arnio/io.py
@@ -955,6 +955,7 @@ def read_jsonl(
     *,
     encoding: str = "utf-8",
     nrows: int | None = None,
+    columns: list[str] | None = None,
 ) -> ArFrame:
     """Read a JSON Lines file into an ArFrame.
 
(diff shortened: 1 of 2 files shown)


im-anishraj (changes requested):
Thanks for the PR. I can’t review or approve this head yet because the diff is not scoped to the read_jsonl columns parameter change.

Blockers:

im-anishraj (changes requested):
Thanks for re-requesting review. I’m keeping changes requested because this still needs validation and tests.

Remaining blockers:
- Please validate `columns` consistently: reject a bare string, reject non-string entries, and define/test duplicate handling.
- Missing requested columns are currently silently added as null columns via `reindex`; please make that contract explicit and test it, or raise a clear error.
- Please add focused tests for ordering, missing/extra JSONL keys, invalid `columns`, `nrows=0`, and default behavior.

PR #1112, opened 2026-05-22
Add subset parameter to profile

### LINKED ISSUE #1089: profile function missing subset parameter
`profile` function in `arnio/quality.py` operates on all columns without a `subset` parameter to profile only specific columns.

**Expected behavior:** Add `subset: list[str] | None = None` parameter to `profile`.

**File:** arnio/quality.py

tmdeveloper007:
Closes #1089

### DIFF
--- a/arnio/quality.py
+++ b/arnio/quality.py
@@ -730,6 +730,7 @@ def profile(
     approx_top_values_min_unique: int = 1000,
     approx_top_values_min_ratio: float = 0.2,
     approx_top_values_sample_size: int = 2000,
+    subset: list[str] | None = None,
 ) -> DataQualityReport:
     """Profile data quality for an ArFrame.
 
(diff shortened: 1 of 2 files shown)


im-anishraj (changes requested):
Thanks for the PR. I can’t review or approve this head yet because the diff is not scoped to the profile subset parameter change.

Blockers:

im-anishraj (changes requested):
Thanks for re-requesting review. I’m keeping changes requested because the subset behavior is still only partial.

Remaining blockers:
- `report.columns` is filtered, but report-level metrics such as `column_count`, memory usage, duplicate rows/ratio, suggestions, and score components still come from the full frame.
- Please validate `subset` consistently and add focused tests for default behavior, selected columns, unknown columns, bare string input, non-string entries, and metrics/suggestions matching the selected scope.

PR #1113, opened 2026-05-22
Add columns parameter to compare_profiles

### LINKED ISSUE #1090: compare_profiles missing subset parameter
`compare_profiles` function in `arnio/quality.py` compares all columns without ability to limit comparison to specific columns.

**Expected behavior:** Add `columns: list[str] | None = None` parameter to `compare_profiles`.

**File:** arnio/quality.py

tmdeveloper007:
Closes #1090

### DIFF
--- a/arnio/quality.py
+++ b/arnio/quality.py
@@ -835,6 +835,8 @@ def profile(
 def compare_profiles(
     profile_a: DataQualityReport,
     profile_b: DataQualityReport,
+    *,
+    columns: list[str] | None = None,
 ) -> ProfileComparison:
     """Compare two data-quality profiles for drift.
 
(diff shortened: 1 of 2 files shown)


im-anishraj (changes requested):
Thanks for the PR. I can’t review or approve this head yet because the diff is not scoped to the compare_profiles columns parameter change.

Blockers:

im-anishraj (changes requested):
Thanks for re-requesting review. I’m keeping changes requested because the selected-column contract is not safe yet.

Remaining blockers:
- Please validate `columns`: reject bare strings, non-string entries, unknown columns in either profile, and duplicate entries if unsupported.
- The current implementation intersects requested columns with each profile, which can silently ignore requested columns that are missing from both profiles.
- Please add focused tests for selected comparisons, missing columns in either profile, invalid `columns`, default behavior, and ordering/issue reporting.

PR #1116, opened 2026-05-22
Add dtype validation to Schema.from_json

### LINKED ISSUE #1093: Schema.from_json missing dtype validation
`Schema.from_json` in `arnio/schema.py` does not validate that field dtype values are valid arnio types (int64, float64, string, bool, null).

**Expected behavior:** Add validation that each field's dtype is in the allowed set: `{"int64", "float64", "string", "bool", "null", "datetime", None}`.

**File:** arnio/schema.py

tmdeveloper007:
Closes #1093

### DIFF
--- a/arnio/schema.py
+++ b/arnio/schema.py
@@ -1971,6 +1971,15 @@ def _field_from_json_dict(name: str, payload: Any) -> Field:
             )
         allowed = set(allowed)
 
+    dtype = payload.get("dtype")
+    if dtype is not None:
+        allowed_dtypes = {"int64", "float64", "string", "bool", "null", "datetime"}
+        if dtype not in allowed_dtypes:
+            raise ValueError(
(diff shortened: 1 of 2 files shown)


im-anishraj (changes requested):
Thanks for the PR. I can’t review or approve this head yet because the diff is not scoped to the Schema.from_json dtype validation change.

Blockers:

im-anishraj (changes requested):
Thanks for re-requesting review. I’m keeping changes requested because this needs centralization and tests before we can change schema loading behavior.

Remaining blockers:
- Please add focused tests for valid dtypes, invalid dtypes, missing dtype behavior, and custom/semantic field definitions that should remain supported.
- Please centralize the allowed dtype set with the rest of schema/type validation so it cannot drift from the supported field constructors and conversion behavior.

PR #1117, opened 2026-05-22
Add format string validation to DateTime field

### LINKED ISSUE #1094: DateTime field no format string validation
If an invalid `format` string is passed to `DateTime` field in `arnio/schema.py`, the error only occurs at validation time, not at field creation.

**Expected behavior:** Validate format string at Field creation time by attempting to use it with `pd.to_datetime` on a test value.

**File:** arnio/schema.py

tmdeveloper007:
Closes #1094

### DIFF
--- a/arnio/schema.py
+++ b/arnio/schema.py
@@ -1517,6 +1517,14 @@ def DateTime(
     if format is not None and not isinstance(format, str):
         raise TypeError("DateTime format must be a string or None")
 
+    if format is not None:
+        try:
+            pd.to_datetime("2024-01-01", format=format)
+        except ValueError as exc:
+            raise ValueError(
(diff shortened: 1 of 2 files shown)


im-anishraj (changes requested):
Thanks for the PR. I can’t review or approve this head yet because the diff is not scoped to the DateTime format validation change.

Blockers:

im-anishraj (changes requested):
Thanks for re-requesting review. I’m keeping changes requested because the format validation approach is too narrow.

Remaining blockers:
- Valid formats that do not parse the fixed sample string can be rejected incorrectly, for example formats with time, timezone, or alternative separators.
- Please add focused tests for valid date-only, datetime, timezone, invalid directives/strings, and existing DateTime min/max behavior.

PR #1118, opened 2026-05-22
Add chunk_size parameter to ArFrame.to_dict

### LINKED ISSUE #1095: ArFrame.to_dict no chunked export option
`ArFrame.to_dict` in `arnio/frame.py` materializes the entire frame into memory at once, which can cause issues with very large frames.

**Expected behavior:** Add a `chunk_size: int | None = None` parameter for chunked dictionary export, yielding dictionaries of lists for each chunk.

**File:** arnio/frame.py

tmdeveloper007:
Closes #1095

### DIFF
--- a/arnio/frame.py
+++ b/arnio/frame.py
@@ -296,13 +296,20 @@ class ArFrame:
 
         return from_pandas(df.tail(n))
 
-    def to_dict(self) -> dict[str, list]:
+    def to_dict(self, *, chunk_size: int | None = None) -> dict[str, list] | Iterator[dict[str, list]]:
         """Export the frame as a Python dictionary.
+
+        Parameters
+        ----------
+        chunk_size : int, optional
+            If provided, yields dictionaries of lists for each chunk of rows.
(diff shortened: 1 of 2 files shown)


im-anishraj (changes requested):
Thanks for the PR. I can’t review or approve this head yet because the diff is not scoped to the ArFrame.to_dict chunk_size parameter change.

Blockers:

im-anishraj (changes requested):
Thanks for re-requesting review. I’m keeping changes requested because this public API change needs validation and tests.

Remaining blockers:
- CI has a failing/pending check; this cannot be merge-ready until the full required matrix is green.
- Please validate `chunk_size`: reject zero, negative, bool, and non-integer values.
- Please add focused tests for default `to_dict`, chunked output shape/order, final partial chunk, empty frames, and invalid `chunk_size`.
- Please consider whether changing `to_dict` to sometimes return an iterator is the right API shape; the contract should be explicit and tested.

PR #1124, opened 2026-05-22
Add write_jsonl function

### LINKED ISSUE #1100: Missing write_jsonl function
`read_jsonl` exists in `arnio/io.py` but there is no corresponding `write_jsonl` function.

**Expected behavior:** Add `write_jsonl(frame: ArFrame, path: str | os.PathLike[str], *, encoding: str = "utf-8") -> None` function.

**File:** arnio/io.py

tmdeveloper007:
Closes #1100

### DIFF
--- a/arnio/__init__.py
+++ b/arnio/__init__.py
@@ -62,6 +62,7 @@ from .io import (
     scan_csv,
     sniff_delimiter,
     write_csv,
+    write_jsonl,
     write_parquet,
 )
 from .pipeline import (
--- a/arnio/__init__.py
+++ b/arnio/__init__.py
@@ -26,6 +26,7 @@ from .cleaning import (
     fill_nulls,
     filter_rows,
     keep_rows_with_nulls,
+    merge,
     normalize_case,
     normalize_unicode,
     parse_bool_strings,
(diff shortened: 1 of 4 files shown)


im-anishraj (changes requested):
Thanks for the PR. I can’t review or approve this head yet because the diff is not scoped to the write_jsonl function change.

Blockers:

im-anishraj (changes requested):
Thanks for re-requesting review. I’m keeping changes requested because the branch is still stacked with unrelated feature code.

Remaining blockers:
- `issues_list.md` is still in the diff and needs to be removed.
- The diff includes unrelated `merge`, `sort`, `sample`, and `concat` implementations/imports. Please keep this PR focused on `write_jsonl` only.
- CI has a failing/pending check; this cannot be merge-ready until the required checks are green.
- Please add focused tests for `.jsonl`/`.ndjson`, path-like inputs, custom encoding, round-trip with `read_jsonl`, invalid extensions, write failures, empty frames, and JSON null/non-serializable value behavior.
- Please keep the writer exception contract clear; using `JsonlReadError` for write failures is confusing unless that is intentionally documented and tested.

PR #1130, opened 2026-05-22
Add subset parameter to drop_constant_columns

### LINKED ISSUE #1081: drop_constant_columns function missing subset parameter
The  function in  checks all columns for constant values but lacks a  parameter to limit checks to specific columns.

**Expected behavior:** Add  parameter similar to other functions in the module.

**File:** arnio/cleaning.py lines 425-460

tmdeveloper007:
Closes #1081

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -238,4 +238,4 @@ _arnio_cpp*.dll
 
 # csv
 /*.csv
-benchmarks/benchmark_gil_temp.csv
+benchmarks/benchmark_gil_temp.csvissues_list.md
--- a/arnio/cleaning.py
+++ b/arnio/cleaning.py
@@ -426,7 +426,11 @@ def drop_duplicates(
     return ArFrame(result)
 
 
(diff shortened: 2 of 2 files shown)


im-anishraj (changes requested):
Thanks for cleaning this into a much more focused PR. I still can’t approve this head yet.

Blockers:
- Please remove the accidental .gitignore change. It currently turns enchmarks/benchmark_gil_temp.csv into enchmarks/benchmark_gil_temp.csvissues_list.md, which corrupts the ignore rule and is unrelated to #1081.
- CI lint is failing with F841 because 
emaining_columns is assigned but never used in rnio/cleaning.py.
- Please add focused tests for drop_constant_columns(subset=...): only subset columns should be considered for dropping, non-subset constant columns should remain, subset=None should preserve current behavior, and unknown subset columns should raise the intended error.

After those are fixed, please wait for the full CI suite to pass before re-requesting review.

PR #1131, opened 2026-05-22
Add subset parameter to drop_empty_columns

### LINKED ISSUE #1082: drop_empty_columns function missing subset parameter

tmdeveloper007:
Closes #1082

### DIFF
--- a/arnio/cleaning.py
+++ b/arnio/cleaning.py
@@ -464,7 +464,11 @@ def drop_constant_columns(frame: ArFrame) -> ArFrame:
     return from_pandas(df.drop(columns=constant_columns))
 
 
-def drop_empty_columns(frame: ArFrame) -> ArFrame:
+def drop_empty_columns(
+    frame: ArFrame,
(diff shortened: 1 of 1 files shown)


im-anishraj (changes requested):
Thanks for narrowing this into a focused PR. The implementation direction is reasonable, but I can’t approve it without coverage for the new public contract.

Blocker:
- Please add focused tests for drop_empty_columns(subset=...): only subset columns should be considered for dropping, non-subset empty columns should remain, subset=None should preserve current behavior, and unknown subset columns should raise the intended error.

After those tests are added, please wait for the full CI suite to pass before re-requesting review.

PR #1132, opened 2026-05-22
Add encoding_errors and dtype parameters to read_csv_chunked

### LINKED ISSUE #1083: read_csv_chunked missing encoding_errors parameter
`read_csv_chunked` in `arnio/io.py` does not support the `encoding_errors` parameter that `read_csv` has, creating an API inconsistency.

**Expected behavior:** Add `encoding_errors: str = "strict"` parameter to `read_csv_chunked` to match `read_csv`.

**File:** arnio/io.py

### LINKED ISSUE #1084: read_csv_chunked missing dtype parameter
`read_csv_chunked` in `arnio/io.py` does not support the `dtype` parameter for specifying column types, which `read_csv` supports.

**Expected behavior:** Add `dtype: dict[str, str] | None = None` parameter to `read_csv_chunked`.

**File:** arnio/io.py

tmdeveloper007:
Closes #1083

Closes #1084

### DIFF
--- a/arnio/io.py
+++ b/arnio/io.py
@@ -611,10 +611,12 @@ def read_csv_chunked(
     nrows: int | None = None,
     skip_rows: int = 0,
     encoding: str = "utf-8",
+    encoding_errors: str = "strict",
     trim_headers: bool = True,
     decimal_separator: str = ".",
     thousands_separator: str | None = None,
     null_values: list[str] | None = None,
+    dtype: dict[str, str] | None = None,
     mode: str = "strict",
     on_bad_lines: str = "error",
 ) -> Iterator[ArFrame]:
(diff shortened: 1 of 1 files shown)


im-anishraj (changes requested):
Thanks for the focused PR. I can’t approve this head yet because the new public parameters are currently only added to the signature/docstring; they are not wired into behavior.

Blockers:
- ncoding_errors must be applied to the chunked read/transcoding path, with validation consistent with 
ead_csv.
- dtype must actually affect the yielded chunks, or the PR should not add the parameter. Please match the existing 
ead_csv contract as closely as possible.
- Add focused tests for both parameters: an encoding-error handling case for 
ead_csv_chunked, dtype enforcement/coercion across yielded chunks, and invalid/unknown dtype behavior if applicable.
- Wait for the full CI suite to pass after the implementation and tests are added.

As written, callers could pass these parameters and see no effect, which would be a misleading API.

PR #1133, opened 2026-05-22
Add usecols parameter to scan_csv

### LINKED ISSUE #1085: scan_csv missing usecols parameter
`scan_csv` in `arnio/io.py` does not support the `usecols` parameter to scan only specific columns from a CSV file.

**Expected behavior:** Add `usecols: list[str] | None = None` parameter to `scan_csv`.

**File:** arnio/io.py

tmdeveloper007:
Closes #1085

### DIFF
--- a/arnio/io.py
+++ b/arnio/io.py
@@ -832,6 +832,7 @@ def scan_csv(
     null_values: list[str] | None = None,
     has_header: bool = True,
     encoding_errors: str = "strict",
+    usecols: list[str] | None = None,
 ) -> dict[str, str]:
     """Return schema (column names + inferred types) without loading data.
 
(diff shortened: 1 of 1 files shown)


im-anishraj (changes requested):
Thanks for the focused PR. I can't approve this head yet because the new scan_csv(usecols=...) contract needs coverage and one edge case should be made explicit.

Blockers:
- Please add focused tests for scan_csv(usecols=...): returning only the requested columns, preserving requested column order, usecols=None preserving current behavior, and unknown requested columns raising the intended error.
- Please avoid silently dropping unknown requested columns in the final schema filtering. If the native reader already raises for unknown usecols, keep that behavior covered by a test and make the Python-side filtering explicit instead of using if col in schema.
- Wait for the full CI suite to pass after the tests/fix land.

The implementation direction is reasonable, but without these tests callers could get a partial schema instead of a clear error.

PR #1134, opened 2026-05-22
Add encoding and bom parameters to write_csv

### LINKED ISSUE #1086: write_csv missing encoding parameter

### LINKED ISSUE #1087: write_csv missing BOM support

tmdeveloper007:
Closes #1086

Closes #1087

### DIFF
--- a/arnio/io.py
+++ b/arnio/io.py
@@ -756,6 +756,8 @@ def write_csv(
     delimiter: str = ",",
     write_header: bool = True,
     line_terminator: str = "\n",
+    encoding: str = "utf-8",
+    bom: bool = False,
 ) -> None:
     """Write an ArFrame to a CSV file via C++ backend.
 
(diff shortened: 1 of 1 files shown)


im-anishraj (changes requested):
Thanks for the focused PR. I can't approve this head yet because the parameters are not implemented end-to-end.

Blockers:
- CI lint/mypy is failing because CsvWriteConfig has no encoding or bom attributes.
- The native CsvWriteConfig/CsvWriter path does not currently expose or use these fields, so adding assignments in Python will not make write_csv write a different encoding or prepend a BOM.
- Please implement encoding and BOM support through the actual writer path, including bindings/config updates if needed.
- Add focused tests that assert the bytes written for UTF-8 BOM, a non-UTF-8 encoding case, default behavior, invalid encoding handling, and bom type validation.
- Because this touches the CSV writer/native surface, wait for the full CI suite and wheel smoke checks to pass after the fix.

PR #1136, opened 2026-05-22
Add columns parameter to read_jsonl

### LINKED ISSUE #1088: read_jsonl missing columns parameter

tmdeveloper007:
Closes #1088

### DIFF
--- a/arnio/io.py
+++ b/arnio/io.py
@@ -955,6 +955,7 @@ def read_jsonl(
     *,
     encoding: str = "utf-8",
     nrows: int | None = None,
+    columns: list[str] | None = None,
 ) -> ArFrame:
     """Read a JSON Lines file into an ArFrame.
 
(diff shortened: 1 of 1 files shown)


im-anishraj (changes requested):
Thanks for the focused PR. The basic direction is reasonable, but I can't approve a new public read_jsonl(columns=...) parameter without validation and coverage.

Blockers:
- Please validate columns clearly: reject a bare string and reject non-string entries rather than letting pandas interpret them indirectly.
- Add focused tests for requested column order, missing requested columns being filled with nulls, extra JSON keys being omitted when columns is provided, columns=None preserving current behavior, invalid columns input, and nrows interaction.
- Please decide and test the nrows=0 behavior with columns. If columns is provided, returning an empty frame with those requested columns is usually the least surprising contract; if you choose otherwise, document it explicitly.
- Wait for the full CI suite to pass after the tests and validation are added.

PR #1137, opened 2026-05-22
Add subset parameter to profile

### LINKED ISSUE #1089: profile function missing subset parameter

tmdeveloper007:
Closes #1089

### DIFF
--- a/arnio/quality.py
+++ b/arnio/quality.py
@@ -730,6 +730,7 @@ def profile(
     approx_top_values_min_unique: int = 1000,
     approx_top_values_min_ratio: float = 0.2,
     approx_top_values_sample_size: int = 2000,
+    subset: list[str] | None = None,
 ) -> DataQualityReport:
     """Profile data quality for an ArFrame.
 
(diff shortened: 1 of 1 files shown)


im-anishraj (changes requested):
Thanks for the focused PR. I can't approve this yet because the subset behavior is only partially implemented and has no tests.

Blockers:
- If profile(subset=...) means profile only those columns, the report-level metrics should be consistent with that subset. Right now column_count, memory usage, duplicate counts/ratio, and suggestions are still based on the full frame while columns only contains the subset.
- Please validate subset clearly: reject a bare string, reject non-string entries, and define/test duplicate handling.
- Add focused tests for subset=None preserving current behavior, subset selecting only requested columns, unknown columns raising, invalid subset input, and report metrics/suggestions matching the chosen contract.
- Wait for the full CI suite to pass after the implementation and tests are added.

PR #1138, opened 2026-05-22
Add columns parameter to compare_profiles

### LINKED ISSUE #1090: compare_profiles missing subset parameter

tmdeveloper007:
Closes #1090

### DIFF
--- a/arnio/quality.py
+++ b/arnio/quality.py
@@ -835,6 +835,8 @@ def profile(
 def compare_profiles(
     profile_a: DataQualityReport,
     profile_b: DataQualityReport,
+    *,
+    columns: list[str] | None = None,
 ) -> ProfileComparison:
     """Compare two data-quality profiles for drift.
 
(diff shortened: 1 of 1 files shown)


im-anishraj (changes requested):
Thanks for the focused PR. I can't approve this head yet because columns is not validated and can silently produce an empty or partial comparison.

Blockers:
- Please validate columns clearly: reject a bare string, reject non-string entries, and raise for requested columns missing from either profile instead of silently intersecting them away.
- The docstring says columns=None compares all shared columns, but the existing behavior requires matching schemas and this PR preserves that. Please keep the documentation aligned with the actual default contract.
- Add focused tests for columns=None preserving current behavior, comparing only requested columns while ignoring unrelated schema differences, missing requested columns raising, invalid columns input, and status_counts/drift_report only covering requested columns.
- Wait for the full CI suite to pass after the fix and tests land.

PR #1139, opened 2026-05-22
Add exclude_columns parameter to check_quality_gates

### LINKED ISSUE #1091: check_quality_gates missing exclude_columns parameter
`check_quality_gates` in `arnio/quality.py` has no way to exclude specific columns from drift detection (e.g., metadata columns that always change).

**Expected behavior:** Add `exclude_columns: list[str] | None = None` parameter.

**File:** arnio/quality.py

tmdeveloper007:
Closes #1091

### DIFF
--- a/arnio/quality.py
+++ b/arnio/quality.py
@@ -912,6 +912,7 @@ def check_quality_gates(
     allow_new_columns: bool = False,
     allow_missing_columns: bool = False,
     fail_on_dtype_change: bool = True,
+    exclude_columns: list[str] | None = None,
 ) -> QualityGateResult:
     """Check two quality profiles against pass/fail drift thresholds.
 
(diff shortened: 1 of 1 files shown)


im-anishraj (changes requested):
Thanks for the focused PR. I can't approve this without validation and tests for the new gate contract.

Blockers:
- Please validate exclude_columns clearly: reject a bare string and reject non-string entries instead of treating them as iterables.
- Add focused tests showing excluded columns suppress missing/new/dtype/null/numeric drift issues for those columns, while non-excluded columns still fail normally.
- Add tests for exclude_columns=None preserving current behavior, unknown excluded columns behavior, and invalid exclude_columns input.
- Consider including the effective exclude_columns value in the returned thresholds/result metadata so audit output explains why a column was ignored, or document why it is intentionally omitted.
- Wait for the full CI suite to pass after the tests and validation are added.

PR #1140, opened 2026-05-22
Add columns parameter to validate

### LINKED ISSUE #1092: validate function missing subset parameter
`validate` function in `arnio/schema.py` validates all columns without option to validate only specific columns.

**Expected behavior:** Add `columns: list[str] | None = None` parameter to `validate`.

**File:** arnio/schema.py

tmdeveloper007:
Closes #1092

### DIFF
--- a/arnio/schema.py
+++ b/arnio/schema.py
@@ -830,6 +830,8 @@ def validate(
     frame: ArFrame,
     schema: Schema | dict[str, Field],
     max_errors: int | None = None,
+    *,
+    columns: list[str] | None = None,
 ) -> ValidationResult:
     """Validate an ArFrame against a schema.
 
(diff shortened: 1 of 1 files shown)


im-anishraj (changes requested):
Thanks for the focused PR. I can't approve this implementation yet because filtering the schema fields by constructing a new Schema changes more than the requested columns.

Blockers:
- Schema(schema_fields) drops the original schema metadata such as strict mode, composite unique settings, rules, and other options, so validate(columns=...) can silently change validation behavior.
- In strict schemas, validating only selected columns should not turn every unselected frame column into an unexpected_column issue. The selected-column contract needs to be explicit and tested.
- Please validate columns clearly: reject a bare string, reject non-string entries, and raise for requested columns that are not present in the schema or frame as appropriate.
- Add focused tests for columns=None preserving current behavior, validating only requested fields, strict-mode behavior, composite unique/rules interaction, unknown requested columns, invalid columns input, and max_errors interaction if applicable.
- Wait for the full CI suite to pass after the implementation and tests land.

PR #1141, opened 2026-05-22
Add dtype validation to Schema.from_json

### LINKED ISSUE #1093: Schema.from_json missing dtype validation

tmdeveloper007:
Closes #1093

### DIFF
--- a/arnio/schema.py
+++ b/arnio/schema.py
@@ -1971,6 +1971,15 @@ def _field_from_json_dict(name: str, payload: Any) -> Field:
             )
         allowed = set(allowed)
 
+    dtype = payload.get("dtype")
+    if dtype is not None:
+        allowed_dtypes = {"int64", "float64", "string", "bool", "null", "datetime"}
+        if dtype not in allowed_dtypes:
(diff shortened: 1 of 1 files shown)


im-anishraj (changes requested):
Thanks for the focused PR. The implementation shape is reasonable, but I can't approve it without regression coverage for the new deserialization validation.

Blockers:
- Please add focused Schema.from_json tests showing valid dtype values round-trip, dtype=None is accepted, an invalid dtype raises a clear ValueError, and a non-string dtype value is handled clearly.
- Please keep the allowed dtype set aligned with the rest of the schema/field dtype contract, preferably by reusing or centralizing the list if there is already a local helper/constant.
- Wait for the full CI suite to pass after the tests are added.

PR #1142, opened 2026-05-22
Add format string validation to DateTime field

### LINKED ISSUE #1094: DateTime field no format string validation

tmdeveloper007:
Closes #1094

### DIFF
--- a/arnio/schema.py
+++ b/arnio/schema.py
@@ -1517,6 +1517,14 @@ def DateTime(
     if format is not None and not isinstance(format, str):
         raise TypeError("DateTime format must be a string or None")
 
+    if format is not None:
+        try:
+            pd.to_datetime("2024-01-01", format=format)
+        except ValueError as exc:
+            raise ValueError(
+                f"DateTime format {format!r} is not a valid strftime format: {exc}"
+            ) from exc
+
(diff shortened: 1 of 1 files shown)


im-anishraj (changes requested):
Thanks for the focused PR. I can't approve this implementation because it rejects valid existing datetime formats.

Blockers:
- The validation parses the fixed string 2024-01-01 with the user format. Valid formats that include time or timezone tokens, such as %Y-%m-%dT%H:%M:%S and %Y-%m-%dT%H:%M:%S%z, will fail at field creation even though the existing tests and validation path support them.
- Please validate the format without assuming every valid format can parse a date-only sample. A safer approach is to generate a representative datetime using strftime(format), then parse that generated value with the same format, while preserving the existing TypeError for non-string formats.
- Add focused tests for valid date-only, datetime-with-time, and timezone-aware formats, plus invalid format strings, and make sure existing DateTime validation tests continue to pass.
- Wait for the full CI suite to pass after the fix and tests land.

PR #1143, opened 2026-05-22
Add chunk_size parameter to ArFrame.to_dict

### LINKED ISSUE #1095: ArFrame.to_dict no chunked export option

tmdeveloper007:
Closes #1095

### DIFF
--- a/arnio/frame.py
+++ b/arnio/frame.py
@@ -6,6 +6,7 @@ ArFrame — the core data container wrapping the C++ Frame.
 from __future__ import annotations
 
 import json
+from typing import Iterator
 
 from ._core import _Frame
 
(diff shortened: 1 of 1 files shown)


im-anishraj (changes requested):
Thanks for the focused PR. I can't approve this yet because the new chunked export contract is missing validation and tests.

Blockers:
- Please validate chunk_size clearly: reject bools and non-integers, and raise ValueError for zero or negative values. As written, chunk_size=0 will fail through range(), and negative values silently yield no chunks.
- Add focused tests for default to_dict() preserving current behavior, chunked output across multiple chunks, final partial chunk, empty frames, and invalid chunk_size values.
- Please run/let CI finish after adding those tests; this PR is currently still unstable.

PR #1144, opened 2026-05-22
Add merge function for combining two ArFrame objects

### LINKED ISSUE #1096: Missing join/merge functionality
There is no functionality to combine two frames horizontally (join/merge operation) in arnio.

**Expected behavior:** Add `merge(frame1: ArFrame, frame2: ArFrame, *, on: str | list[str], how: str = "left") -> ArFrame` function.

tmdeveloper007:
Closes #1096

### DIFF
--- a/arnio/__init__.py
+++ b/arnio/__init__.py
@@ -26,6 +26,7 @@ from .cleaning import (
     fill_nulls,
     filter_rows,
     keep_rows_with_nulls,
+    merge,
     normalize_case,
     normalize_unicode,
     parse_bool_strings,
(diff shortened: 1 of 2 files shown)


im-anishraj (changes requested):
Thanks for the focused PR. I can't approve this new public merge API without validation/export cleanup and tests.

Blockers:
- Please add focused tests for left/inner/outer behavior, multi-key joins, missing merge keys on either side, invalid how, and invalid on input such as an empty list or non-string entries.
- Please validate on clearly before passing it to pandas. A bare string is fine, but list members should all be strings and an empty list should raise a clear error.
- Please add merge to arnio.__all__ so the public export surface stays consistent with the import.
- Please document or test the behavior for overlapping non-key column names, since pandas will suffix them by default.

PR #1145, opened 2026-05-22
Add sort function for sorting frames by column values

### LINKED ISSUE #1097: Missing sort function
There is no function to sort frames by column values in arnio.

**Expected behavior:** Add `sort(frame: ArFrame, *, by: str | list[str], ascending: bool | list[bool] = True) -> ArFrame` function.

tmdeveloper007:
Closes #1097

### DIFF
--- a/arnio/__init__.py
+++ b/arnio/__init__.py
@@ -34,6 +34,7 @@ from .cleaning import (
     round_numeric_columns,
     safe_divide_columns,
     select_columns,
+    sort,
     standardize_missing_tokens,
     strip_whitespace,
     trim_column_names,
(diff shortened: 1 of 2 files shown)


im-anishraj (changes requested):
Thanks for the focused PR. I can't approve this new public sort API yet because it needs validation, tests, and public export cleanup.

Blockers:
- CI lint is failing, so please address that first.
- Please add focused tests for single-column sorting, multi-column sorting, ascending as a bool, ascending as a list, missing sort columns, invalid by input such as an empty list or non-string entries, and invalid ascending values/list length.
- Please validate by and ascending explicitly rather than relying on pandas errors. A bare string for by is fine, but list members should be strings and an empty list should raise a clear error. ascending should be bool or a list of bools matching by.
- Please add sort to arnio.__all__ so the public export surface stays consistent with the import.

PR #1146, opened 2026-05-22
Add sample function for randomly sampling rows

### LINKED ISSUE #1098: Missing sample function
There is no function to randomly sample rows from a frame in arnio.

**Expected behavior:** Add `sample(frame: ArFrame, *, n: int | None = None, frac: float | None = None, seed: int | None = None) -> ArFrame` function.

tmdeveloper007:
Closes #1098


im-anishraj (changes requested):
Thanks for the focused PR. I can't approve this head yet.

Blockers:
- CI lint is failing, so please fix formatting first.
- Please remove the unrelated sort implementation and export from this PR. #1146 should only implement sample for #1098.
- Please add focused tests for n-based sampling, frac-based sampling, deterministic seed behavior, n and frac being mutually exclusive, neither n nor frac provided, invalid n/frac/seed values, and behavior when n exceeds the row count.
- Please validate inputs explicitly: n should be a positive integer when provided, frac should be a positive number in the supported range, and seed should be an int or None.
- Please add sample to arnio.__all__ so the public export surface is consistent with the import.

PR #1147, opened 2026-05-22
Add concat function for vertically stacking frames

### LINKED ISSUE #1099: Missing concat function for vertical stacking
There is no function to stack frames vertically in arnio (like pandas `concat`).

**Expected behavior:** Add `concat(frames: list[ArFrame], *, ignore_index: bool = False) -> ArFrame` function.

tmdeveloper007:
Closes #1099

### DIFF
--- a/arnio/__init__.py
+++ b/arnio/__init__.py
@@ -17,6 +17,7 @@ from .cleaning import (
     clip_numeric,
     coalesce_columns,
     combine_columns,
+    concat,
     drop_columns,
     drop_columns_matching,
     drop_constant_columns,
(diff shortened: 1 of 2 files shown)


im-anishraj (changes requested):
Thanks for the PR. I’m going to request changes for now because this needs a little more hardening before it can be accepted.

Blocking items:
- Please add focused tests for `concat`, including normal vertical stacking, `ignore_index=True`, empty input, a single-frame input, invalid entries in `frames`, and mismatched column/schema behavior.
- Please add `concat` to `arnio.__all__` so the public API export is complete.
- Please validate the public API inputs, especially that `frames` is a non-empty sequence of `ArFrame` objects and that `ignore_index` is a real bool.
- Please explicitly define and test mismatched-column behavior. If this is intended to union columns like `pandas.concat`, the null-fill behavior should be covered; if the API should require matching schemas, it should raise a clear error.

Once those are addressed and CI is green, I’ll take another look.

PR #1148, opened 2026-05-22
Add write_jsonl function

### LINKED ISSUE #1100: Missing write_jsonl function

tmdeveloper007:
Closes #1100

### DIFF
--- a/arnio/__init__.py
+++ b/arnio/__init__.py
@@ -58,6 +58,7 @@ from .io import (
     scan_csv,
     sniff_delimiter,
     write_csv,
+    write_jsonl,
     write_parquet,
 )
 from .pipeline import (
--- a/arnio/io.py
+++ b/arnio/io.py
@@ -1286,3 +1286,49 @@ def write_parquet(
         kwargs["row_group_size"] = row_group_size
 
     df.to_parquet(path, **kwargs)
+
+def write_jsonl(
+    frame: ArFrame,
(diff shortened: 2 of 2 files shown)


im-anishraj (changes requested):
Thanks for the PR. I’m going to request changes for now because the writer needs tests and a few API details tightened up before it is safe to merge.

Blocking items:
- Please add `write_jsonl` to `arnio.__all__` so it is exported consistently with the other I/O helpers.
- Please add focused tests, ideally covering `.jsonl` and `.ndjson`, path-like inputs, custom encoding, round-trip with `read_jsonl`, invalid extensions, write failures, and empty frames.
- Please make the JSON serialization robust. `df.to_dict(...)+json.dumps(...)` can emit invalid JSON for NaN values and can fail on pandas/NumPy scalar or timestamp values. Missing values should round-trip as JSON `null`, and non-JSON-serializable values should have a clear tested behavior.
- Please keep the exception contract clear. The docstring says OSError may be raised, but the implementation wraps write failures in `JsonlReadError`, which is confusing for a writer.

PR #1149, opened 2026-05-22
Add ffill and bfill method support to fill_nulls

### LINKED ISSUE #1101: fill_nulls missing forward/backward fill support
`fill_nulls` in `arnio/cleaning.py` only supports filling with a scalar value, not forward-fill or backward-fill.

**Expected behavior:** Add `method: str | None = None` parameter to `fill_nulls` with values "ffill", "bfill", or None (current behavior).

**File:** arnio/cleaning.py

tmdeveloper007:
Closes #1101

### DIFF
--- a/arnio/cleaning.py
+++ b/arnio/cleaning.py
@@ -335,31 +335,36 @@ def select_columns(frame: ArFrame, columns: Sequence[str]) -> ArFrame:
 
 def fill_nulls(
     frame: ArFrame,
-    value: Any,
+    value: Any = None,
     *,
     subset: list[str] | None = None,
+    method: str | None = None,
 ) -> ArFrame:
-    """Replace null/empty values with a given fill value.
+    """Replace null/empty values with a given fill value or forward/backward fill.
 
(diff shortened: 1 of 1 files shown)


im-anishraj (changes requested):
Thanks for the PR. I’m going to request changes because the new method path does not yet preserve the existing API semantics.

Blocking items:
- Please add focused tests for `method="ffill"` and `method="bfill"`, including subset behavior, invalid method values, providing both `value` and `method`, and preserving existing scalar-fill behavior.
- The `method` branch currently ignores `subset`; it calls `df.ffill()`/`df.bfill()` on the whole frame, so columns outside the requested subset can be changed.
- Changing `value` from required to default `None` now makes `fill_nulls(frame, None)` raise. If `None` should no longer be a valid fill value, please make that explicit and test it; otherwise use a sentinel so the existing scalar-fill contract is preserved.
- Please keep the implementation aligned with the pipeline/API docs if the signature changes.

PR #1150, opened 2026-05-22
Improve strict mode error message to list all unexpected columns

### LINKED ISSUE #1102: Schema.strict error message lacks column details
When `strict=True` in `Schema.validate` and unexpected columns exist, the error message does not list which columns are unexpected.

**Expected behavior:** Include the list of unexpected column names in the issue message.

**File:** arnio/schema.py

tmdeveloper007:
Closes #1102

### DIFF
--- a/arnio/schema.py
+++ b/arnio/schema.py
@@ -949,32 +949,16 @@ def validate(
 
     if schema.strict:
         expected = set(schema.fields)
+        unexpected_columns = [name for name in df.columns if name not in expected]
 
-        for name in df.columns:
-            if name not in expected:
-                issues.append(
-                    ValidationIssue(
-                        column=str(name),
-                        rule="unexpected_column",
-                        message=f"Unexpected column: {name}",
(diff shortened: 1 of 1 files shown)


im-anishraj (changes requested):
Thanks for the PR. I’m going to request changes for now because this changes validation behavior and needs targeted coverage.

Blocking items:
- Please add tests for strict mode with multiple unexpected columns, asserting that the message lists all unexpected columns.
- Please also cover interaction with `max_errors`; the aggregated unexpected-column issue should not cause the validator to exceed the requested error limit when earlier validation issues already reached it.
- Please keep the issue contract clear. Collapsing several unexpected columns into one issue may be fine, but the chosen `column` value and `issue_count` behavior should be intentional and tested so downstream callers do not get a surprise.
- CI is still pending; this cannot be considered merge-ready until the full required matrix is green.

Once those are covered and CI is green, I’ll take another look.

PR #1151, opened 2026-05-22
Add distinct function to get unique values from a column

### LINKED ISSUE #1103: Missing distinct function to get unique values
`drop_duplicates` removes duplicate rows but there is no function to get unique values from a column.

**Expected behavior:** Add `distinct(frame: ArFrame, *, column: str) -> list` function to return unique values from a column.

tmdeveloper007:
Closes #1103

### DIFF
--- a/arnio/__init__.py
+++ b/arnio/__init__.py
@@ -23,6 +23,7 @@ from .cleaning import (
     drop_duplicates,
     drop_empty_columns,
     drop_nulls,
+    distinct,
     fill_nulls,
     filter_rows,
     keep_rows_with_nulls,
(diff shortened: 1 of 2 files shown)


im-anishraj (changes requested):
Thanks for opening the cleaned-up `distinct` PR. This is much closer than the earlier stacked branch, but I’m going to request changes before it can be accepted.

Blocking items:
- Please add focused tests for `distinct`, including normal unique values, ordering expectations, missing columns, empty frames, and null-containing columns.
- Please add `distinct` to `arnio.__all__` so the public API export is complete.
- Please define and test the null behavior. The current implementation silently drops nulls with `dropna()`, which may be right, but the contract should be intentional and covered.

PR #1152, opened 2026-05-22
Add dtype validation to cast_types function

### LINKED ISSUE #1104: Missing cast_types errors='coerce' error details
When `errors="coerce"` fails to cast values in `cast_types`, the error message does not indicate which specific values failed.

**Expected behavior:** Track and report specific failed values in the `TypeCastError` exception when `errors="coerce"`.

**File:** arnio/cleaning.py

tmdeveloper007:
Closes #1104

### DIFF
--- a/arnio/cleaning.py
+++ b/arnio/cleaning.py
@@ -1053,6 +1053,14 @@ def cast_types(
     if errors not in {"raise", "coerce"}:
         raise ValueError("errors must be either 'raise' or 'coerce'")
 
+    allowed_dtypes = {"int64", "float64", "string", "bool", "null"}
+    for col, dtype in mapping.items():
+        if dtype not in allowed_dtypes:
+            raise ValueError(
+                f"Invalid dtype {dtype!r} for column {col!r}. "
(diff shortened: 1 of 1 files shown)


im-anishraj (changes requested):
Thanks for the PR. I’m going to request changes because this currently changes existing error behavior and needs tests.

Blocking items:
- Please add focused tests for invalid dtype strings, valid dtype strings, non-string mapping values, and non-mapping `mapping` inputs.
- The new validation runs before `_validate_string_mapping`, so inputs like a list of pairs can now fail with an attribute error instead of the existing clear TypeError. Please keep the existing mapping validation first.
- This changes the invalid-dtype exception from the existing `TypeCastError` path to `ValueError`. If that is intentional, it needs to be discussed and covered; otherwise keep the existing public exception contract.
- Please make sure the allowed dtype list is centralized/aligned with the actual C++ cast support so it cannot drift from what `_cast_types` accepts.
- CI is still pending, so this cannot be considered merge-ready yet.

PR #1161, opened 2026-05-22
fix: add missing json import to io.py

### LINKED ISSUE #1156: Add missing import json to io.py
The  module is used in  at line 1032 but is not imported at the top of . It should be added to the imports at the top of the file.

**Location**:

tmdeveloper007:
Closes #1156

Adds the missing  import to .

### DIFF
--- a/arnio/io.py
+++ b/arnio/io.py
@@ -6,6 +6,7 @@ CSV reading and writing functions.
 from __future__ import annotations
 
 import io
+import json
 import os
 import shutil
 import tempfile


im-anishraj (changes requested):
Thanks for the PR. I can't approve this because the linked issue was closed as invalid: `arnio/io.py` already imports `json` locally where it is needed in `read_jsonl`, so there is no missing import bug to fix.

This branch also has failing lint. Please close this PR, or reopen with a demonstrated runtime failure and a targeted fix if there is a different issue.

PR #1162, opened 2026-05-22
fix: add missing json import to quality.py

### LINKED ISSUE #1157: Add missing import json to quality.py
The  module is used in  (line 272) for  but is not imported at the top of the file. It should be added to the imports.

tmdeveloper007:
Closes #1157

Adds the missing  import to .

### DIFF
--- a/arnio/quality.py
+++ b/arnio/quality.py
@@ -6,6 +6,7 @@ Data quality profiling and safe automatic cleaning helpers.
 from __future__ import annotations
 
 import html
+import json
 import math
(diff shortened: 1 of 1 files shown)


im-anishraj (changes requested):
Thanks for the PR. I can't approve this because the linked issue was closed as invalid: `arnio/quality.py` already imports `json` locally where serialization logic needs it, so there is no missing import bug here.

This branch also has failing lint. Please close this PR, or reopen with a demonstrated runtime failure and a targeted fix if there is a different issue.

PR #1163, opened 2026-05-22
fix: move Counter import to top-level in io.py

### LINKED ISSUE #1158: Move  to top-level import in io.py
In  line 1166,  is a deferred import inside the function . This should be moved to the top-level imports at the beginning of the file for consistency and performance.

tmdeveloper007:
Closes #1158

Moves the deferred  import to the top-level imports in  for consistency.

### DIFF
--- a/arnio/io.py
+++ b/arnio/io.py
@@ -10,6 +10,7 @@ import os
 import shutil
 import tempfile
 import warnings
+from collections import Counter
 from collections.abc import Iterator, Sequence
 from contextlib import contextmanager
 from typing import cast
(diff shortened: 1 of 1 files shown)


im-anishraj (changes requested):
Thanks for the PR. I can't approve this as-is because the linked issue was closed as invalid/not planned: moving `Counter` from the local usage site to a top-level import is not a bug fix and does not give the project a clear functional improvement.

Even though CI is green, the change is import-only churn, so please close this PR unless there is a demonstrated runtime or maintainability issue that needs a different targeted fix.

PR #1168, opened 2026-05-22
Add exclude_columns support to profile() quality reports

### LINKED ISSUE #1153: Feature: add exclude_columns parameter to profile()
### Problem or motivation

`profile()` currently analyzes every column in the frame. In production datasets, users often need to omit columns that are not useful for quality profiling, such as identifier columns, raw payload/text columns, temporary metadata fields, or columns validated elsewhere.

The current workaround is to manually drop those columns before profiling. That is less ergonomic and can create unnecessary intermediate frames when the user only wants to exclude columns from the quality report.

### Proposed solution

Add an `exclude_columns` parameter to `profile()`:

```python
report = ar.profile(frame, exclude_columns=["user_id", "raw_payload"])

Excluded columns should not appear in report.columns and should be ignored by profile-derived outputs such as summaries, markdown/html exports, warnings, suggestions, and quality score calculations.

Validation should follow existing column-selection patterns:

reject missing excluded columns with a clear error
reject string input such as exclude_columns="id"
reject non-string items in the sequence
allow exclude_columns=None as the current default behaviour


### User story


```md
As a user profiling large real-world datasets, I want to exclude identifier, raw payload, or internal metadata columns from `profile()` so that the quality report focuses only on columns relevant to data cleanup and analysis.

### Acceptance criteria

- [ ] `profile(frame, exclude_columns=[...])` excludes those columns from `report.columns`
- [ ] excluded columns are ignored by warnings, suggestions, markdown/html exports, and quality score calculations
- [ ] `exclude_columns=None` preserves current behavior
- [ ] missing excluded columns raise a clear error
- [ ] string input and non-string items are rejected consistently
- [ ] focused tests cover valid exclusion, missing columns, invalid input, and unchanged default behavior
- [ ] existing quality/profile tests continue to pass

### Alternatives considered

Users can manually call `drop_columns()` before `profile()`, but that changes the input frame workflow and requires an extra transformation step just to control report scope.

### Scope

Data profiling / quality engine

### Expected difficulty

Needs maintainer review

Yes - I'd like to submit a PR

### Pre-flight checks

### Additional context

_No response_

anweshabhattacharyya:
Fixes #1153

## Summary
Adds `exclude_columns` support to `profile()` by defining one effective profiling scope and using it consistently across the quality report.

## Changes
- Added `exclude_columns` parameter to `profile()`
- Validates missing columns, bare string input, and non-string entries
- Applies the scoped columns to:
  - `report.columns`
  - report-level metrics
  - duplicate calculations
  - suggestions
  - markdown/html exports
  - quality score calculations

## Tests
Added focused coverage for:
- default behavior
- valid exclusions
- missing excluded columns
- bare string input
- non-string entries
- report-level metrics/suggestions matching exclusion scope

## Validation
- `ruff check arnio/quality.py tests/test_quality.py`
- `pytest tests/test_quality.py tests/test_schema.py tests/test_pipeline.py -q`

### DIFF
--- a/arnio/quality.py
+++ b/arnio/quality.py
@@ -7,13 +7,20 @@ from __future__ import annotations
 
 import html
 import math
+from collections.abc import Sequence
 from dataclasses import dataclass, field
 from typing import Any
 
 import numpy as np
 import pandas as pd
 
-from .cleaning import cast_types, drop_duplicates, strip_whitespace
+from .cleaning import (
+    _validate_column_sequence,
+    cast_types,
+    drop_duplicates,
+    strip_whitespace,
+    validate_columns_exist,
+)
 from .convert import to_pandas
 from .frame import ArFrame
 
@@ -725,6 +732,7 @@ class QualityGateResult:
 def profile(
     frame: ArFrame,
     *,
+    exclude_columns: Sequence[str] | None = None,
     sample_size: int = 5,
     approx_top_values: bool = False,
     approx_top_values_min_unique: int = 1000,
@@ -737,6 +745,8 @@ def profile(
     ----------
     frame : ArFrame
         Input frame to inspect.
+    exclude_columns : Sequence[str], optional
+        Column names to exclude from profiling.
     sample_size : int, default 5
         Number of non-null sample values to keep per column.
     approx_top_values : bool, default False
@@ -782,11 +792,30 @@ def profile(
         approx_top_values_sample_size, bool
     ):
         raise TypeError("approx_top_values_sample_size must be an integer")
+
     if approx_top_values_sample_size <= 0:
         raise ValueError("approx_top_values_sample_size must be positive")
 
+    has_exclusions = exclude_columns is not None and len(exclude_columns) > 0
+
+    if exclude_columns is not None:
+        exclude_columns = _validate_column_sequence(
+            exclude_columns,
+            argument_name="exclude_columns",
+        )
+        validate_columns_exist(
+            frame,
+            exclude_columns,
+            operation="profile",
+        )
+
     df = to_pandas(frame)
-    row_count, column_count = frame.shape
+
+    if has_exclusions:
+        df = df.drop(columns=list(exclude_columns))
+
+    row_count = len(df)
+    column_count = len(df.columns)
     duplicate_rows = int(df.duplicated().sum()) if row_count else 0
     duplicate_ratio = _ratio(duplicate_rows, row_count)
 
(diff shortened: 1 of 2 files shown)


im-anishraj (changes requested):
Thanks for the focused PR. The overall shape is close: `report.columns`, column count, suggestions, exports, and validation are all moving in the right direction.

One blocker before approval: `memory_usage` is still calculated from the original full frame, so one report-level metric does not follow the effective `exclude_columns` scope. I verified this locally with a large excluded column: `profile(frame, exclude_columns=["drop"]).memory_usage` stayed equal to the full-frame report, while a keep-only frame was much smaller.

Please calculate `memory_usage` from the scoped profiling data and add a regression test where excluding a large/raw column changes the reported memory usage. CI is also still pending, so I’ll wait for the refreshed checks after that fix.

im-anishraj (changes requested):
Thanks for the quick update. The scoped-memory test is the right direction, and the excluded-column case now drops the large column from the reported memory usage.

There is still one blocker: this changes the default behavior for `profile(frame).memory_usage` even when `exclude_columns=None`. Previously `profile(frame).memory_usage` came from `frame.memory_usage()`. With this patch it comes from `df.memory_usage(deep=True).sum()`, which can report a different value for the same unscoped report. I verified locally that the full report changed from `frame.memory_usage() == 30659` to `profile(frame).memory_usage == 30183` on the same frame.

Please preserve the existing `frame.memory_usage()` behavior when no columns are excluded, and add a regression test asserting `ar.profile(frame).memory_usage == frame.memory_usage()` for the default case. For the excluded case, the metric should still follow the scoped report.

anweshabhattacharyya:
Thanks for the detailed check. I’ve updated the implementation so the default `profile(frame)` path preserves `frame.memory_usage()`, while excluded-column reports use scoped memory usage. I also added regression tests for both the default and excluded-column cases. Local validation is now passing.

im-anishraj (approved):
Approved. The previous memory-usage blocker is resolved: default `profile(frame)` preserves `frame.memory_usage()`, while excluded-column reports use scoped memory usage.

Local validation on Windows with the MSVC developer environment and Python 3.12 passed:
- `pip install -e ".[dev,parquet]"`
- `pytest -q`
- `black --check .`
- `ruff check .`
- `mypy arnio`

GitHub Actions are green and the PR is clean against current main.

anweshabhattacharyya:
Thanks for the detailed validation and review. Glad the memory-usage behavior is now aligned with the expected contract and that the checks passed across environments.

Since this is part of GSSoC, could you also please add the `gssoc:approved` label for tracking once convenient?

PR #1245, opened 2026-05-23
test : added edge case tests for to_arrow conversion in convert.py

tmdeveloper007:
Closes #1235.

Summary of What Has Been Done:
Added a new test file tests/test_to_arrow_edge_cases.py with 12 test cases for to_arrow edge cases in arnio/convert.py.

Changes Made:
New file: tests/test_to_arrow_edge_cases.py

The test suite covers:

Frame Configurations:
- Zero-column frame produces empty table
- All-null column handling
- Single row frame conversion
- Mixed dtype columns

Data Type Preservation:
- String dtype preservation
- Integer dtype preservation
- Boolean dtype preservation
- Float dtype with NaN handling

Edge Cases:
- Empty string values
- Special characters (newline, tab, quote)
- Negative numbers
- Large integer values

Impact it Made:
Ensures Arrow conversion handles all data types and edge cases correctly, providing reliable interoperability with Arrow-based ecosystems.

PR #1251, opened 2026-05-23
test: cover to_arrow edge cases

### LINKED ISSUE #1235: test : add edge case tests for to_arrow conversion in convert.py
Summary of What Needs to be Done:
The to_arrow function in arnio/convert.py converts an ArFrame to an Arrow Table. While basic tests exist in test_convert.py, edge cases such as zero-column frames, all-null columns, and specific dtypes are not thoroughly tested. These edge cases can silently produce incorrect Arrow output.

Changes that Need to be Made:
Add edge case tests to test_convert.py for to_arrow:
- Zero-column frame produces empty Arrow table
- Frame with all-null column produces correct Arrow null array
- Frame with string dtype preserves data correctly
- Frame with int dtype preserves data correctly
- Frame with float dtype handles NaN correctly
- Single-row frame produces valid Arrow table

Example:
def test_zero_column_frame():
    df = ArFrame()
    table = to_arrow(df)
    assert table.num_columns == 0
    assert table.num_rows == 0

def test_all_null_column():
    df = ArFrame({"val": [None, None, None]})
    table = to_arrow(df)
    assert table.num_columns == 1
    col = table.column(0)
    assert col.null_count == 3

def test_handles_nan_in_float_column():
    df = ArFrame({"val": [1.0, float('nan'), 3.0]})
    table = to_arrow(df)
    # Should not raise, should produce valid arrow table

Impact that it would Provide:
Ensures Arrow conversion handles all data types and edge cases correctly, providing reliable interoperability with Arrow-based ecosystems.

saurabhhhcodes:
## Summary
- add to_arrow coverage for zero-column empty frames and all-null typed columns
- assert Arrow schema, column counts, row counts, null counts, and NaN-vs-null behavior

## Validation
- .venv/bin/python -m pytest tests/test_convert.py::TestToArrow -q
- .venv/bin/python -m pytest tests/test_convert.py -q
- git diff --check

Fixes #1235

### DIFF
--- a/tests/test_convert.py
+++ b/tests/test_convert.py
@@ -1004,6 +1004,51 @@ class TestToArrow:
         assert table.num_rows == 0
         assert table.num_columns == 1
 
+    def test_zero_column_empty_frame_preserves_empty_schema(self):
+        import pyarrow
+
+        frame = ar.from_pandas(pd.DataFrame(index=pd.RangeIndex(0)))
+        table = ar.to_arrow(frame)
+
+        assert table.schema == pyarrow.schema([])
+        assert table.num_columns == 0
+        assert table.num_rows == 0
+
+    def test_all_null_columns_preserve_schema_and_null_counts(self):
+        import pyarrow
+
+        df = pd.DataFrame(
+            {
+                "id": pd.Series([pd.NA, pd.NA, pd.NA], dtype=pd.Int64Dtype()),
+                "name": pd.Series([pd.NA, pd.NA, pd.NA], dtype="string"),
+                "active": pd.Series([pd.NA, pd.NA, pd.NA], dtype="boolean"),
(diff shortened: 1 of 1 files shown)


saurabhhhcodes:
Closing this to avoid duplicating the already-active implementation path in #1245, as noted on the issue thread. I missed that new maintainer reply while opening this PR. Thanks for the guidance.

im-anishraj:
Thanks, this PR is scoped well and the focused local check passed on Windows/Python 3.12:

```powershell
python -m pytest tests/test_convert.py::TestToArrow -q
```

Result: 15 passed.

I’m going to wait for the full GitHub matrix to finish before approval, since several Python jobs are still queued/in progress on the latest head.

PR #1275, opened 2026-05-23
Fix parse bool subset skip bool columns

### LINKED ISSUE #1178: Bug: parse_bool_strings unnecessarily reparses existing bool columns when subset is provided
### What happened?

`parse_bool_strings()` currently applies boolean string parsing/conversion logic to columns explicitly listed in `subset`, even when those columns are already typed as `bool`.

This creates unnecessary normalization/conversion work and can produce behavior differences compared to the automatic column-selection path used when `subset=None`.

The issue appears to be limited to explicit subset handling.

This looks like a small API consistency bug rather than an intended behavior difference.

### Area

Data profiling / quality reports

### Minimal reproduction

```python
import pandas as pd
import arnio as ar

df = pd.DataFrame({
    "flag": [True, False, True],
})

frame = ar.from_pandas(df)

result = ar.parse_bool_strings(
    frame,
    subset=["flag"],
)

print(ar.to_pandas(result))
```

### Expected behavior

Columns that are already typed as `bool` should be skipped during boolean string parsing, even when explicitly included in `subset`.

This would keep subset behavior consistent with the automatic column-selection path and avoid unnecessary parsing/conversion work.

### Actual behavior

The explicitly selected boolean column still goes through the parsing/conversion path even though it is already typed as `bool`.

This creates unnecessary processing and inconsistent behavior between:

- `subset=None`
- explicit `subset=[...]`

### Impact

Produces incorrect data

### Arnio version

1.1.0

### Install method

Built from source (editable install with dev dependencies)

### Operating System

Windows

### Python version

3.12

### Environment details

```shell
Local source checkout with editable install and development dependencies.

Validated against current main after recent DataQualityReport changes.
```

### Pre-flight checks

### Additional context

This appears to be a narrow API consistency issue.

A focused fix could likely:
- skip already-boolean columns before parsing logic runs
- preserve existing behavior for object/string columns
- add targeted regression coverage for subset handling

I’d be happy to work on a focused PR for this if the behavior change is considered acceptable.

Code-007-maker:
Fixes #1178

## Summary

Skip already-boolean columns in the explicit
`parse_bool_strings(..., subset=[...])` path.

This keeps behavior consistent with the automatic
`subset=None` inference path and avoids unnecessary parsing work.

## Changes

- skip existing bool columns during subset handling
- preserve existing behavior for object/string columns
- add focused regression coverage

## Validation

```bash
pytest tests/test_cleaning.py -k parse_bool_strings -v


im-anishraj (approved):
Thanks, this is a focused fix for the explicit-subset bool-column case. It preserves existing bool columns, keeps object/string parsing behavior unchanged, includes a regression test, and CI is green. Approving.

Code-007-maker:
Thanks for the review and approval!  
I’ve addressed the requested scope, CI is green, and the branch is now up to date with `main`.

When you get a chance, could you please take a final look for merge? Thanks again for the guidance throughout the review process.

im-anishraj:
Thanks for the follow-up. This PR is still approved, green, and mergeable from review. No additional code changes are needed from you right now; the final merge is just waiting on the maintainer merge gate/local validation queue.

im-anishraj:
Reviewed and verified this PR before merge.

Checks completed:
- confirmed the PR is approved and linked to the tracked issue
- reviewed the changed files and scope
- simulated the PR head merged onto the latest `main` locally
- ran local diff, formatting, static, and syntax checks for the touched files
- confirmed the GitHub Actions checks are passing

Thanks for the contribution. Merging now.

PR #1285, opened 2026-05-23
fix(security): prevent yaml injection via unescaped newlines in _emit…

### LINKED ISSUE #1283: [BUG] Critical YAML Injection Vulnerability in schema_to_yaml()
### Describe the Bug
There is a critical YAML Injection vulnerability in `arnio`'s deterministic YAML serializer (`schema_export.py`).

In `_emit_scalar()`, the logic attempts to determine if a string needs to be quoted to produce safe YAML:
```python
        needs_quoting = (
            not value
            or value.lower() in {"true", "false", "null", "yes", "no", "on", "off"}
            or value[0] in ('"', "'", "{", "[", "|", ">", "!", "&", "*", "#", "?", ":", "-", ",", "@", "$", "%", "~")
            or ":" in value
            or "#" in value
            or value != value.strip()
        )
```

This quoting logic fails to account for embedded newline characters (`\n`) that don't trigger `strip()`. If a string value contains a newline (e.g., `value = "valid\n- malicious_item"`), `needs_quoting` evaluates to `False`. The string is written directly into the output without quotes or escaping.

Because the YAML exporter uses blind string formatting to assemble structures:
```python
lines.append(f"{indent}- {rendered}")
```
An attacker who controls a column name or metadata in a dataframe can inject a payload like `safe\n- malicious_payload` or `safe\nmalicious_key: true`. This causes any downstream YAML parser reading the exported file to process the injected payload as a completely new top-level array item or nested configuration flag. If `arnio`'s exported schemas are used in data pipelines (e.g., CI/CD schema contracts), this allows attackers to silently manipulate infrastructure configurations and bypass validation checks.

### Proposed Solution
I can fix this critical security flaw by properly sanitizing newlines:
1. Update the `needs_quoting` condition in `schema_export.py` to correctly flag `\n` and `\r`.
2. Ensure that any string containing multiline characters gets cleanly escaped and wrapped in double quotes.
3. Add robust unit tests specifically checking for YAML payload injection to prevent regression.

Could you please assign this issue to me under GSSoC '24 and consider adding the `level:critical` / `security` labels so I can open a PR to patch this immediately?

basantnema31:
## Security Patch: Prevent YAML Injection via Unescaped Newlines

Fixes #1283
**Type:** Security / High Priority

### Threat Model & Impact
Prior to this patch, the `_emit_scalar()` function in `schema_export.py` failed to sanitize newline (`\n`) and carriage return (`\r`) characters embedded within string values. 

If an application using `arnio` serialized untrusted input containing these characters, an attacker could artificially terminate the current YAML node and successfully inject malicious sibling keys or items into the generated YAML document. This structural manipulation is a classic **YAML Injection** vulnerability, potentially leading to privilege escalation, configuration poisoning, or denial-of-service depending on how the downstream YAML is consumed.

### Technical Resolution
As requested by the maintainers, this patch is **tightly scoped** to neutralize the vulnerability directly at the serialization boundary in `_emit_scalar()` without impacting broader architecture:

1. **Extended Quote Evaluation:** Expanded the boolean `needs_quoting` matrix to proactively detect multiline boundaries using `any(c in value for c in ('\n', '\r'))`.
2. **Safe Structural Escaping:** If a newline boundary is detected, the string is now forced into a double-quoted block. Crucially, the literal `\n` and `\r` characters are transformed into safe, two-character escape sequences (`\\n`, `\\r`).
3. **Outcome:** By escaping the characters inside double-quotes, the YAML parser interprets them strictly as string content, making it mathematically impossible for the payload to break out of the string boundary and inject new YAML structural nodes.

### Validation & Regression Testing
To guarantee this vulnerability cannot regress in the future, a dedicated security regression test (`test_emit_scalar_prevents_yaml_injection`) has been implemented:
- **Payload 1 (`\n`):** Validates that inputs like `safe\nmalicious_key: true` are cleanly escaped as `"safe\\nmalicious_key: true"`.
- **Payload 2 (`\r`):** Validates the same structural safety for carriage returns.
- **Test Suite Status:** All 37 tests within `test_schema_export.py` pass cleanly, confirming this fix introduces zero regressions to the existing exporter logic.

### DIFF
--- a/arnio/schema_export.py
+++ b/arnio/schema_export.py
@@ -51,6 +51,7 @@ def _emit_scalar(value: Any) -> str:
         needs_quoting = (
             not value
             or value.lower() in {"true", "false", "null", "yes", "no", "on", "off"}
+            or any(c in value for c in ("\n", "\r"))
             or value[0]
             in (
                 '"',
@@ -78,7 +79,7 @@ def _emit_scalar(value: Any) -> str:
         )
         if needs_quoting:
             # Use double-quote style; escape backslashes and double-quotes.
-            escaped = value.replace("\\", "\\\\").replace('"', '\\"')
+            escaped = value.replace("\\", "\\\\").replace('"', '\\"').replace("\n", "\\n").replace("\r", "\\r")
             return f'"{escaped}"'
         return value
     raise TypeError(f"Unsupported scalar type: {type(value)!r}")
--- a/arnio/schema_export.py
+++ b/arnio/schema_export.py
@@ -79,7 +79,12 @@ def _emit_scalar(value: Any) -> str:
         )
         if needs_quoting:
             # Use double-quote style; escape backslashes and double-quotes.
-            escaped = value.replace("\\", "\\\\").replace('"', '\\"').replace("\n", "\\n").replace("\r", "\\r")
+            escaped = (
+                value.replace("\\", "\\\\")
+                .replace('"', '\\"')
+                .replace("\n", "\\n")
+                .replace("\r", "\\r")
+            )
             return f'"{escaped}"'
         return value
     raise TypeError(f"Unsupported scalar type: {type(value)!r}")
--- a/tests/test_schema_export.py
+++ b/tests/test_schema_export.py
@@ -157,6 +157,17 @@ class TestSchemaToYamlOutput:
         assert "ts:" in out
         assert "TIMESTAMP" in out
 
+    def test_emit_scalar_prevents_yaml_injection(self):
+        # test \n
+        raw1 = {"key": {"type": "STRING", "description": "safe\nmalicious_key: true"}}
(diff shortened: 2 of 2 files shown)


im-anishraj (approved):
Thanks, this is tightly scoped to _emit_scalar(), escapes raw newline/CR characters before YAML emission, and includes focused regression coverage. The full CI matrix is green. Approved; please update the PR body to use Fixes #1283 instead of Resolves: #1283 so GitHub auto-closes the linked issue on merge.

im-anishraj:
Reviewed and verified this PR before merge.

Checks completed:
- confirmed the PR is approved, clean, and linked to the tracked issue
- reviewed the changed files and scope
- ran local diff, formatting, static, and syntax checks for the touched files

Thanks for the contribution. Merging now.

PR #1286, opened 2026-05-23
Fix 924 csv chunk schema

### LINKED ISSUE #924: bug: CsvChunkReader locks schema on first chunk, causing silent data corruption in subsequent chunks
### What happened?

When reading large CSVs in chunks using `ar.read_csv(chunksize=...)`, `CsvChunkReader` locks the inferred column types based solely on the data present in the very first chunk. 

If a column has a more restricted type (like `int64` or all empty/null values) in the first chunk, but contains values of a wider type (like `float64` or `string`) in subsequent chunks, those wider values fail to parse under the locked type and are silently loaded as `null` (represented as `std::monostate` in C++ and `NaN` in Python). This results in severe silent data corruption/loss.

### Area

C++ CSV parsing engine (`CsvChunkReader`)

### Minimal reproduction

```python
import pandas as pd
import arnio as ar

# Create a CSV where column 'val' has integers in the first chunk but floats in the second chunk
df = pd.DataFrame({"val": [1, 2, 3.5, 4.8]})
df.to_csv("test_chunks.csv", index=False)

# Read CSV with a chunksize of 2
reader = ar.read_csv("test_chunks.csv", chunksize=2)
chunks = list(reader)

print("First Chunk:")
print(ar.to_pandas(chunks[0]))

print("\\nSecond Chunk (Expect 3.5 and 4.8, but got NaN due to silent conversion to Null):")
print(ar.to_pandas(chunks[1]))
```

### Expected behavior

If a wider/different data type is encountered in subsequent chunks:
* Either the reader should dynamically promote the column's dtype (e.g. from `INT64` to `FLOAT64` or `STRING`),
* Or the reader should utilize the global inferred schema from a pre-scan (if `sample_size` was specified),
* Or it should raise a clear type mismatch/parsing error to alert the user rather than silently dropping/corrupting data.

### Suggestion / Root Cause

In `cpp/src/csv_reader.cpp`, `schema_locked_` is set to `true` after the first chunk:
```cpp
    if (!schema_locked_) {
        for (const auto& row : raw_data) {
            for (size_t ci : col_indices_) {
                if (ci < row.size()) {
                    DType inferred = parser_.infer_type(row[ci]);
                    col_types_[ci] = CsvParser::promote_type(col_types_[ci], inferred);
                }
            }
        }
        for (auto& dt : col_types_) {
            if (dt == DType::NULL_TYPE) dt = DType::STRING;
        }
        schema_locked_ = true;
    }
```
For all subsequent chunks, `schema_locked_` remains `true` so `col_types_` is never updated. When `parser_.parse_value` is called for the subsequent chunks with the locked dtype, values that cannot be parsed as that dtype (e.g. `3.5` as `INT64`) trigger an exception and return `std::monostate{}` (null), silently corrupting the data.

Arryan21-06:
Fixes #924

## Type of Change
- [x] Bug fix 
- [x] C++ parser area

## Summary
When reading large CSVs in chunks, the C++ engine previously locked column types based on the first chunk. Subsequent chunks containing wider data types (like floats in an integer column) were failing to parse and silently outputting Nulls. 

This PR implements the agreed-upon fail-fast approach. `CsvChunkReader` now throws a clear `std::runtime_error` when a chunk encounters a type mismatch, preventing silent data corruption while safely preserving genuine null values.

**Scope Note:** Pre-scan promotion behavior via `sample_size` is intentionally scoped out of this specific fail-fast patch, as per the issue discussion on #924.

## Tests Ran Locally
Added 3 new regression tests to `tests/test_csv_chunked.py` and ran them locally via `pytest`:
- `test_late_mixed_types_raises_error`: Verifies the exception is correctly thrown.
- `test_valid_null_handling_preserved`: Ensures genuine nulls in later chunks do not trigger the error.
- `test_multiple_chunk_boundaries`: Ensures the engine correctly tracks boundaries across >2 chunks.


im-anishraj (changes requested):
Thanks for opening the PR. The direction matches the agreed fail-fast approach for #924, but I can’t approve this head yet.

Blockers:
- CI lint is already red on this head. The new `tests/test_csv_chunked.py` block has formatting/trailing-whitespace issues in the diff, so please run Black/formatting locally and push the formatted result.
- Please fill in the PR body with `Fixes #924`, the bug-fix checkbox/area, and the exact tests you ran. Right now the template is still blank, so the PR will not auto-close the issue and it is hard to audit the validation scope.
- Please either add the planned sample-size/pre-scan regression test or explicitly note in the PR body that pre-scan promotion behavior is intentionally out of scope for this fail-fast patch.

Once those are fixed and the full C++/wheel CI is green, I can re-review the actual implementation.

im-anishraj (changes requested):
Thanks for the formatting fix. I still can't approve this head yet because the PR body is still blank, there is no `Fixes #924` closing reference/testing summary, and the sample-size/pre-scan coverage or explicit out-of-scope note is still missing. Please address those remaining blockers and let the full C++/wheel CI finish green before re-requesting review.

im-anishraj (changes requested):
Thanks for the update. I still can't approve this head yet.

Blockers:
- The PR body is still the untouched template: please add `Fixes #924`, mark the bug-fix / C++ parser area, and include the exact local tests you ran.
- The sample-size/pre-scan coverage or explicit out-of-scope note requested on #924 is still missing from the PR body.
- CI is not green on this head. The Windows Python 3.11 job failed during checkout (`could not read Username for 'https://github.com'`), so it needs a clean green rerun before merge.

The wheel smoke jobs that completed are green, which is good; please address the metadata/scope notes and re-request review after the full matrix is green.

Arryan21-06:
Thanks for the review @im-anishraj ! I have just updated the main PR description body to include the Fixes #924 tag, the exact local tests ran, the checkboxes, and the explicit out-of-scope note regarding sample_size pre-scan. >
As for the CI, you are exactly right—the Windows Python 3.11 job just hit a random GitHub auth/checkout glitch while the others passed. Because I'm contributing from a fork, I don't have the permission button to re-run the failed job myself. Could you please trigger a re-run for that specific checkout job when you get a chance?

im-anishraj:
Thanks for updating the PR body; the metadata, `Fixes #924`, local test list, and sample-size/pre-scan scope note are in place now.

I triggered a rerun of the failed CI job. Since this touches the C++ chunk reader, I’ll wait for that rerun to go green before doing the final local review gate.

Arryan21-06:
Hi @im-anishraj , The CI rerun finished and all checks are completely green now!  The PR is ready for that final local review gate whenever you have the time. Thanks again for your help getting this across the finish line.

im-anishraj (changes requested):
Thanks for getting CI green and for the PR body updates. I ran the local gate on a clean PR checkout with Python 3.12 and MSVC; the build/test side is good:
- editable install with dev/parquet extras passed (using local Catch2 cache)
- `pytest tests/test_csv_chunked.py -v --tb=short`: 18 passed
- full `pytest -q`: 1625 passed, 7 skipped
- `black --check arnio tests`: passed
- `ruff check .`: passed
- `mypy arnio`: passed (mypy noted project `python_version=3.9` is below current mypy's supported floor, but exited 0)
- MSVC native C++ build plus `ctest`: 4/4 passed
- GitHub C++/wheel/full Python CI is green

I found one remaining blocker before approval: the new runtime error tells users to “specify 'dtypes' explicitly or use a larger 'sample_size'”, but `read_csv_chunked()` does not expose `dtypes`, `dtype`, or `sample_size` today. Please change the hint to only mention available chunked-reader remedies, or add the public parameters in a separate scoped PR. After that wording fix and green checks on the new head, I can approve quickly.

im-anishraj (approved):
Thanks for the quick follow-up. I re-reviewed the new head and the only change since my local gate is the chunked-reader error message. The hint now points to available remedies instead of unavailable `dtypes` / `sample_size`, so the remaining blocker is resolved.

Previously completed local gate on this PR head family:
- editable install with dev/parquet extras passed using MSVC/Python 3.12
- `pytest tests/test_csv_chunked.py -v --tb=short`: 18 passed
- full `pytest -q`: 1625 passed, 7 skipped
- `black --check arnio tests`: passed
- `ruff check .`: passed
- `mypy arnio`: passed
- MSVC native C++ build plus `ctest`: 4/4 passed

The latest GitHub lint, C++ native tests, Windows MSVC wheel smoke, manylinux wheel smoke, macOS arm64 wheel smoke, examples, and full Python matrix are green. Approving.

PR #1298, opened 2026-05-23
feat(csv): add on_bad_lines parameter to scan_csv

### LINKED ISSUE #1294: Feature: Add on_bad_lines parameter to scan_csv for API consistency with read_csv
## Description
`scan_csv` currently has no `on_bad_lines` parameter, while `read_csv` supports `"error"`, `"warn"`, and `"skip"`. This inconsistency means malformed rows during schema inference are always treated as errors with no way to warn or skip.

## Expected Behavior
```python
schema = ar.scan_csv("data.csv", on_bad_lines="warn")
```

## Suggested Implementation
1. Add `on_bad_lines` parameter to `scan_csv` in `arnio/io.py`
2. Pass it through to `CsvReader::scan_schema` in `cpp/src/csv_reader.cpp`
3. Add tests for warn/skip behavior during schema inference

**Files:** `arnio/io.py`, `cpp/src/csv_reader.cpp`, `cpp/include/arnio/csv_reader.h`

### Proposed solution

Add `on_bad_lines` parameter to `scan_csv` in `arnio/io.py` with the same `"error"`, `"warn"`, `"skip"` options as `read_csv`. Pass it through to `CsvReader::scan_schema` in `cpp/src/csv_reader.cpp`.

### User story

As a data engineer scanning large CSV files for schema inference, I want to handle malformed rows gracefully during scan_csv so that schema detection doesn't fail on files with occasional bad rows.

### Acceptance criteria

- [ ] scan_csv("file.csv", on_bad_lines="warn") works without error
- [ ] scan_csv("file.csv", on_bad_lines="skip") skips bad rows silently
- [ ] Default behavior (on_bad_lines="error") is unchanged
- [ ] Tests added for warn/skip behavior during schema inference

### Alternatives considered

_No response_

### Scope

CSV reader / I/O

### Expected difficulty

Intermediate

Yes - I'd like to submit a PR

### Pre-flight checks

### Additional context

Relevant files: `arnio/io.py` `scan_csv` function. `read_csv` already supports `on_bad_lines` with full warn/skip/error handling; this is primarily a plumbing task to expose the same parameter in `scan_csv`.

 I would like to be assigned this issue.

SiddharthRiot:
Fixes #1294

## Changes
- Added `on_bad_lines` parameter to `scan_csv` in `arnio/io.py` with `"error"`, `"warn"`, `"skip"` options matching `read_csv`
- Updated `CsvReader::scan_schema` signature in `cpp/include/arnio/csv_reader.h`
- Updated `scan_schema` implementation in `cpp/src/csv_reader.cpp` to handle bad rows conditionally
- Updated pybind11 binding in `bindings/bind_arnio.cpp`
- Added focused tests in `tests/test_csv_on_bad_lines.py`

## Behavior
- Default `on_bad_lines="error"` is unchanged
- `on_bad_lines="warn"` skips bad rows silently during schema inference
- `on_bad_lines="skip"` skips bad rows silently during schema inference
- Invalid values raise `ValueError`

## Tests
- 4 new tests added in `TestScanCsvOnBadLines`
- Full test suite: 1590 passed, 43 skipped


im-anishraj (changes requested):
Thanks for the PR. The implementation direction looks right, but I need to request changes before this can merge:

- CI is currently failing in the `clang-format` check. Please run clang-format on the touched C++ files, especially `cpp/include/arnio/csv_reader.h` and `bindings/bind_arnio.cpp`.
- Please document the new `scan_csv(on_bad_lines=...)` parameter in the `scan_csv` docstring/API docs so this public API addition is discoverable.

Once those are fixed and CI is green, I will re-review.

SiddharthRiot:
Addressed both review points:

1. clang-format-17 applied to `cpp/include/arnio/csv_reader.h` and `bindings/bind_arnio.cpp`
2. `on_bad_lines` and `encoding_errors` documented in `scan_csv` docstring

CI should be green now.

im-anishraj (changes requested):
Thanks for the quick follow-up. The clang-format and docstring items are addressed, and CI is green now.

One blocker remains before I can approve: `scan_csv(..., on_bad_lines="warn")` currently skips malformed rows without emitting a warning, so it behaves the same as `"skip"`. Since this is intended to match `read_csv` semantics and the new docstring says `"warn"` emits `UserWarning`, please surface the malformed-row information from `scan_schema` or otherwise emit a `UserWarning` for `"warn"`, and add a focused test such as `pytest.warns(UserWarning)` while keeping `"skip"` silent.

After that I can re-review.

SiddharthRiot:
Fixed: `scan_csv(..., on_bad_lines="warn")` now emits a `UserWarning` listing malformed rows, matching `read_csv` semantics. `"skip"` remains silent. Added `pytest.warns(UserWarning)` test to cover this.

Ready for re-review.

im-anishraj (changes requested):
Thanks for addressing the warning case, but I still need changes before approval.

The new `_scan_collect_bad_rows()` is a parallel Python CSV parser based on `line.count(delimiter)`. That will diverge from the real scan path for quoted delimiters, escaped/newline-containing fields, non-default encodings/`encoding_errors`, `has_header=False`, and `sample_size`; it also scans the file independently from the C++ parser that actually determines schema. This would make `warn` behavior inconsistent with both `read_csv` and `scan_schema`.

Please avoid adding a second parser. Instead, have the scan-schema path surface the malformed-row information it already detects, or otherwise collect warnings from the same parser/row-width validation logic used for schema inference. Then add tests that would catch this divergence, especially quoted delimiters and the warning-vs-skip behavior.

SiddharthRiot:
Reworked the implementation: dropped the parallel Python parser. `scan_schema` now collects malformed-row messages in C++ (same parser, same row-width validation) and returns them alongside the schema. Python emits `UserWarning` from those messages when `on_bad_lines="warn"`. `"skip"` remains silent.

Needs a rebuild to pick up the C++ changes can you trigger CI or let me know if there's a preferred way to verify locally?

im-anishraj:
Thanks for the update. I do not see the reworked implementation on the PR branch yet: the current head still appears to be `fc99baf`, and the diff still contains `_scan_collect_bad_rows()` in Python.

Please push the C++-based rework to this PR branch. CI should trigger automatically after the push; if GitHub asks for maintainer approval on the new workflow run, I can approve it. Once the new diff is visible and CI is green, I will re-review.

im-anishraj (changes requested):
Thanks, the C++ scan-schema rework is visible now and it removes the parallel Python parser. I still need changes before approval:

- `arnio/_arnio_cpp.pyi` still types `CsvReader.scan_schema()` as returning `Mapping[str, str]`, but the pybind binding now returns `(schema, bad_rows)`. Please update the stub to the new tuple return type so `scan_csv()` and type checking agree with the native API.
- Please add the quoted-delimiter regression test requested earlier, e.g. a valid row containing a quoted delimiter should not be reported as malformed while a real malformed row still warns/skips from the same scan path.
- CI has a failed lint job on the new commit. Logs are not available while the run is still in progress, but please fix that once GitHub exposes the details.

After those are addressed and CI is green, I can re-review.

SiddharthRiot:
Updated:
- `_arnio_cpp.pyi`: `scan_schema` return type updated to `tuple[Mapping[str, str], list[str]]`
- Added quoted-delimiter regression tests: valid quoted fields are not flagged as malformed, real bad rows still warn from the same C++ scan path

Waiting for CI.

im-anishraj (changes requested):
Thanks, the stub return type and quoted-delimiter tests are now addressed.

One remaining correctness issue before approval: `scan_schema()` builds warning messages with `sample_count + 2`, but `sample_count` only increments for accepted sample rows. If there are multiple skipped/warned malformed records, the reported CSV row numbers can be duplicated or wrong. Please track the physical record number in the scan loop, similar to the read/chunk paths, and add a focused test with multiple malformed rows to verify the warning messages use the real CSV row numbers.

CI is also still pending, so I will re-review once this is fixed and the checks are green.

SiddharthRiot:
Fixed: `scan_schema` now tracks physical record number independently of `sample_count`, so row numbers in warnings are always correct even when multiple bad rows are skipped. Added test with multiple malformed rows verifying exact row numbers in the warning message.

im-anishraj (changes requested):
Thanks, this fixes the consecutive malformed-row case and the new test is useful. One row-numbering blocker remains before approval:

`scan_schema()` now says it increments `record_number` for every physical row, but it currently increments after `if (line.empty()) continue;`, and initializes to `0` when `has_header=False` even though the first record has already been consumed. That means warnings/errors can still report the wrong physical CSV row when a blank record appears before a malformed row, or when `scan_csv(..., has_header=False, on_bad_lines="warn")` sees a malformed row after the first record.

Please align this with the read/chunk paths by counting every consumed record before the blank-line skip and starting from the already-consumed first record for both header and no-header scans. Please also add focused tests for a blank row before a bad row and for `has_header=False` row numbering.

CI is still pending on the latest commit, so I will re-review once this is fixed and the wheel/matrix checks are green.

SiddharthRiot:
Reworked: `record_number` now increments before the blank-line skip and initializes correctly for `has_header=False` from the already-consumed first record. Added focused tests for blank-before-bad-row and `has_header=False `row numbering. Waiting on CI.

im-anishraj (approved):
Looks good now. The scan-schema path surfaces bad-row warnings from the native parser, the stub and Python wrapper agree on the tuple return shape, and the regression coverage now includes quoted delimiters, multiple malformed rows, blank-line row numbering, and has_header=False row numbering.

Local verification on Windows/Python 3.12 with MSVC-built editable install passed:
- black --check arnio tests
- ruff check .
- mypy arnio
- pytest tests/test_csv_on_bad_lines.py -v --tb=short
- pytest -v --tb=short

GitHub Actions is green, including the Windows MSVC, manylinux, macOS arm64 wheel smoke checks and full Python matrix.

SiddharthRiot:
Thank you @im-anishraj  for the thorough reviews each round genuinely improved the implementation. Glad it's in!
Also.. please add the *GSSoC* labels to this PR.

im-anishraj:
Thanks for the careful follow-through on the review rounds. I’ve added the GSSoC labels to this merged PR.

PR #1305, opened 2026-05-24
fix(io): add skiprows alias to read_csv_chunked for API consistency

### LINKED ISSUE #1293: Bug: read_csv and read_csv_chunked use inconsistent skip-row parameter names
### What happened?

<html>
<body>
<h3><span>Problem</span></h3><p class="isSelectedEnd"><code dir="ltr"><span>read_csv()</span></code><span> and </span><code dir="ltr"><span>read_csv_chunked()</span></code><span> currently use different parameter names for the same functionality:</span></p>
Function | Parameter
-- | --
read_csv() | skiprows
read_csv_chunked() | skip_rows

<p><span>This creates an API inconsistency and makes it easy for users to accidentally pass the wrong keyword when switching between the two functions.</span></p>
</body>
</html>

### Area

### Minimal reproduction

Reproduction
import arnio as ar

# Works
frame = ar.read_csv("data.csv", skiprows=2)

# Fails
for chunk in ar.read_csv_chunked("data.csv", skiprows=2):
    pass

### Expected behavior

Both APIs should use the same parameter name for consistency with each other and with pandas.

### Actual behavior

Current result:

TypeError: unexpected keyword argument 'skiprows'

### Impact

Breaks documented API

### Arnio version

1.1.0

### Install method

pip install -e .[dev]

### Operating System

Linux

### Python version

3.12.4

### Environment details

```shell

```

### Pre-flight checks

### Additional context

Suggested Fix

Add skiprows as an alias for skip_rows in read_csv_chunked() (or rename the parameter entirely while preserving backward compatibility).

Files involved
io.py

Hey @im-anishraj !I would like to work on this!

ayushlad9108:
## Summary

Fixes #1293

`read_csv()` uses `skiprows` (matching pandas convention) but `read_csv_chunked()` only accepted `skip_rows`, causing a confusing `TypeError` when users switched between the two functions.

## Changes

- Added `skiprows` as a keyword-only alias in `read_csv_chunked()`
- Conflict resolution logic:
  - Only `skiprows` → used as the row-skip count
  - Only `skip_rows` → existing behavior unchanged
  - Both with same value → no error
  - Both with conflicting values → `ValueError` with clear message
- `skip_rows` retained for full backward compatibility
- Docstring updated to document both parameters

## Tests Added (`TestReadCsvChunkedSkiprowsAlias`)

- `skiprows` alias skips the correct number of rows
- `skiprows=0` is a no-op
- `skip_rows` still works unchanged
- Both params with same value do not raise
- Conflicting values raise `ValueError` with clear message
- Invalid types (`bool`, `float`) raise `TypeError`
- Negative value raises `ValueError`

## Files Changed

- `arnio/io.py`
- `tests/test_csv.py`

### DIFF
--- a/arnio/io.py
+++ b/arnio/io.py
@@ -642,12 +642,10 @@ def read_csv_chunked(
     nrows : int, optional
         Maximum total number of data rows to read across all chunks.
     skip_rows : int, default 0
-        Number of lines to skip before the header row. Useful for CSV files
-        with metadata preambles before the actual data.
+        Number of data rows to skip after the header row.
         Deprecated alias: use ``skiprows`` instead (matches ``read_csv``).
     skiprows : int, optional
         Alias for ``skip_rows``, matching the ``read_csv`` parameter name.
-        Skips lines before the header row.
         If both ``skiprows`` and ``skip_rows`` are provided they must agree;
         conflicting values raise ``ValueError``.
     encoding : str, default "utf-8"
--- a/arnio/io.py
+++ b/arnio/io.py
@@ -642,10 +642,12 @@ def read_csv_chunked(
     nrows : int, optional
         Maximum total number of data rows to read across all chunks.
     skip_rows : int, default 0
-        Number of data rows to skip after the header row.
+        Number of lines to skip before the header row. Useful for CSV files
+        with metadata preambles before the actual data.
         Deprecated alias: use ``skiprows`` instead (matches ``read_csv``).
     skiprows : int, optional
         Alias for ``skip_rows``, matching the ``read_csv`` parameter name.
+        Skips lines before the header row.
         If both ``skiprows`` and ``skip_rows`` are provided they must agree;
         conflicting values raise ``ValueError``.
     encoding : str, default "utf-8"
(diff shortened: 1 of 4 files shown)


ayushlad9108:
@im-anishraj Done with the changes!

im-anishraj (changes requested):
Thanks for the update. Adding a `skiprows` alias to `read_csv_chunked()` is the right goal, but the second commit changes existing `skip_rows` semantics in `CsvChunkReader::open()` from post-header data-row skipping to pre-header line skipping.

That is a backwards-incompatible behavior change for the existing `skip_rows` parameter and is larger than an alias/consistency fix. Please keep the current `skip_rows` behavior compatible, or split the semantic change into a separate design PR with explicit migration notes and tests for both old and new behavior. The alias PR should not silently change how existing callers read chunked CSV files.

im-anishraj:
Hi @ayushlad9108, maintainer deadline update: please push the requested fixes/rebase or reply with your ETA by May 31, 2026 at 11:59 PM IST.

PR #1318, opened 2026-05-24
fix(pipeline): correct row_counts and step_timings metadata when dry_run=True

### LINKED ISSUE #1315: Bug: pipeline() returns misleading row_counts and inconsistent step_timings when dry_run=True is combined with return_metadata=True
### What happened?

When `pipeline()` is called with both `dry_run=True` and `return_metadata=True`, 
the returned metadata is incorrect in two ways:

1. `row_counts["after"]` reflects the transformed frame's row count instead of 
   the original frame's row count even though dry_run means no mutation should occur.

2. `step_timings["seconds"]` is captured via a second `perf_counter()` call made 
   after `elapsed_ms` was already computed, making both values inconsistent with 
   each other.

There is also no `"dry_run"` flag in the metadata to indicate the run was simulated.

### Area

Pipeline / custom steps

### Minimal reproduction

```
import arnio as ar

frame = ar.read_csv("sample.csv")

result, meta = ar.pipeline(
    frame,
    [("drop_nulls",), ("strip_whitespace",)],
    dry_run=True,
    return_metadata=True,
)

# result.shape[0] == original row count (correct, no mutation)
# but:
print(meta["row_counts"])
# "after" shows transformed count, not original misleading

print(meta["step_timings"])
# "seconds" captured from second perf_counter() call inconsistent with elapsed_ms
```

### Expected behavior

1. `row_counts["after"]` should equal `row_counts["before"]` when dry_run=True, 
   since no mutation occurs.

2. `step_timings["seconds"]` should be captured once using a single 
   `perf_counter()` call, consistent with `elapsed_ms`.

3. Metadata should include a `"dry_run": True` flag so callers can 
   distinguish simulated runs from real ones.

### Actual behavior

1. `row_counts["after"]` reflects the transformed frame's row count 
   (e.g., after drop_nulls removes rows), even though `result` was 
   never updated due to dry_run=True. This makes metadata misleading.

2. `step_timings["seconds"]` is computed from a second `perf_counter()` 
   call made after `elapsed_ms` was already captured both values 
   differ and are inconsistent.

3. No `"dry_run"` indicator exists in the returned metadata dict.

### Impact

Produces incorrect data

### Arnio version

1.1.1

### Install method

pip install arnio

### Operating System

Windows

### Python version

3.12.4

### Environment details

```shell
Bug is reproducible via static code analysis of arnio/pipeline.py.
The double perf_counter() call and dry_run row_count logic are 
visible at the return_metadata block for both C++ and Python step paths.
```

### Pre-flight checks

### Additional context

Root cause identified in `arnio/pipeline.py`:

**Problem A** (row_counts):
```python
# Both C++ and Python step paths have this issue
row_counts.append({
    "after": step_result.shape[0],  # step_result, not result
})
# When dry_run=True, result is never updated but step_result still 
# reflects the transformed frame so "after" count is wrong.
```

**Problem B** (step_timings):
```python
elapsed_ms = (perf_counter() - started_at) * 1000  # captured here
# ... more code ...
step_timings.append({
    "seconds": round(perf_counter() - started_at, 9),  # captured AGAIN
})
# Two different perf_counter() calls = two different values.
```

Fix is straightforward capture elapsed once, and use `result.shape[0]` 
for "after" when dry_run=True. Same fix needed in both C++ and Python step paths.

I have identified the exact lines in `arnio/pipeline.py` where both issues occur 
and have a fix ready. Would like to take this up and submit a PR.

SiddharthRiot:
## Summary

Fixes two bugs in `pipeline()` metadata when `dry_run=True` is combined with `return_metadata=True`.

## Problems Fixed

**Bug 1: `row_counts["after"]` was misleading in dry_run mode**

When `dry_run=True`, `result` is never mutated but `row_counts["after"]` was 
reporting `step_result.shape[0]` (the transformed frame) instead of `result.shape[0]` 
(the unchanged frame). This made metadata appear as if rows were actually dropped.

**Bug 2: `step_timings["seconds"]` was inconsistent**

`elapsed_ms` was captured once, but `step_timings["seconds"]` was computed from a 
second `perf_counter()` call made later producing a different, larger value.

## Changes

**`arnio/pipeline.py`**
- Capture elapsed time once via `elapsed_sec = perf_counter() - started_at`
- Use `result.shape[0] if dry_run else step_result.shape[0]` for `row_counts["after"]`
- Add backward-compatible `"dry_run"` flag to both `row_counts` and `step_timings` entries
- Same fix applied to both C++ step path and Python step path

**`tests/test_pipeline.py`**
- `test_pipeline_dry_run_with_metadata_row_counts_unchanged` verifies `after == before` in dry_run
- `test_pipeline_dry_run_with_metadata_step_timings_consistent` verifies `dry_run` flag present and seconds >= 0
- `test_pipeline_dry_run_false_metadata_unchanged` regression: `dry_run=False` metadata unaffected
- Updated `test_pipeline_return_metadata_handles_python_steps` to include new `dry_run` field

## Test Results
All 95 tests passing. Black and Ruff checks clean.

Closes #1315

### DIFF
--- a/arnio/pipeline.py
+++ b/arnio/pipeline.py
@@ -382,7 +382,8 @@ def pipeline(
                 if not dry_run:
                     result = step_result
 
-            elapsed_ms = (perf_counter() - started_at) * 1000
+            elapsed_sec = perf_counter() - started_at
+            elapsed_ms = elapsed_sec * 1000
 
             if verbose:
                 execution_path = f"{fn.__module__}.{fn.__name__}"
@@ -404,13 +405,15 @@ def pipeline(
                     {
                         "step": name,
                         "before": rows_before,
-                        "after": step_result.shape[0],
+                        "after": result.shape[0] if dry_run else step_result.shape[0],
+                        "dry_run": dry_run,
                     }
                 )
                 step_timings.append(
                     {
                         "step": name,
-                        "seconds": round(perf_counter() - started_at, 9),
+                        "seconds": round(elapsed_sec, 9),
+                        "dry_run": dry_run,
                     }
                 )
         elif name in python_step_registry:
@@ -456,7 +459,8 @@ def pipeline(
             if not dry_run:
                 result = step_result
 
-            elapsed_ms = (perf_counter() - started_at) * 1000
+            elapsed_sec = perf_counter() - started_at
+            elapsed_ms = elapsed_sec * 1000
 
             if verbose:
                 step_name = getattr(fn, "__name__", name)
@@ -479,13 +483,15 @@ def pipeline(
                     {
                         "step": name,
                         "before": rows_before,
-                        "after": step_result.shape[0],
+                        "after": result.shape[0] if dry_run else step_result.shape[0],
+                        "dry_run": dry_run,
                     }
                 )
                 step_timings.append(
                     {
                         "step": name,
-                        "seconds": round(perf_counter() - started_at, 9),
+                        "seconds": round(elapsed_sec, 9),
+                        "dry_run": dry_run,
                     }
                 )
         else:
--- a/tests/test_pipeline.py
+++ b/tests/test_pipeline.py
@@ -589,6 +589,7 @@ class TestPipeline:
                 "step": "timed_python_step",
                 "before": frame.shape[0],
                 "after": result.shape[0],
+                "dry_run": False,
             }
         ]
         assert len(metadata["step_timings"]) == 1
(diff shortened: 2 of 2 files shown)


im-anishraj (approved):
Thanks for re-opening this as a clean scoped PR. I reviewed the diff and ran the local gate on a clean worktree:
- `py_compile arnio/pipeline.py`: passed
- `black --check arnio/pipeline.py tests/test_pipeline.py`: passed
- `ruff check arnio/pipeline.py tests/test_pipeline.py`: passed
- `mypy arnio`: passed
- `pytest tests/test_pipeline.py -q`: 95 passed

The full GitHub CI matrix is green, the PR is non-draft and mergeable, and the change is limited to the #1315 dry-run metadata fix. Approving.

SiddharthRiot:
Thanks @im-anishraj  for the thorough review and for running the full gate locally. 
Really appreciate the detailed feedback on keeping PRs scoped.

Could you please add the relevant GSSoC labels to the linked issue #1315? 
Wanted to make sure it gets counted under the GSSoC 2026 program.

PR #1341, opened 2026-05-24
feat: add ArFrame._repr_html_() for notebook-friendly display

### LINKED ISSUE #221: ArFrame: Add notebook-friendly HTML representation
## Context
Jupyter users should see a readable preview instead of only a compact repr.

## Suggested files

## Scope

## Acceptance criteria

## Contributor notes

AaryanInzalkar:
## Summary
Adds `ArFrame._repr_html_()` so Jupyter/IPython users see a readable HTML table preview instead of a plain `ArFrame(N rows × M cols)` repr.

## Changes
- `arnio/frame.py` — added `_repr_html_()` method to `ArFrame`
- `tests/test_frame.py` — added tests covering normal, edge, and HTML-escaping cases

## Behaviour
- Shows a shape + dtype summary line above the table
- Renders up to 10 data rows (bounded — no unbounded full-frame rendering)
- Shows "Showing 10 of N rows" notice when the frame is larger
- All column names and cell values are HTML-escaped to prevent markup injection
- Empty frames return a safe fallback string without crashing

## Tests added
- Return type is `str`
- Table structure (`<table>`, `<thead>`, `<tbody>`) is present
- Column names and cell values appear in output
- Shape and dtypes appear in summary line
- Truncation notice present for large frames, absent for small ones
- `<tbody>` capped at exactly 10 rows
- Empty frame and null values don't crash
- `<script>` in cell values is escaped to `&lt;script&gt;`
- HTML in column names is escaped

Fixes #221

### DIFF
--- a/arnio/frame.py
+++ b/arnio/frame.py
@@ -831,7 +831,6 @@ class ArFrame:
         """
         import html as _html
 
-        from .convert import to_pandas
 
         _REPR_HTML_MAX_ROWS = 10
 
@@ -866,11 +865,17 @@ class ArFrame:
         )
         header = f"<thead><tr>{header_cells}</tr></thead>"
 
-        # ── data rows — only convert the bounded slice, not the full frame ──
+                # Read only the rows needed for display; do not convert the full frame.
+        preview_rows = min(num_rows, _REPR_HTML_MAX_ROWS)
+
         try:
-            bounded = self.head(_REPR_HTML_MAX_ROWS)
-            df = to_pandas(bounded)
-            preview = df
+            preview_values = [
+                [
+                    self._frame.column_by_index(col_idx).at(row_idx)
+                    for col_idx in range(num_cols)
+                ]
+                for row_idx in range(preview_rows)
+            ]
         except Exception as exc:
             return (
                 summary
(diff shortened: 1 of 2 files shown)


im-anishraj (changes requested):
Thanks for the clean scope and the HTML escaping coverage. I can’t approve this implementation yet because the preview is bounded visually but not bounded computationally.

Required change:

`_repr_html_()` currently calls `to_pandas(self)` and only then slices to 10 rows. For a large `ArFrame`, opening a notebook cell would still materialize the entire dataset into pandas just to render a small preview. That can be slow and memory-heavy, which is risky for a display hook that runs automatically.

Please change the implementation so it only reads/converts the rows needed for the preview, or reuse an existing bounded preview path instead of converting the full frame. Please also add a regression test that would fail if `_repr_html_()` calls full-frame `to_pandas()` for a large frame.

im-anishraj (approved):
Reviewed the updated implementation. The previous blocker is resolved: `_repr_html_()` no longer converts the full frame to pandas and renders only a bounded preview, with HTML escaping and regression coverage. PR CI is green except the known Vercel authorization check, and merge simulation against current main passes local static checks.

PR #1360, opened 2026-05-25
fix: improve compare_profiles error message when exclude_columns causes schema mismatch

### LINKED ISSUE #1343: Bug: compare_profiles raises ValueError when profiles are built with different exclude_columns no user warning
### What happened?

When `profile()` is called with `exclude_columns` on one frame and without it on another, then both profiles are passed to `compare_profiles()`, it raises a `ValueError` about incompatible schemas. There is no warning or guard to prevent this the user has no way to know that `exclude_columns` makes profiles incompatible with `compare_profiles`.

### Area

Data profiling / quality reports

### Minimal reproduction

```
import arnio as ar

# Profile with some columns excluded
profile_a = ar.profile(frame, exclude_columns=["col1"])

# Profile without exclusions (e.g. on a newer batch)
profile_b = ar.profile(frame)

# Raises ValueError: Profiles have incompatible schemas
comparison = ar.compare_profiles(profile_a, profile_b)
```

### Expected behavior

compare_profiles should either:
1. Raise a clear, actionable error that explicitly mentions `exclude_columns` was likely used, helping the user understand why schemas are incompatible.
2. Or accept partial profiles and compare only the shared columns, similar to how SQL JOIN handles mismatched columns.

### Actual behavior

Raises a generic ValueError:
Profiles have incompatible schemas: missing from `profile_a=['col1']`, missing from `profile_b=[]`

No mention of `exclude_columns`. User has no context to understand why this happened or how to fix it.

### Impact

Breaks documented API

### Arnio version

1.1.0

### Install method

pip install arnio

### Operating System

Windows

### Python version

3.12.4

### Environment details

```shell
Bug is reproducible via static code analysis of arnio/quality.py.
The missing exclude_columns hint is visible in the compare_profiles() function
where ValueError is raised when column sets don't match, with no reference to
exclude_columns being the likely cause.

```

### Pre-flight checks

### Additional context

Root cause is in `quality.py` `compare_profiles()` raises ValueError when column sets don't match, but gives no hint that exclude_columns on profile() is the likely cause.

```
Relevant code in quality.py:
if columns_a != columns_b:
    raise ValueError(
        "Profiles have incompatible schemas: "
        f"missing from profile_a={missing_from_a}, "
        f"missing from profile_b={missing_from_b}"
    )
```

Suggested fix: detect if column mismatch size matches typical `exclude_columns` usage and add a hint to the error message, or add a `shared_columns_only` parameter to `compare_profiles`.

I have identified the exact lines in `arnio/quality.py` where this issue occurs and have a fix ready. Would like to take this up and submit a PR.

SiddharthRiot:
Fixes #1343

## What changed
- Updated the `ValueError` in `compare_profiles()` to mention that mismatched `exclude_columns` usage in `profile()` is the likely cause of schema incompatibility
- Added actionable guidance to profile both datasets with the same included/excluded columns before comparing

## Tests
- Added regression test `test_compare_profiles_mismatched_exclude_columns_hints_user` in `tests/test_quality.py`

### DIFF
--- a/arnio/quality.py
+++ b/arnio/quality.py
@@ -1132,7 +1132,10 @@ def compare_profiles(
         raise ValueError(
             "Profiles have incompatible schemas: "
             f"missing from profile_a={missing_from_a}, "
-            f"missing from profile_b={missing_from_b}"
+            f"missing from profile_b={missing_from_b}. "
+            "This is likely caused by calling profile() with different exclude_columns "
+            "on each dataset. Profile both datasets with the same included/excluded "
+            "columns before comparing."
         )
 
     drift_report: dict[str, dict[str, Any]] = {}
--- a/tests/test_quality.py
+++ b/tests/test_quality.py
@@ -2576,4 +2576,3 @@ def test_compare_profiles_mismatched_exclude_columns_hints_user():
 
     with pytest.raises(ValueError, match="exclude_columns"):
         ar.compare_profiles(profile_a, profile_b)
-
--- a/tests/test_quality.py
+++ b/tests/test_quality.py
@@ -2568,112 +2568,11 @@ def test_data_quality_report_to_json_redact_sample_values():
     assert parsed["columns"]["name"]["sample_values"] == ["[REDACTED]"]
 
 
-# --- Tests for ProfileComparison.to_json() ---
+def test_compare_profiles_mismatched_exclude_columns_hints_user():
+    df = pd.DataFrame({"a": [1, 2, 3], "b": [4, 5, 6], "c": [7, 8, 9]})
+    frame = ar.from_pandas(df)
+    profile_a = ar.profile(frame, exclude_columns=["c"])
+    profile_b = ar.profile(frame)
 
-
-def test_profile_comparison_to_json_returns_string():
-    frame = ar.from_pandas(pd.DataFrame({"x": [1, 2, 3]}))
-    p = ar.profile(frame)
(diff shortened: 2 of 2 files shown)


SiddharthRiot:
@im-anishraj changes are pushed, ready for re-review!

im-anishraj (changes requested):
Thanks for the update. This cannot be merged in the current form.

The `quality.py` error-message change is small and relevant, but the merge-conflict resolution removed a large block of existing `tests/test_quality.py` coverage for `ProfileComparison.to_json()`, `ProfileComparison.to_markdown()`, and `QualityGateResult.to_json()`. Those tests were added on main and should not be deleted by this PR.

Please rebase/sync with current `main`, restore the removed tests, and keep this PR limited to the `compare_profiles()` schema-mismatch hint plus its regression test.

im-anishraj:
The latest commit still leaves the same blocker: this PR deletes the existing \\ProfileComparison\\ and \\QualityGateResult\\ export tests from \\	ests/test_quality.py\\. Please restore those tests and keep only the compare_profiles error-message regression in this PR.

SiddharthRiot:
restored deleted tests and added the regression test. Ready for re-review!

im-anishraj (approved):
Approved after re-review. The deleted quality export tests are restored, the new regression test is scoped, and CI is green across lint, examples, native tests, and the Python matrix.

SiddharthRiot:
thanks for the review and merge!

PR #1367, opened 2026-05-25
feat: support generic sequence for subset in coalesce_columns

### LINKED ISSUE #1364: feat: Support any sequence type in coalesce_columns subset parameter
### Description
Currently, `coalesce_columns` has a rigid validation check on its `subset` parameter:
```python
if not isinstance(subset, list):
    raise TypeError("subset must be a list of column names")
```
This restricts the user from passing any other standard sequence/iterable types (like `tuple`, `set`, or pandas `Index`), throwing a `TypeError`. In contrast, other subset-taking functions in `arnio` (such as `drop_duplicates` and `combine_columns`) gracefully accept a generic `Sequence[str]` or custom sequences.

### Proposed Solution
Relax the strict `list` check on the `subset` parameter in `coalesce_columns` and use the unified internal sequence helper `_validate_existing_column_sequence` to coerce the subset into a standard list of column names, improving API consistency across all cleaning functions.

---
I am contributing on behalf of GSSoc’26.

gowthamrdyy:
### Summary
This PR unifies the `subset` parameter validation in `coalesce_columns` with other subset-taking functions in `arnio` (like `drop_duplicates` and `combine_columns`) by supporting any python `Sequence[str]` (such as lists or tuples) instead of strictly checking for a `list` type.

### Details
- Updated type hints and relaxed type validation of the `subset` parameter in `coalesce_columns`.
- Utilized the standard `_validate_existing_column_sequence` helper to normalize subset inputs.
- Aligned validation exception messages and tests to be consistent with all other subset operations.
- Added comprehensive unit tests in `tests/test_cleaning.py`.

Closes #1364

---
I am contributing on behalf of GSSoc’26.

### DIFF
--- a/arnio/cleaning.py
+++ b/arnio/cleaning.py
@@ -82,7 +82,7 @@ def _validate_column_sequence(
         raise TypeError(
             f"{argument_name} must be a sequence of column names, not a string"
         )
-    if not isinstance(columns, Sequence):
+    if not isinstance(columns, Sequence) and not isinstance(columns, pd.Index):
         raise TypeError(f"{argument_name} must be a sequence of column names")
 
     normalized = list(columns)
--- a/arnio/cleaning.py
+++ b/arnio/cleaning.py
@@ -1749,7 +1749,7 @@ def standardize_missing_tokens(frame, tokens=None, subset=None):
 def coalesce_columns(
     frame,
     *,
-    subset: list[str],
+    subset: Sequence[str],
     output_column: str = "coalesced",
 ):
     """Select the first non-null value from a list of columns.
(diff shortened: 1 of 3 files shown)


im-anishraj (changes requested):
Thanks for the update. This is not merge-ready yet.

CI is failing across the Python matrix because `tests/test_coalesce_columns.py::test_coalesce_columns_validation` still expects the old message `subset must be a list`, while the implementation now raises `subset must be a sequence of column names, not a string`. Please update the existing regression test so the full suite passes.

Also, please keep the scope explicit: `coalesce_columns` is order-sensitive, so unordered `set` input should not be accepted accidentally. If this PR is intended to support pandas `Index` as mentioned in the issue, add direct coverage for it; otherwise narrow the PR description to sequence-only behavior.

im-anishraj (approved):
Reviewed the updated head. The previous test expectation blocker is fixed, unordered set input is explicitly rejected, pandas Index input is covered, and the change stays scoped to #1364. Merge simulation against current main passes, and local Ruff/Black pass on the touched files. Current main CI is green; the only PR status is the known Vercel authorization check.

Issue #1337, opened 2026-05-24
Bug: read_csv throws C++ runtime_error on duplicate column headers


divyansha12:
### What happened?

I noticed that if a CSV file has duplicate column headers, read_csv completely crashes with a C++ std::runtime_error.

Real-world datasets often have messy or duplicate headers. Instead of a hard crash, Arnio should probably mangle the duplicate names (e.g., col, col.1, col.2) to safely load the file, similar to how pandas handles it.

### Area

CSV parser / read_csv / scan_csv

### Minimal reproduction

import arnio as ar

with open("test_dup.csv", "w") as f:
    f.write("id,name,name\n1,john,doe")

# This crashes
ar.read_csv("test_dup.csv")

### Expected behavior

It should load the dataframe and auto-rename the duplicate columns (like name and name.1) so they are unique.

### Actual behavior

RuntimeError: Duplicate column name: name

### Impact

Causes crash or data loss

### Arnio version

main

### Install method

built from source

### Operating System

macOS

### Python version

3.12.4

### Environment details

OS: macOS (Apple Silicon / ARM64)
Shell: zsh
Compiler: Apple clang
Python env: Local venv (/Users/divyanshasingh/arnio/.venv)

### Pre-flight checks

- [x] I searched existing issues and did not find a duplicate.
- [x] I included a minimal reproduction or explained why one is not possible.

### Additional context

HI @im-anishraj  I am participating in GSSoC '26. Please assign this to me, I can update the C++ parser to handle the header string deduplication!

im-anishraj:
Hi, thanks for the clear reproduction.

I checked the current Arnio behavior and test suite. Duplicate CSV headers are intentionally rejected today because accepting them silently can make column access ambiguous and can hide data-quality problems. Public `ar.read_csv(...)` should raise `CsvReadError`, not leak a raw C++ `RuntimeError`; if you are still seeing a raw `RuntimeError`, please share the full stack trace and exact commit/version.

Auto-renaming duplicate headers like pandas (`name`, `name.1`, `name.2`) would be a public API/design change. I’m marking this as needing maintainer decision before implementation. A likely safe path would be an explicit opt-in option such as `mangle_duplicate_headers=True`, while keeping strict rejection as the default.

divyansha12:
Hi @im-anishraj, thanks for the review!

[...]

PR #1370, opened 2026-05-25
fix: surface duplicate header rejection as CsvReadError

divyansha12:
## Summary
Caught the raw C++ `RuntimeError` caused by duplicate headers and raised a public `CsvReadError` instead. Used `from None` in `arnio/io.py` to stop the C++ exception trace from leaking to the user. Added a focused regression assertion to verify `__cause__ is None`.

This PR is strictly the error-boundary fix (Scope 1).

## Linked issue
Refs #1337

## Type of change
- [x] Bug fix

## Area
- [x] CSV parser / I/O
- [x] Python API / ArFrame

## Testing
- [x] `make test` (all tests passed locally)
- [x] `make lint`

## Contributor checklist
- [x] I kept the PR focused on one issue.
- [x] My PR title uses Conventional Commits.

Maintainer notes  
Scope 1 of #1337. The opt-in `mangle_duplicate_headers=True` behavior will be implemented in a follow-up PR.

### DIFF
--- a/arnio/io.py
+++ b/arnio/io.py
@@ -607,7 +607,7 @@ def read_csv(
     except CsvReadError:
         raise
     except RuntimeError as e:
-        raise CsvReadError(str(e)) from e
+        raise CsvReadError(str(e)) from None
 
     finally:
         if should_cleanup and os.path.exists(path):
@@ -766,7 +766,7 @@ def read_csv_chunked(
     except CsvReadError:
         raise
     except RuntimeError as e:
-        raise CsvReadError(str(e)) from e
+        raise CsvReadError(str(e)) from None
     finally:
         reader.close()
         if should_cleanup and os.path.exists(path):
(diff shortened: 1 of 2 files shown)


im-anishraj (changes requested):
Thanks, this implementation is scoped and the CI matrix is green. The `RuntimeError` -> `CsvReadError` boundary change and regression assertion look good.

One blocking metadata issue before merge: the PR body currently uses `Fixes #1337`, but this PR only handles the error-boundary part. The original issue still asks for opt-in duplicate-header mangling, so merging this as-is would close #1337 prematurely.

Please change the linked issue line to `Refs #1337` or `Part of #1337`. After that, this should be safe to accept as the first scoped fix.

divyansha12:
Hey @im-anishraj  its done. I changed the PR linkage to `Refs #1337` and added a short note that this is Scope 1 (error-boundary only). Please re-review when convenient.

im-anishraj (approved):
Approved. The PR is now correctly marked as `Refs #1337`, so the broader duplicate-header mangling issue will remain open. The code change is scoped to the Python error boundary, the regression assertion is focused, and the CI matrix is green.

PR #1474, opened 2026-05-25
Make duckdb smoke test blocking

### LINKED ISSUE #1399: CI: make DuckDB example smoke failures blocking
## Summary
The `examples-duckdb` job in `.github/workflows/examples-smoke.yml` is configured with `continue-on-error: true`, so a broken DuckDB example can fail without failing the examples smoke workflow.

## Why it matters
The workflow installs `duckdb` before running `examples/arnio_with_duckdb.py`, so this is not an optional-dependency-missing scenario. If that example breaks, maintainers should see a red check instead of a green workflow with a hidden failure.

## Evidence
- `.github/workflows/examples-smoke.yml` defines an `examples-duckdb` job.
- The job installs `duckdb` in the dependencies step.
- The job has `continue-on-error: true`, which makes failures non-blocking.

## Reproduction or trigger
1. Break `examples/arnio_with_duckdb.py` intentionally.
2. Open a PR that triggers `Examples Smoke`.
3. The DuckDB job can fail without blocking the workflow because `continue-on-error` is enabled.

## Expected behavior
The DuckDB example smoke job should fail the workflow when the example fails after dependencies are installed successfully.

## Actual behavior
The job is allowed to fail without blocking the workflow, reducing confidence in example coverage.

## Suggested direction
- Remove `continue-on-error: true` from the DuckDB example job.
- If maintainers want optional behavior for fork/network instability, make that explicit with a separate dependency-install guard or documented reason.
- Add a short comment in the workflow explaining why the job is required or optional.

## References
- `.github/workflows/examples-smoke.yml`
- `examples/arnio_with_duckdb.py`

SarthakKharche:
## Summary
Removed continue-on-error: true from the examples-duckdb job in .github/workflows/examples-smoke.yml and added a comment clarifying that this is a required smoke test. Since duckdb is explicitly installed as a dependency before the example runs, a failure in examples/arnio_with_duckdb.py should block the workflow visibly rather than being silently ignored.

## Linked issue
Fixes #1399 

## Type of change
- [ ] Bug fix
- [ ] Feature
- [ ] Performance improvement
- [ ] Documentation
- [ ] Tests
- [ ] Refactor
- [X] CI / packaging

## Area
- [ ] CSV parser / I/O
- [ ] C++ cleaning engine
- [ ] Python API / ArFrame
- [ ] Pipeline / custom steps
- [ ] Schema validation
- [ ] pandas interoperability
- [ ] Docs / website
- [X] Developer tooling

## Testing
No code logic was changed — only CI configuration. Verified by reviewing the workflow file directly.
- [ ] `make test`
- [ ] `make lint`
- [X] Other: Manual review of .github/workflows/examples-smoke.yml to confirm continue-on-error: true is removed only from examples-duckdb and no other jobs are affected.

## Contributor checklist
- [X] I read the contributing guide.
- [X] I kept the PR focused on one issue or one logical change.
- [X] I checked that generated files, logs, and local build artifacts are not committed.
- [X] My PR title uses Conventional Commits (`feat:`, `fix:`, `docs:`, `test:`, `refactor:`, `chore:`).

## Maintainer notes
This is a low-risk, one-line CI change. The only behavioral difference is that a broken examples/arnio_with_duckdb.py will now correctly fail the Examples Smoke workflow instead of producing a green check with a hidden failure. No follow-up work required unless maintainers want to add an explicit dependency-install guard for genuinely optional scenarios.

### DIFF
--- a/.github/workflows/examples-smoke.yml
+++ b/.github/workflows/examples-smoke.yml
@@ -58,7 +58,9 @@ jobs:
   examples-duckdb:
     name: Smoke – duckdb example
     runs-on: ubuntu-latest
-    continue-on-error: true
+
+    # Required smoke test: duckdb is explicitly installed here, so failures
+    # in examples/arnio_with_duckdb.py must block the workflow visibly.
 
     steps:
       - uses: actions/checkout@v4


im-anishraj (approved):
Reviewed the workflow diff and validated that the DuckDB smoke job no longer has continue-on-error. The change is scoped to CI/example smoke gating and matches #1399.

PR #1479, opened 2026-05-25
fix(quality): filter cast_types in exclude_columns to prevent suggestion leakage

### LINKED ISSUE #1466: Security: exclude_columns leaks cast_types suggestions
## Severity
High

## Files / lines
- `arnio/quality.py:201-258`
- `arnio/quality.py:1527-1559`

## Description
`DataQualityReport.to_dict(exclude_columns=...)` removes excluded columns from the `columns` section, but `cast_types` suggestions still leak excluded column names in both `kwargs` and `confidence_reason`.

## Why it happens
The suggestion filtering logic only removes excluded names from kwargs when the kwarg key is `subset`, `columns`, or literally `cast_types`. For a `cast_types` suggestion, the suggestion kwargs are the cast mapping itself (`{"secret": "int64"}`), so the excluded column key is never removed. The confidence reason generated by `suggest_cleaning()` also embeds the column name.

## How this fails in real use
A caller can intentionally exclude sensitive columns from a serialized report, but the output still reveals that a sensitive column exists and what type it appears to contain.

Reproduction:

```python

    "secret": ["1", "2"],
    "public": [" a ", "b "],
}))
report = ar.profile(frame)
out = report.to_dict(exclude_columns=["secret"])

assert "secret" not in out["columns"]
print(out["suggestions"])
# cast_types suggestion still contains {'secret': 'int64'} and reason text mentions 'secret'
```

## Exact fix / mitigation
Filter `cast_types` suggestion mappings by excluded column names when `step == "cast_types"`, not by kwarg key. Also redact or drop `confidence_reason` entries that mention excluded columns, or regenerate reasons after filtering.

## Existing tests
No. Tests cover column exclusion/redaction paths, but not suggestion kwargs or confidence reasons.

## Suggested test
Profile a frame with a cast suggestion on an excluded column, serialize with `exclude_columns`, and assert the excluded name appears nowhere in `json.dumps(report.to_dict(...))`.

AdityaM-IITH:
## Summary
Fixed a logic error in `DataQualityReport.to_dict()` where `cast_types` suggestions bypassed the `exclude_columns` filter. Changed the dictionary comprehension to correctly evaluate `s[0] == "cast_types"` (the step name) rather than mistakenly checking the nested key, which previously allowed excluded columns to leak into the final suggestions output.

## Linked issue
Fixes #1466

## Type of change
- [x] Bug fix
- [ ] Feature
- [ ] Performance improvement
- [ ] Documentation
- [ ] Tests
- [ ] Refactor
- [ ] CI / packaging

## Area
- [ ] CSV parser / I/O
- [ ] C++ cleaning engine
- [ ] Python API / ArFrame
- [ ] Pipeline / custom steps
- [x] Data quality / profiling
- [ ] Schema validation
- [ ] pandas interoperability
- [ ] Docs / website
- [ ] Developer tooling

## Testing
- [x] `make test`
- [x] `make lint`
- [ ] Other:

## Screenshots / Media
*N/A - Data dictionary JSON structure fixed behind the scenes.*

## Performance Impact
*N/A - Standard dictionary comprehension fix, no performance penalty.*

## Contributor checklist

## Maintainer notes
N/A

### DIFF
--- a/arnio/quality.py
+++ b/arnio/quality.py
@@ -218,6 +218,13 @@ class DataQualityReport:
 
             exclude_columns = set(exclude_columns)
 
+        def _redact_reason(reason: str | None) -> str | None:
+            if not reason or not exclude_columns:
+                return reason
+            for col in exclude_columns:
+                reason = reason.replace(f"'{col}'", "'[REDACTED]'")
+            return reason
+
         return {
             "row_count": self.row_count,
             "column_count": self.column_count,
@@ -247,7 +254,7 @@ class DataQualityReport:
                         }
                     ),
                     "confidence_score": getattr(s, "confidence_score", None),
-                    "confidence_reason": getattr(s, "confidence_reason", None),
+                    "confidence_reason": _redact_reason(getattr(s, "confidence_reason", None)),
                 }
                 for s in self.suggestions
             ],
--- a/arnio/quality.py
+++ b/arnio/quality.py
@@ -242,19 +242,26 @@ class DataQualityReport:
                 {
                     "step": s[0],
                     "kwargs": (
-                        {k: v for k, v in dict(s[1]).items() if k not in exclude_columns}
+                        {
+                            k: v
+                            for k, v in dict(s[1]).items()
+                            if k not in exclude_columns
+                        }
                         if s[0] == "cast_types"
                         else {
                             key: (
                                 [item for item in value if item not in exclude_columns]
-                                if key in {"subset", "columns"} and isinstance(value, list)
+                                if key in {"subset", "columns"}
+                                and isinstance(value, list)
                                 else value
                             )
                             for key, value in dict(s[1]).items()
                         }
                     ),
                     "confidence_score": getattr(s, "confidence_score", None),
-                    "confidence_reason": _redact_reason(getattr(s, "confidence_reason", None)),
+                    "confidence_reason": _redact_reason(
+                        getattr(s, "confidence_reason", None)
+                    ),
                 }
                 for s in self.suggestions
             ],
--- a/arnio/quality.py
+++ b/arnio/quality.py
@@ -234,22 +234,18 @@ class DataQualityReport:
             "suggestions": [
                 {
                     "step": s[0],
-                    "kwargs": {
-                        key: (
-                            [item for item in value if item not in exclude_columns]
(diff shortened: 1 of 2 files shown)


im-anishraj (changes requested):
Thanks for the focused PR. This is not ready yet.

Blocking issues:
- The linked issue reports leakage through both `cast_types` kwargs and `confidence_reason`. This PR filters the kwargs only, so excluded column names can still leak through the reason text.
- Please add a regression test that serializes `report.to_dict(exclude_columns=[...])` and asserts the excluded column name does not appear anywhere in `json.dumps(...)`.
- No CI/status checks are visible yet, so this also needs checks to run and pass before merge.

AdityaM-IITH:
I've updated the PR to also filter out the excluded column names from the confidence_reason strings (replacing them with [REDACTED]) so they don't leak into the report explanations. I've also added the requested regression test in tests/test_quality.py that serializes the report to JSON and asserts the excluded column name does not appear anywhere in the output. The checks should run and pass now. Let me know if you need anything else.

im-anishraj (changes requested):
This still cannot be merged.

I checked the current branch against latest `main`. The merge applies cleanly and Ruff passes, but Black still fails on the merge result:

```text
black --check arnio/quality.py tests/test_quality.py
would reformat arnio/quality.py
would reformat tests/test_quality.py
```

Also, no Arnio CI checks are visible on the PR yet. Please run Black on the touched files, push the formatted result, and make sure the normal CI matrix is visible/green before review continues.

AdityaM-IITH:
Done. I ran `black` on the modified files to fix the formatting inconsistency and pushed the result. The normal CI matrix checks should now trigger and turn green.

im-anishraj (approved):
Thanks for the update. I rechecked the current head against latest main: merge simulation is clean, Ruff passes, Black passes, and whitespace checks pass. The change is now scoped to the exclude_columns leakage fix and includes a JSON-level regression assertion. The only failing PR status is the external Vercel authorization check, which is unrelated to this Python/package change. Approving this.

im-anishraj:
Post-merge CI found a regression in the existing quality serialization test: the nested cast_types kwargs shape is not fully preserved/filtered. I opened follow-up issue #1563 to track the fix and keep main recovery focused.

PR #1483, opened 2026-05-25
fix(schema): normalize custom validator return values to prevent pd.N…

### LINKED ISSUE #1469: Bug: Custom validators returning pd.NA abort validation
## Severity
Medium

## Files / lines
- `arnio/schema.py:1984-2009`
- `arnio/schema.py:2301-2359`
- `tests/test_schema.py:2165-2236`
- `tests/test_register_validator.py:76-209`

## Description
Custom validators are not required to return a strict `bool`. A validator that returns `pd.NA` aborts validation with `TypeError: boolean value of NA is ambiguous` instead of producing a structured validation issue or a clear contract error.

## Why it happens
The custom validator path maps the function over non-null values and immediately calls `.astype(bool)`:

```python
invalid = non_null[~non_null.map(fn).astype(bool)]
```

`pd.NA` cannot be coerced to a Python boolean, so pandas raises before Arnio can return a `ValidationResult`.

## How this fails in real use
Validator functions often delegate to pandas/string operations that naturally return nullable booleans. A single ambiguous return value can crash an entire production validation run instead of marking that row invalid.

Reproduction:

```python

frame = ar.from_pandas(pd.DataFrame({"x": [1, 0, -1]}))
ar.register_validator("nullable_bool", lambda v: pd.NA if v == 0 else True)
ar.validate(frame, {"x": ar.Custom("nullable_bool")})
# TypeError: boolean value of NA is ambiguous
```

Returning `None` currently becomes `False`, so the return contract is also inconsistent.

## Exact fix / mitigation
Evaluate custom validators row-by-row and normalize the result explicitly. Accept only `True` as pass; treat `False`, `None`, and `pd.NA` as validation failures, or raise a clear `TypeError` that names the validator and row. Avoid pandas `.astype(bool)` on nullable results.

## Existing tests
Partial. Tests cover normal True/False custom validators and intentionally propagating raised exceptions, but not nullable/non-bool return values.

## Suggested test
Add custom validators returning `pd.NA`, `None`, and a non-bool truthy object. Assert either structured row-level failures or a clear `TypeError` with the validator name, depending on the chosen contract.

ayushlad9108:
<html>
<body>
<html><head></head><body><h2>Summary</h2><p>Fixes #1469</p><p>Custom validators currently go through:</p><pre><code class="language-python">non_null.map(fn).astype(bool)
</code></pre><p>which causes pandas to raise:</p><pre><code class="language-text">TypeError: boolean value of NA is ambiguous
</code></pre><p>when a validator returns <code inline="">pd.NA</code>.</p><p>This prevents Arnio from producing a structured validation result.</p><h2>Fix</h2><p>Added <code inline="">_normalize_validator_result()</code> in <code inline="">schema.py</code> to normalize validator return values before pandas evaluates them as booleans.</p><p>Behavior after the fix:</p>
Return value | Behavior
-- | --
True | validation passes
False | structured ValidationIssue
None | structured ValidationIssue
pd.NA | structured ValidationIssue (no crash)
anything else | raises TypeError naming the validator

<p>This keeps validation behavior strict and predictable while preventing pandas boolean-coercion errors.</p><h2>Tests</h2><p>Added 7 regression tests in <code inline="">test_register_validator.py</code> covering:</p><ul><li><p><code inline="">pd.NA</code> returns</p></li><li><p><code inline="">None</code> returns</p></li><li><p>invalid non-bool truthy returns</p></li><li><p>existing <code inline="">True</code> / <code inline="">False</code> behavior</p></li><li><p>the exact reproduction case from the issue report</p></li></ul></body></html>
</body>
</html>

### DIFF
--- a/arnio/schema.py
+++ b/arnio/schema.py
@@ -1994,7 +1994,11 @@ def _validate_column(
                     )
                 )
             else:
-                invalid = non_null[~non_null.map(fn).astype(bool)]
+                invalid = non_null[
+                    ~non_null.map(
+                        lambda v: _normalize_validator_result(fn(v), validator_name)
+                    )
+                ]
                 issues.extend(
                     _row_issues(
                         invalid,
@@ -2320,6 +2324,40 @@ def register_validator(name: str, fn: callable) -> None:
     if not isinstance(name, str) or not name:
         raise ValueError("name must be a non-empty string")
     _CUSTOM_VALIDATORS[name] = fn
+
+
+def _normalize_validator_result(result: object, validator_name: str) -> bool:
+    """Normalize a custom validator return value to a strict bool.
+
+    Contract:
+    - ``True``  → passes validation
+    - ``False`` → fails validation
+    - ``None``  → fails validation
+    - ``pd.NA`` → fails validation
+    - anything else → raises TypeError naming the validator
+
+    This avoids calling bool() on pd.NA (which raises
+    "TypeError: boolean value of NA is ambiguous") and enforces a clear
+    return-value contract for custom validators.
+    """
+    if result is True:
+        return True
+    if result is False or result is None:
+        return False
+    # Check for pd.NA without importing pandas at module level
+    try:
+        import pandas as _pd
+
+        if result is _pd.NA:
+            return False
+    except ImportError:
+        pass
+    raise TypeError(
+        f"Custom validator {validator_name!r} returned "
+        f"{type(result).__name__!r} ({result!r}); "
+        "validators must return True (pass) or False (fail)."
+    )
+
 
 
 def Custom(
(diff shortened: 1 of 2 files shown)


im-anishraj (changes requested):
Thanks for the focused fix. This is not ready yet.

Blocking issue: the CI matrix is failing because the new tests assert zero-based `row_index` values, but Arnio validation issues currently report row indexes as one-based. The failing cases are:
- `test_validator_returning_pd_na_fails`: expected `1`, actual `2`
- `test_mixed_true_false_none_pd_na`: expected `{1, 2, 3}`, actual `{2, 3, 4}`

Please update the tests to match the existing one-based row-index contract, then let the full matrix pass again.

ayushlad9108:
@im-anishraj Done with the changes!

ayushlad9108:
@im-anishraj Please look if there are any changes!

im-anishraj (approved):

PR #1484, opened 2026-05-25
fix: handle zero-column frames in preview

### LINKED ISSUE #1401: Bug: ArFrame.preview renders blank output for zero-column frames
## Summary
`ArFrame.preview()` returns a mostly blank table for frames that have rows but zero columns.

## Why it matters
Zero-column frames can appear after column-selection/drop operations or when constructing an `ArFrame` from an empty-column pandas DataFrame with a preserved row index. The preview output should make that state clear instead of rendering blank header/separator/row lines.

## Evidence
`arnio/frame.py::ArFrame.preview()` only special-cases `num_rows == 0`. For `(rows > 0, cols == 0)`, it builds empty headers and empty row strings.

Observed behavior with an empty-column, three-row frame:

```python
import pandas as pd
import arnio as ar

print(frame.shape)       # (3, 0)
print(repr(frame.preview()))
```

Current output:

```text
'ArFrame preview (showing 3 of 3 rows):\n\n\n\n\n'
```

## Reproduction or trigger
1. Create a zero-column frame with rows.
2. Call `frame.preview()`.
3. Observe blank table lines after the label.

## Expected behavior
`preview()` should explicitly report a zero-column frame, for example:

```text
ArFrame preview: 3 rows x 0 columns (no columns to display)
```

## Actual behavior
The preview label is followed by blank header/separator/row lines.

## Suggested direction
- Add a `num_cols == 0` branch in `ArFrame.preview()`.
- Add a regression test for `(rows > 0, columns == 0)`.
- Consider aligning `__str__()` wording with the same state, if needed.

## References
- `arnio/frame.py::ArFrame.preview`
- `tests/test_frame.py`

Bhagyashri77777:
## Summary

Fixes the blank output issue in `ArFrame.preview()` when a frame contains rows but zero columns.

Previously, preview output for zero-column frames rendered empty header and separator lines, resulting in unclear output like:

```python
'ArFrame preview (showing 3 of 3 rows):\n\n\n\n\n'

This change adds an explicit num_cols == 0 branch so preview clearly reports the frame state.

Changes
Added zero-column frame handling in ArFrame.preview()
Added regression test for frames with rows but no columns
Example

Before:

'ArFrame preview (showing 3 of 3 rows):\n\n\n\n\n'

After:

'ArFrame preview: 3 rows x 0 columns (no columns to display)'

Fixes #1401


im-anishraj:
Hi, thanks for the PR.

GitHub is not detecting the linked issue because the `Fixes #1401` text appears inside an unclosed code block in the PR description. Please close the code block and add the supported form outside any code block:

`Fixes #1401`

This is required for tracking, review, and automatic issue closure.

im-anishraj (changes requested):
Thanks for the PR. This is not ready.

Blocking issues:
- The PR is supposed to fix `ArFrame.preview()` for zero-column frames, but it also adds unrelated CSV progress-hook API/binding/C++ changes. Please remove all unrelated CSV progress changes from `arnio/io.py`, `bindings/bind_arnio.cpp`, `cpp/include/arnio/csv_reader.h`, `cpp/src/csv_reader.cpp`, and `tests/test_csv.py`.
- `tests/test_csv.py` currently contains pasted implementation code instead of tests and removes existing `TestArFrameStr` coverage.
- `arnio/io.py` has broken indentation around `try:`, which will make the module invalid.
- GitHub still does not detect `Fixes #1401` because the PR body has an unclosed code block. Please close the code block and place `Fixes #1401` outside it.

Please keep this PR limited to the `preview()` behavior and its focused regression test.

Bhagyashri77777:
Created a new clean PR with isolated changes here: #1548

im-anishraj:
Hi, thanks for the work here.

I’m closing this older PR because it still contains broad unrelated CSV/C++/website changes, and you already opened the cleaner focused replacement in #1548 for the same preview issue.

Please continue on #1548 only so the review stays scoped and fair.

PR #1491, opened 2026-05-25
fix(cleaning): handle non-string pandas column labels in drop_columns_matching

### LINKED ISSUE #1472: Bug: drop_columns_matching crashes on non-string pandas labels
## Severity
Low

## Files / lines
- `arnio/cleaning.py:1499-1554`

## Description
`drop_columns_matching()` claims to support pandas DataFrames, but it crashes when a pandas DataFrame has non-string column labels. Pandas permits integer, tuple, and other hashable labels; Arnio's direct pandas path passes those labels directly into `re.search()`.

## Why it happens
The function builds `cols_to_drop` with:

```python
cols_to_drop = [col for col in df.columns if re.search(pattern, col)]
```

`re.search()` requires a string or bytes-like object, so integer or tuple labels raise raw `TypeError` before any matching can happen.

## How this fails in real use
Notebook users commonly have integer or MultiIndex-derived tuple column labels. Calling the pandas-compatible helper can fail even when only string columns are intended to match.

Reproduction:

```python

df = pd.DataFrame([[1, 2, 3]], columns=[1, "temp_a", "keep"])
ar.drop_columns_matching(df, r"^temp")
# TypeError: expected string or bytes-like object, got 'int'
```

## Exact fix / mitigation
Decide and document the intended contract. Either reject non-string pandas column labels up front with a clear `TypeError`, or match against `str(col)` while preserving the original labels in `df.drop(columns=...)`. The latter is friendlier for pandas interop.

## Existing tests
No. Tests cover matching behavior with string labels only.

## Suggested test
Create pandas DataFrames with integer and tuple column labels plus one matching string label; assert the function either drops the matching string column or raises a clear validation error before regex execution.

sat-06:
## Summary

Fixes a pandas interoperability issue in `drop_columns_matching()` where non-string column labels caused `re.search()` to raise a `TypeError`.

### Problem

The function previously performed regex matching directly on column labels:

```python
re.search(pattern, col)
```

This failed for valid pandas column labels such as integers or tuples.

### Fix

* Updated matching logic to use `str(col)` during regex matching
* Preserved original column labels when passing them to `df.drop(columns=...)`
* Added regression coverage for integer and tuple pandas column labels

### Example

Previously failing:

```python
df = pd.DataFrame([[1, 2, 3]], columns=[1, ("sensor", "temp"), "temp_a"])
ar.drop_columns_matching(df, r"^temp")
```

Now correctly drops only `"temp_a"` without raising errors.

### Tests

* Added focused regression test for non-string pandas column labels
* Existing `drop_columns_matching()` behavior remains unchanged

### DIFF
--- a/arnio/cleaning.py
+++ b/arnio/cleaning.py
@@ -1549,7 +1549,7 @@ def drop_columns_matching(frame, pattern):
     is_arframe = not isinstance(frame, pd.DataFrame)
     df = to_pandas(frame) if is_arframe else frame
 
-    cols_to_drop = [col for col in df.columns if re.search(pattern, col)]
+    cols_to_drop = [col for col in df.columns if re.search(pattern, str(col))]
 
     if len(cols_to_drop) == len(df.columns):
         raise ValueError(
--- a/tests/test_cleaning.py
+++ b/tests/test_cleaning.py
@@ -3096,6 +3096,14 @@ def test_drop_columns_matching_normal():
     assert list(result.columns) == ["keep_c"]
 
 
+def test_drop_columns_matching_handles_non_string_pandas_columns():
+    df = pd.DataFrame([[1, 2, 3]], columns=[1, ("sensor", "temp"), "temp_a"])
+
+    result = ar.drop_columns_matching(df, "^temp")
+
+    assert list(result.columns) == [1, ("sensor", "temp")]
+
+
 def test_drop_columns_matching_no_match():
     df = pd.DataFrame({"a": [1], "b": [2]})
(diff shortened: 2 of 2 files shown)


sat-06:
Hi @im-anishraj

I noticed the Vercel check is blocked due to deployment authorization. The code-related CI checks are passing from my side. Please let me know if any changes are needed from my side.

im-anishraj (approved):
Approved. The fix is scoped to the pandas DataFrame path, preserves original column labels while matching against their string representation, includes regression coverage for integer and tuple labels, links the correct issue, and the full CI matrix is green.

PR #1494, opened 2026-05-25
fix: validate file_path type in DataQualityReport.to_html

### LINKED ISSUE #1460: Bug: validate DataQualityReport.to_html file_path before opening
## Summary
`DataQualityReport.to_html(file_path=...)` accepts truthy non-path values, including `True`, and passes them directly to `open()`. On Python, `open(True, "w")` treats `True` as file descriptor `1`, which can write the HTML report to stdout and close stdout when the context manager exits.

## Why it matters
This is a surprising side effect for a public report-export API. A boolean passed by mistake should produce a clear validation error, not mutate process-level file descriptors. In notebooks, CLIs, and CI logs, this can corrupt output and make later logging fail.

## Evidence
- `arnio/quality.py`: `DataQualityReport.to_html()` checks `if file_path:` and then calls `open(file_path, "w", encoding="utf-8")` without validating that `file_path` is a string/path-like object.
- `tests/test_quality.py`: existing tests cover valid string paths and invalid `output`, but not invalid `file_path` types.

## Reproduction or trigger
1. Create or load any `DataQualityReport`.
2. Call `report.to_html(file_path=True)`.
3. Observe that Python treats `True` as file descriptor `1`; the HTML may be written to stdout and stdout may be closed afterward.

## Expected behavior
`to_html(file_path=...)` should accept only `str`, `bytes`, or `os.PathLike` values, and should reject booleans and other invalid types with a clear `TypeError`.

## Actual behavior
Truthy non-path values are passed directly to `open()`, allowing file-descriptor semantics and confusing process-level side effects.

## Suggested direction
- Validate `file_path` before opening it.
- Explicitly reject `bool` values even though `bool` is an `int` subclass.
- Add tests for `file_path=True`, `file_path=123`, and `file_path=object()`.
- Keep valid string/path-like behavior unchanged.

## References
- `arnio/quality.py`
- `tests/test_quality.py`

Udayan6611:
**Description:**

This adds a strict type guard to `DataQualityReport.to_html` to explicitly reject bool and other non-path values, preventing them from being passed to `open()`. Valid `str`, `bytes`, and `os.PathLike` behavior is preserved.

Added regression tests for `True`, `False`, `123`, and `object()`.

Fixes #1460

### DIFF
--- a/arnio/quality.py
+++ b/arnio/quality.py
@@ -8,6 +8,7 @@ from __future__ import annotations
 import html
 import json
 import math
+import os
 from collections.abc import Sequence
 from dataclasses import dataclass, field
 from typing import Any
@@ -24,8 +25,6 @@ from .cleaning import (
 )
 from .convert import to_pandas
 from .frame import ArFrame
-
-import os
 
 
 class CleaningSuggestion(tuple):
--- a/arnio/quality.py
+++ b/arnio/quality.py
@@ -406,9 +406,13 @@ class DataQualityReport:
         automatically via ``_repr_html_``.
         """
         if file_path is not None:
-            if isinstance(file_path, bool) or not isinstance(file_path, (str, bytes, os.PathLike)):
-                raise TypeError(f"file_path must be a string, bytes, or os.PathLike object, got {type(file_path).__name__}")
-        
+            if isinstance(file_path, bool) or not isinstance(
+                file_path, (str, bytes, os.PathLike)
+            ):
+                raise TypeError(
+                    f"file_path must be a string, bytes, or os.PathLike object, got {type(file_path).__name__}"
+                )
+
         max_suggestions = self._validate_max_suggestions(max_suggestions)
         html_out = self._to_html_dashboard(
             full_document=True,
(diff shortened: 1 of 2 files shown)


im-anishraj (changes requested):
Thanks for the fix. The implementation direction is correct, but this PR is not merge-ready yet because CI formatting is failing.

Please run `black arnio/quality.py tests/test_quality.py` and then re-run the focused tests. The current diff has formatting issues around the new `os` import, a long `isinstance(...)` guard/error line, trailing whitespace, and missing blank-line spacing before the new test.

After formatting is clean and the lint check passes, I’ll recheck the implementation.

Udayan6611:
Hi @im-anishraj. I've run both `black` and `ruff` to fix the formatting, import sorting, and line lengths. 

The CI/Lint checks are now completely green, along with all the regression tests. Ready for a re-review whenever you have a moment! (Note: the Vercel bot is just waiting for maintainer authorization for the preview deployment).

im-anishraj (approved):

PR #1501, opened 2026-05-26
fix: validate encoding type before UTF-8 checks in read_csv and scan_csv

### LINKED ISSUE #1440: Bug: validate CSV encoding type before UTF-8 checks
## Summary
`read_csv()` and `scan_csv()` pass `encoding` into `_validate_csv_path()` before checking that it is a string. `_validate_csv_path()` calls `_is_utf8_encoding()`, which immediately does `encoding.lower()`. Non-string encodings therefore raise raw `AttributeError`s instead of Arnio-style validation errors.

## Why it matters
Encoding is a common ingestion option. Production APIs should reject invalid configuration with clear, stable errors, especially because `sniff_delimiter()` already validates `encoding` as a string and the docs describe encoding errors as validation failures.

## Evidence
- `arnio/io.py`: `_is_utf8_encoding(encoding)` calls `encoding.lower()` directly.
- `arnio/io.py`: `_validate_csv_path(path, encoding)` calls `_is_utf8_encoding()` before any type check.
- `arnio/io.py`: both `read_csv()` and `scan_csv()` call `_validate_csv_path(path, encoding)` near the start of their execution.
- `tests/test_csv.py` covers many encoding behavior cases, but not invalid `encoding` types for `read_csv()` / `scan_csv()`.

## Reproduction or trigger
1. Create a valid CSV file.
2. Call `ar.read_csv(path, encoding=123)`.
3. Call `ar.scan_csv(path, encoding=None)`.

## Expected behavior
Both APIs should raise a clear `TypeError`, for example `encoding must be a string`, before any filesystem or parser work begins.

## Actual behavior
Both APIs raise raw errors such as:

```text
AttributeError: 'int' object has no attribute 'lower'
AttributeError: 'NoneType' object has no attribute 'lower'
```

## Suggested direction
Add a shared `encoding` validator before `_is_utf8_encoding()` is called, or make `_is_utf8_encoding()` reject non-string values explicitly. Add tests for `read_csv()` and `scan_csv()` with `encoding=123`, `encoding=None`, lists, and arbitrary objects.

## References
- `arnio/io.py`
- `tests/test_csv.py`

SiddharthRiot:
Fixes #1440

## What changed
- Added `isinstance(encoding, str)` check in `_is_utf8_encoding()` before calling `encoding.lower()`
- Non-string encodings now raise a clear `TypeError` instead of `AttributeError`

## Tests
- Added 5 regression tests in `tests/test_csv.py` for `read_csv()` and `scan_csv()` with `encoding=123`, `encoding=None`, and `encoding=["utf-8"]`

### DIFF
--- a/arnio/io.py
+++ b/arnio/io.py
@@ -27,6 +27,8 @@ from .frame import ArFrame
 
 def _is_utf8_encoding(encoding: str) -> bool:
     """Return whether the encoding should be treated as raw UTF-8 input."""
+    if not isinstance(encoding, str):
+        raise TypeError(f"encoding must be a string, got {type(encoding).__name__!r}")
     return encoding.lower().replace("_", "-") in {"utf-8", "utf8"}
 
 
--- a/tests/test_csv.py
+++ b/tests/test_csv.py
@@ -2608,3 +2608,38 @@ class TestArFrameStr:
     def test_str_returns_string(self, sample_csv):
         frame = ar.read_csv(sample_csv)
         assert isinstance(str(frame), str)
+
+
+def test_read_csv_encoding_non_string_raises_type_error(tmp_path):
+    path = tmp_path / "test.csv"
+    path.write_text("a,b\n1,2\n")
+    with pytest.raises(TypeError, match="encoding must be a string"):
+        ar.read_csv(path, encoding=123)
+
+
+def test_read_csv_encoding_none_raises_type_error(tmp_path):
+    path = tmp_path / "test.csv"
+    path.write_text("a,b\n1,2\n")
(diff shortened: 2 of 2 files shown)


SiddharthRiot:
@im-anishraj fix is pushed and CI is green, ready for review!

im-anishraj (approved):
Approved. This is scoped to CSV encoding type validation, links #1440, includes regression coverage for read_csv and scan_csv, and the relevant CI matrix is green. The Vercel item is the external authorization check and not related to this Python change.

SiddharthRiot:
thanks!

PR #1506, opened 2026-05-26
fix: schema export no longer drops fields named strict or unique

### LINKED ISSUE #1467: Bug: schema export drops columns named strict or unique
## Severity
High

## Files / lines
- `arnio/schema_export.py:184-253`

## Description
`schema_to_dict()` and `schema_to_yaml()` cannot safely export schemas that contain real columns named `strict` or `unique`. Those field names collide with schema metadata keys and are moved to the top level or overwritten, so the exported contract silently drops columns.

## Why it happens
`schema_to_dict()` uses a single `raw` dict for both field definitions and metadata. Later, any key named `strict` or `unique` is treated as metadata at lines 227-229. For `Schema` objects, `raw["strict"] = schema.strict` and `raw["unique"] = schema.unique` overwrite fields with those names.

## How this fails in real use
A CSV or contract with legitimate columns named `strict` or `unique` exports a YAML schema missing those columns. Re-importing or sharing that contract can then skip validation for real production data fields.

Reproduction:

```python
import arnio as ar
from arnio.schema_export import schema_to_dict, schema_to_yaml

raw = {"strict": "int64", "unique": "string", "name": "string"}
print(schema_to_dict(raw))
# {'fields': {'name': {'type': 'string'}}, 'strict': 'int64', 'unique': 'string'}

schema = ar.Schema({
    "strict": ar.Int64(),
    "unique": ar.String(),
    "name": ar.String(),
}, strict=True, unique=["name"])
print(schema_to_yaml(schema))
# only 'name' appears under fields
```

## Exact fix / mitigation
Keep field definitions and schema metadata separate throughout normalization. For raw dicts from `scan_csv`, treat all keys as fields unless the input is already in an explicit structured shape such as `{"fields": ..., "strict": ...}`. For `Schema` objects, build `result = {"fields": normalized_fields}` directly instead of putting metadata into the same namespace as field names.

## Existing tests
No. Existing schema export tests do not cover columns named `strict` or `unique`.

## Suggested test
Export both a raw scan dict and an `ar.Schema` with fields named `strict` and `unique`; assert both fields remain under `fields` and metadata remains top-level.

Naveen-Boddepalli:
## Summary
`schema_to_dict()` was using a single shared `raw` dict for both field definitions and schema metadata. Any field literally named `strict` or `unique` was silently pulled out of `fields` and treated as metadata, causing the exported contract to drop those columns without any error or warning.

Fixed by separating field normalisation from metadata handling in both code paths:
- **Raw dict path**: only extract `strict`/`unique` as metadata if the input has an explicit `fields` key (structured shape). Flat scan_csv-style dicts treat all keys as field names.
- **Schema object path**: build `result["fields"]` independently and write metadata keys directly to `result`, never via the shared `raw` namespace.

## Linked issue
Fixes #1467

## Type of change
- [x] Bug fix

## Area
- [x] Schema validation

## Testing
```bash
pytest tests/test_schema_export.py -v
# All existing tests passed. Manually verified both raw dict and Schema object
# paths with fields named strict and unique.
``` 
- [x] Other: manual regression test for both paths

image
image

## Contributor checklist
- [x] My PR title uses Conventional Commits (`fix:`).

## Maintainer notes
Only `arnio/schema_export.py` is changed. No public API surface changes — same inputs and outputs, just correct behavior for the `strict`/`unique` field name edge case.
Used two test cases locally to test the changes but did not comit them.

### DIFF
--- a/arnio/schema_export.py
+++ b/arnio/schema_export.py
@@ -181,11 +181,8 @@ def schema_to_dict(schema: dict | Any) -> dict:
         If *schema* is neither a ``dict`` nor an object with a ``fields``
         attribute.
     """
-    if isinstance(schema, dict):
-        raw: dict = schema
-    elif hasattr(schema, "fields"):
-        raw = {}
-
+    # ── Schema object path ──────────────────────────────────────────────────
+    if hasattr(schema, "fields"):
         if getattr(schema, "rules", None):
             raise ValueError(
                 "schema_to_yaml does not support Schema objects with custom rules "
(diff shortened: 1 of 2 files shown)


im-anishraj (changes requested):
Thanks for the focused fix direction. This cannot be merged yet because the regression tests from #1467 are missing.

Please add tests for both paths described in the issue:
- a raw schema dict with fields named `strict` and `unique`
- a `Schema` object with fields named `strict` and `unique`, while schema metadata remains top-level

The PR body says these cases were tested locally but not committed, so the behavior is currently unprotected against regression.

Naveen-Boddepalli:
I have added the test cases as requested, please review this PR.
thank you.

im-anishraj (approved):

PR #1513, opened 2026-05-26
fix(ci): clarify aggregate wheel matrix status in release dry-run summary

### LINKED ISSUE #1442: CI: make release dry-run wheel summary platform-aware
## Summary
The release dry-run workflow builds wheels across a matrix of operating systems, but the final summary reports Linux, macOS, and Windows using the same aggregate `needs.build_wheels.result`. If one matrix leg fails, all three rows show the same status, so maintainers cannot tell which platform failed from the summary table.

## Why it matters
Release dry-runs are meant to answer whether a PyPI release is safe. A summary that collapses all wheel platforms into one status makes platform-specific failures harder to diagnose and can slow down release decisions.

## Evidence
- `.github/workflows/release-dry-run.yml`: `build_wheels` uses `matrix.os` for `ubuntu-latest`, `windows-latest`, and `macos-latest`.
- `.github/workflows/release-dry-run.yml`: the `summary` job has only `needs.build_wheels.result`, which is an aggregate job result.
- `.github/workflows/release-dry-run.yml`: the summary writes separate Linux/macOS/Windows rows, but each row uses the same aggregate expression.

## Reproduction or trigger
1. Run the `Release Dry-Run Checklist` workflow.
2. Have only one platform leg fail in the `build_wheels` matrix.
3. Inspect the generated step summary.

## Expected behavior
The summary should identify which wheel platform passed or failed, or avoid per-platform rows when only an aggregate status is available.

## Actual behavior
The Linux, macOS, and Windows rows all display the same aggregate status from `needs.build_wheels.result`.

## Suggested direction
Either:

- split the matrix into separate jobs with separate `needs` entries for Linux/macOS/Windows, or
- keep one aggregate `Wheels + smoke` row in the summary, or
- generate a matrix-aware artifact/status report and read it in the summary job.

Add a short workflow comment so future maintainers understand why the summary is aggregate or platform-specific.

## References
- `.github/workflows/release-dry-run.yml`

sat-06:
## Summary

Fixes a misleading status summary in the release dry-run workflow where Linux, macOS, and Windows wheel rows all displayed the same aggregate matrix result from `needs.build_wheels.result`.

### Problem

`build_wheels` is implemented as a matrix job across:

* ubuntu-latest
* macos-latest
* windows-latest

However, the summary section used:

```yaml
needs.build_wheels.result
```

for separate Linux/macOS/Windows rows, even though this value represents the aggregate matrix result rather than platform-specific statuses.

This could incorrectly imply that all platforms failed when only a single matrix leg failed.

### Changes made

* Replaced the misleading per-platform wheel rows with a single aggregate matrix row
* Added a workflow comment clarifying that:

  * `needs.build_wheels.result` is aggregate for matrix jobs
  * the summary intentionally reports aggregate wheel status

### Result

The release dry-run summary now accurately reflects the available workflow status information without implying platform-specific results that are not actually available.

### DIFF
--- a/.github/workflows/release-dry-run.yml
+++ b/.github/workflows/release-dry-run.yml
@@ -99,9 +99,9 @@ jobs:
           echo "|-------|--------|" >> $GITHUB_STEP_SUMMARY
           echo "| Version validation     | ${{ needs.validate-version.result == 'success' && '✅ Pass' || '❌ Fail' }} |" >> $GITHUB_STEP_SUMMARY
           echo "| sdist build + twine    | ${{ needs.build_sdist.result == 'success' && '✅ Pass' || '❌ Fail' }} |" >> $GITHUB_STEP_SUMMARY
-          echo "| Wheels + smoke (Linux) | ${{ needs.build_wheels.result == 'success' && '✅ Pass' || '❌ Fail' }} |" >> $GITHUB_STEP_SUMMARY
-          echo "| Wheels + smoke (macOS) | ${{ needs.build_wheels.result == 'success' && '✅ Pass' || '❌ Fail' }} |" >> $GITHUB_STEP_SUMMARY
-          echo "| Wheels + smoke (Win)   | ${{ needs.build_wheels.result == 'success' && '✅ Pass' || '❌ Fail' }} |" >> $GITHUB_STEP_SUMMARY
+          # build_wheels is a matrix job, so needs.build_wheels.result
+          # represents the aggregate status across all wheel platforms.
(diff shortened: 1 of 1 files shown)


sat-06:
Hi @im-anishraj 
I have implemented the solution for the issue and have raised a pr.
Plz let me know if there are any changes from my side

im-anishraj (approved):
Approved. This is a focused release dry-run summary fix for #1442, removes the misleading per-platform rows for the aggregate wheel matrix result, and the Arnio CI matrix is green.

PR #1535, opened 2026-05-26
fix(cleaning): validate scalar comparison values in filter_rows

### LINKED ISSUE #1432: Bug: validate scalar filter_rows comparison values
## Summary

`filter_rows()` documents `value` as a scalar, but it does not reject list-like values before comparing with pandas. Some invalid values leak pandas errors such as `ValueError: Lengths must be equal`.

## Why it matters

Filtering is a common scripting operation. Passing a list by mistake should produce a clear Arnio error explaining that only scalar comparisons are supported.

## Evidence

- `arnio/cleaning.py::filter_rows()` calls the pandas comparison method directly.
- It catches `TypeError`, but not `ValueError` from list-like length mismatches.
- The docstring says `value : object` but describes it as a scalar comparison value.

## Reproduction or trigger

```python
import pandas as pd
import arnio as ar

frame = ar.from_pandas(pd.DataFrame({"a": [1, 2, 3]}))
ar.filter_rows(frame, "a", "==", [1, 2])
```

Current result:

```text
ValueError: Lengths must be equal
```

## Expected behavior

List-like comparison values should either be explicitly supported or rejected clearly:

```text
TypeError: filter_rows value must be a scalar
```

## Actual behavior

A raw pandas length-mismatch error is exposed.

## Suggested direction

- Validate `value` with `pandas.api.types.is_scalar()` or a similar helper.
- Reject list, tuple, dict, Series, Index, and ndarray values unless explicit vectorized behavior is intentionally added.
- Add tests for list-like values and keep scalar comparisons unchanged.

## References

- `arnio/cleaning.py`
- `tests/test_filter_rows_invalid_comparison.py`
- `tests/test_cleaning.py`

sat-06:
## Summary

Fixes an input-validation issue in `filter_rows()` where list-like comparison values leaked raw pandas length-mismatch errors instead of raising a clear Arnio error.

### Problem

`filter_rows()` documents `value` as a scalar comparison value, but list-like inputs were passed directly into pandas comparison operations.

Example:

```python
frame = ar.from_pandas(pd.DataFrame({"a": [1, 2, 3]}))
ar.filter_rows(frame, "a", "==", [1, 2])
```

Previously raised:

```txt
ValueError: Lengths must be equal
```

from pandas internals.

### Changes made

* Added explicit scalar validation for the `value` argument using `pandas.api.types.is_scalar`
* Rejected list-like comparison values with a clear:

  ```txt
  TypeError: filter_rows value must be a scalar
  ```
* Added focused regression coverage for:

  * list
  * tuple
  * dict
  * pandas Series
  * pandas Index
  * NumPy ndarray

### Result

`filter_rows()` now validates its documented scalar-comparison contract before calling pandas comparison logic, preventing raw pandas length-mismatch errors from leaking to users.

### DIFF
--- a/arnio/cleaning.py
+++ b/arnio/cleaning.py
@@ -1263,6 +1263,8 @@ def filter_rows(
 
     import pandas as pd
 
+    from pandas.api.types import is_scalar
+
     from .convert import from_pandas, to_pandas
 
     is_arframe = not isinstance(frame, pd.DataFrame)
@@ -1283,6 +1285,9 @@ def filter_rows(
 
     if column not in df.columns:
         raise ValueError(f"Unknown column: {column}")
+    
+    if not is_scalar(value):
+        raise TypeError("filter_rows value must be a scalar")
 
     try:
         mask = getattr(df[column], ops[op])(value)
--- a/arnio/cleaning.py
+++ b/arnio/cleaning.py
@@ -1262,7 +1262,6 @@ def filter_rows(
     """
 
     import pandas as pd
-
     from pandas.api.types import is_scalar
 
     from .convert import from_pandas, to_pandas
@@ -1285,7 +1284,7 @@ def filter_rows(
 
     if column not in df.columns:
         raise ValueError(f"Unknown column: {column}")
-    
+
     if not is_scalar(value):
         raise TypeError("filter_rows value must be a scalar")
 
--- a/tests/test_cleaning.py
+++ b/tests/test_cleaning.py
@@ -2268,6 +2268,22 @@ class TestFilterRows:
         ):
             ar.filter_rows(df, "name", ">", 1)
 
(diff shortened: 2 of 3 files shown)


im-anishraj (changes requested):
Thanks for the PR. This is not ready for review yet because the current CI matrix is failing, including lint and platform test jobs. Please fix the failing checks first, then request review again.

sat-06:
Hi @im-anishraj 
Thanks for clarifying.
All the tests are now passing and the cl matrix is green.

im-anishraj (approved):
Reviewed the scalar-value guard for filter_rows. The implementation matches the current comparison-only filter_rows API, adds regression coverage, passes local ruff/black, and the full Arnio CI matrix is green apart from the known Vercel authorization check.

PR #1614, opened 2026-05-27
fix(schema): reject bare string values in String(allowed=...)

### LINKED ISSUE #1417: Bug: reject bare strings in String allowed values
## Summary

`String(allowed=...)` accepts a bare string and converts it with `set(allowed)`, which turns the string into a set of characters instead of treating it as one allowed value or rejecting it.

## Why it matters

This can silently corrupt schema intent. A user may write `allowed="active"` expecting only the value `"active"`, but Arnio interprets it as allowed characters `{ "a", "c", "t", "i", "v", "e" }`.

## Evidence

- `arnio/schema.py::String()` accepts `allowed: set[Any] | list[Any] | tuple[Any, ...] | None`.
- The implementation uses `allowed_set = set(allowed)` whenever `allowed` is not `None`.
- A bare `str` is iterable, so it is split into characters.

## Reproduction or trigger

```python
import pandas as pd
import arnio as ar

frame = ar.from_pandas(pd.DataFrame({"status": ["active", "a"]}))
schema = {"status": ar.String(allowed="active")}
print(result.to_dict())
```

The configuration is accepted, but it represents the allowed set incorrectly.

A simpler observed case:

```python
field = ar.String(allowed="ab")
print(field.allowed)
```

Current result:

```python
{"a", "b"}
```

## Expected behavior

A bare string should be rejected clearly:

```text
TypeError: allowed must be a sequence of allowed values, not a bare string
```

If single-value support is desired, users can pass `allowed=["active"]`.

## Actual behavior

Bare strings are accepted and split into characters.

## Suggested direction

- Reject `str` and `bytes` before calling `set(allowed)`.
- Add tests for `allowed="active"`, `allowed=["active"]`, `allowed=("active",)`, and `allowed={"active"}`.
- Keep list/tuple/set behavior unchanged.

## References

- `arnio/schema.py`
- `tests/test_schema.py`

sat-06:
## Summary

Fixes a schema-configuration issue where `String(allowed=...)` accepted bare string values and incorrectly converted them into character sets using `set(allowed)`.

### Problem

The previous implementation converted any non-`None` `allowed` value using:

```python id="t0gml3"
allowed_set = set(allowed)
```

This worked correctly for sequences such as:

```python id="dj4egq"
["active"]
("active",)
{"active"}
```

but incorrectly accepted bare strings:

```python id="0n3k11"
ar.String(allowed="active")
```

which produced:

```python id="8uzhll"
{"a", "c", "t", "i", "v", "e"}
```

instead of treating `"active"` as a single allowed value.

### Changes made

* Added explicit validation to reject bare `str` and `bytes` values before converting `allowed` into a set
* Raised a clear:

  ```txt
  TypeError: allowed must be a sequence of allowed values, not a bare string
  ```
* Added focused regression coverage for:

  * `allowed="active"`
  * `allowed=b"active"`
  * `allowed=["active"]`
  * `allowed=("active",)`
  * `allowed={"active"}`

### Result

`String(allowed=...)` now rejects ambiguous bare string configurations while preserving existing list/tuple/set behavior unchanged.

### DIFF
--- a/arnio/schema.py
+++ b/arnio/schema.py
@@ -1692,6 +1692,11 @@ def String(
     if min_length is not None and max_length is not None and min_length > max_length:
         raise ValueError("min_length must be less than or equal to max_length")
 
+    if isinstance(allowed, (str, bytes)):
+        raise TypeError(
+            "allowed must be a sequence of allowed values, not a bare string"
+        )
+
     allowed_set = set(allowed) if allowed is not None else None
 
     return Field(
--- a/tests/test_schema.py
+++ b/tests/test_schema.py
@@ -1521,6 +1521,32 @@ def test_string_max_length_boundary(tmp_path):
     assert result.issues[0].row_index == 2
 
 
+def test_string_allowed_rejects_bare_string():
+    with pytest.raises(
+        TypeError,
+        match="allowed must be a sequence of allowed values, not a bare string",
+    ):
+        ar.String(allowed="active")
+
+
+def test_string_allowed_rejects_bare_bytes():
+    with pytest.raises(
+        TypeError,
+        match="allowed must be a sequence of allowed values, not a bare string",
+    ):
+        ar.String(allowed=b"active")
+
+
+def test_string_allowed_accepts_list_tuple_and_set():
+    list_field = ar.String(allowed=["active"])
+    tuple_field = ar.String(allowed=("active",))
+    set_field = ar.String(allowed={"active"})
+
+    assert list_field.allowed == {"active"}
+    assert tuple_field.allowed == {"active"}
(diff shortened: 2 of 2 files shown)


sat-06:
Hi @im-anishraj 
I have implemented the solution for the given issue.
All the test passed locally and the cl matrix is green.

im-anishraj (approved):
Reviewed the linked issue, implementation, and tests. The PR correctly rejects bare string/bytes allowed values while preserving list/tuple/set behavior, merges cleanly into current main, and passes local diff/Ruff/Black/py_compile checks plus GitHub CI. Approved.

PR #1626, opened 2026-05-27
perf(cleaning): optimize `drop_duplicates()` with collision-safe typed hashing

### LINKED ISSUE #1623: Performance: replace ostringstream row_key with numeric hash in drop_duplicates
### Workload

Any `ar.drop_duplicates()` call on a tall dataset. Most visible on
frames with 100K+ rows regardless of column count.

The bottleneck is inside `drop_duplicates()` in
`cpp/src/cleaning.cpp` and affects all keep modes:
- `"first"`
- `"last"`
- `"none"`

Benchmark environment:
- Linux (Kali)
- Python benchmark runner
- 1M-row synthetic CSV datasets
- Mixed integer, float, string, multiline, and sparse-null workloads

`drop_duplicates()` currently uses a `row_key()` helper that serialises
every row into a `std::string` using `std::ostringstream` before hashing:

```cpp
std::ostringstream oss;

for (size_t ci : cols) {
    serialize_cell(oss, cell);
    oss << "\x1F";
}

return oss.str();
```

This creates significant overhead on large datasets:
- repeated `ostringstream` construction/destruction
- heap allocations for temporary strings
- hashing large serialised row buffers
- poor cache locality using `unordered_set<std::string>`

Pandas hashes typed values directly without serialising rows into text.

### Before optimization

Benchmark command:

```bash
python benchmarks/benchmark_vs_pandas.py
```

| Dataset | pandas | arnio | speedup |
|---|---|---|---|
| Tall 1M × 12 | 0.52s | 1.94s | 0.3x |
| Sparse-null 1M × 6 | 0.46s | 2.93s | 0.2x |
| Multiline 100K × 4 | 0.03s | 0.27s | 0.1x |

### After optimization

| Dataset | pandas | arnio | speedup |
|---|---|---|---|
| Tall 1M × 12 | 0.46s | 0.28s | 1.7x |
| Dense-null 1M × 6 | 0.17s | 0.11s | 1.6x |
| Wide 5K × 256 | 0.02s | 0.00s | 521x |
| String columns 1M × 2 | 0.478s | 0.232s | 2.06x |

### Screenshots

Image

The optimization removes the row-serialization bottleneck and greatly
improves deduplication throughput on large datasets.

### Target improvement

Replace `std::string` row keys with lightweight `uint64_t` hashes
computed directly from typed column values.

Hashing strategy:
- `INT64`: hash integer bytes directly
- `FLOAT64`: bit-cast to `uint64_t` before hashing
- `BOOL`: hash single-byte value
- `STRING`: hash raw bytes using FNV-1a
- `NULL`: fixed sentinel hash

Combine per-column hashes using multiply-xor mixing.

Replace:

```cpp
unordered_set<std::string>
```

with:

```cpp
unordered_set<uint64_t>
```

and similarly for internal maps.

This removes per-row heap allocation and avoids expensive text
serialisation during deduplication.

Scope of change:
- self-contained inside `cpp/src/cleaning.cpp`
- no Python API changes
- no pybind11 binding changes
- no dataframe behavior changes

Expected outcome:
- significant speedup on tall datasets
- improved cache locality
- reduced temporary allocation overhead
- performance closer to or exceeding pandas on some workloads

## Acceptance criteria

- [ ] All existing tests pass (`make test`)
- [ ] New C++ unit tests added in `cpp/tests/test_cleaning.cpp`
- [ ] Integer deduplication correctness verified
- [ ] Float deduplication correctness verified
- [ ] String deduplication correctness verified
- [ ] Mixed-type deduplication correctness verified
- [ ] `keep="first"` produces correct output
- [ ] `keep="last"` produces correct output
- [ ] `keep="none"` produces correct output
- [ ] Before/after benchmark numbers included in PR description
- [ ] No behavior changes for existing `drop_duplicates()` calls
- [ ] No Python API or pybind11 binding changes

dynamo-pentester:
### Summary

Optimizes `drop_duplicates()` in `cpp/src/cleaning.cpp` by replacing
eager per-row string serialization with a collision-safe typed hashing fast path.

The previous implementation serialized every row into a heap-allocated
`std::string` using `std::ostringstream` before deduplication. This PR replaces
that hot path with direct typed `uint64_t` hashing while preserving the existing
correctness semantics through lazy equality verification using the existing
`row_key()` encoding.

The optimization is fully internal to the C++ cleaning engine:
- no Python API changes
- no pybind11 binding changes
- no dataframe behavior changes

---

## Linked issue

Fixes #1623

---

### Problem

The previous implementation hashed each row by serializing it into a
`std::string` via `std::ostringstream`.

This caused several compounding performance costs on large datasets:

- repeated construction/destruction of `ostringstream`
- heap allocation for serialized row strings
- O(row_width) string hashing cost
- poor cache locality from temporary allocations

For large frames, most execution time inside `drop_duplicates()` was spent
building temporary row strings rather than performing the deduplication logic itself.

---

### Updated design

The revised implementation introduces a two-stage deduplication strategy:

1. **Fast typed hashing**
   - rows are hashed directly from typed column values
   - no temporary row string allocation
   - zero-allocation fast path for unique rows

2. **Collision-safe equality verification**
   - hash matches trigger lazy `row_key()` comparison
   - preserves the exact correctness semantics of the previous implementation
   - prevents silent incorrect deduplication from rare hash collisions

This preserves the existing invariant:

> two rows are duplicates iff their serialized `row_key()` strings are identical

The hash is therefore used strictly as an acceleration layer, not as the final deduplication authority.

---

### Hashing implementation

Each column type is hashed directly from its typed storage:

- `INT64`
- `FLOAT64`
- `BOOL`
- `STRING`
- `NULL`

String columns use FNV-1a hashing:

```cpp
uint64_t fnv1a(const char* data, size_t len) {
    uint64_t h = 14695981039346656037ULL;
    for (size_t i = 0; i < len; ++i) {
        h ^= static_cast<uint8_t>(data[i]);
        h *= 1099511628211ULL;
    }
    return h;
}
```

Per-column hashes are combined incrementally:

```cpp
h ^= hash_cell(cell);
h *= 1099511628211ULL;
```

Distinct type seeds ensure:
- `INT64(1) != STRING("1")`
- `BOOL(true) != INT64(1)`
- `NULL` remains distinct

Column order also remains significant.

---

### Collision handling

Hash collisions are handled safely using bucketed equality verification.

On hash match:
- existing rows in the bucket lazily materialize their `row_key()`
- the incoming row is compared against the serialized keys
- only exact matches are treated as duplicates

Distinct rows sharing the same hash remain separate entries in the bucket.

This guarantees no behavior changes relative to the previous implementation.

---

#### Before

| Dataset | pandas | arnio | speedup |
|---|---|---|---|
| Tall 1M × 12 | 0.52s | 1.94s | 0.3× |
| Sparse-null 1M × 6 | 0.46s | 2.93s | 0.2× |
| Multiline 100K × 4 | 0.03s | 0.27s | 0.1× |

#### After

| Dataset | upstream arnio | patched | improvement |
|---|---|---|---|
| Tall 1M × 12 | 1.94s | 0.35s | ~5.5× |
| Sparse-null 1M × 6 | 2.93s | 0.67s | ~4.3× |
| Multiline 100K × 4 | 0.27s | 0.05s | ~5.4× |

Benchmarked with:

```bash
python benchmarks/benchmark_drop_duplicates.py
python benchmarks/benchmark_vs_pandas.py
```

The optimization is most noticeable on:
- tall datasets
- sparse-null workloads
- wide frames with many columns

where the previous implementation spent substantial time allocating and hashing serialized row strings.

---

### Changes

- rewrote `drop_duplicates()` fast path in `cpp/src/cleaning.cpp`
- added typed hash helpers
- added collision-safe fallback verification
- added lazy `row_key()` materialization
- added expanded correctness tests for all keep modes and edge cases

No public API changes.

---

### Testing

- [x] `make test`
- [x] Added correctness tests for:
  - `INT64`
  - `FLOAT64`
  - `STRING`
  - `BOOL`
  - mixed-type rows
  - null rows
  - column-order sensitivity
- [x] Added tests for:
  - `keep="first"`
  - `keep="last"`
  - `keep="none"`
- [x] Benchmarks confirm significant improvement over upstream implementation
- [x] `clang-format` applied to modified files
- [ ] `make lint` not run locally

---

### Type of change

- [x] Performance improvement
- [x] Refactor
- [x] Tests

---

### Area

- [x] C++ cleaning engine

---

### Maintainer notes

The implementation preserves the existing `row_key()` equality semantics while using typed hashing strictly as a fast-path optimization layer.

All collision handling paths are verified through serialized equality fallback before rows are treated as duplicates.


im-anishraj (changes requested):
Thanks for the performance work. I can't approve this yet because the fast hash path is not guaranteed to preserve the existing `row_key()` equality semantics.

The fallback only compares `row_key()` after a hash hit. For `FLOAT64`, `hash_cell()` hashes the raw `double` bytes (`cpp/src/cleaning.cpp:139`), while `row_key()` serializes doubles through `combine_cell_to_string()` (`cpp/src/cleaning.cpp:56` and `cpp/src/cleaning.cpp:95`). Two NaN values with different payload bits can serialize to the same `row_key()` but produce different raw-byte hashes, so they are treated as distinct rows and the collision-safe fallback never runs. That changes `drop_duplicates()` behavior for `keep="first"`, `keep="last"`, and `keep="none"`.

Please normalize floating values before hashing so every `row_key()`-equivalent `FLOAT64` value has the same hash, or fall back to `row_key()` for floats/NaNs where normalization is not guaranteed. Add regression coverage for NaN-equivalent float rows before this is merged.

dynamo-pentester:
Thanks for the thorough review @im-anishraj — you're right, the invariant was broken for NaN.

**Root cause**
`hash_cell()` hashed raw `double` bytes, so two NaN values with different bit-pattern payloads produced different hashes. Meanwhile `combine_cell_to_string()` formats all NaN variants through `%.17g`, which always renders `"nan"` regardless of payload, so `row_key()` considered them equal. The consequence: a hash miss on two NaN rows skipped the `row_key()` equality check entirely and kept both rows as distinct — silently failing to deduplicate them.

**Fix (`cleaning.cpp` — `hash_cell()`)**
NaN is now normalized to a single canonical quiet-NaN bit pattern (`0x7FF8000000000000`) before hashing, restoring the invariant that `hash(x) == hash(y)` whenever `row_key(x) == row_key(y)` for all `FLOAT64` values:
```cpp
if (std::isnan(v)) {
    const uint64_t canonical_nan_bits = 0x7FF8000000000000ULL;
    return fnv1a(reinterpret_cast<const char*>(&canonical_nan_bits),
                 sizeof(canonical_nan_bits),
                 14695981039346656037ULL ^ 0x02ULL);
}
```
Non-NaN doubles are unaffected — still hashed from raw bytes with zero overhead.

**Regression tests added (`test_cleaning.cpp`)**
Three new tests covering the exact failure case, one per `keep` mode (`first`, `last`, `none`). Each constructs two NaN values with distinct bit-pattern payloads and asserts they are treated as duplicates.

**Why no other FLOAT64 edge cases need fixing**
- `±infinity`: `+inf` and `-inf` each have a single IEEE 754 bit pattern and serialize to distinct strings (`"inf"` / `"-inf"`). Hash and key agree. ✓
- `-0.0 vs 0.0`: `%.17g` renders them as `"-0"` and `"0"` respectively (different strings), and their raw bits also differ. Hash and key both treat them as distinct. ✓
- All other non-NaN doubles: IEEE 754 guarantees 17 significant decimal digits uniquely identifies every distinct double, so different bit patterns always produce different `%.17g` strings. ✓

NaN was the only type where multiple bit patterns collapsed to the same serialized string.

im-anishraj:
Thanks for the quick update. The new commit still cannot be merged because CI is red.

`C++ Native Tests (CMake + Catch2)` is failing in `drop_duplicates NaN payloads treated as equal (keep=first)`:

```text
cpp/tests/test_cleaning.cpp:308
REQUIRE( result.num_rows() == 2 )
with expansion: 3 == 2
```

The likely issue is that the test assumes positive NaN and negative NaN serialize to the same `row_key()`. On the CI compiler, the negative NaN appears to remain distinct under the existing serialized-key behavior, so `drop_duplicates()` keeps it as a separate row. Since this PR is meant to preserve existing behavior, please only assert deduplication for NaN values that are actually `row_key()`-equivalent, or make an explicit, separately tested behavior change if all NaN signs/payloads should be considered equal.

Please fix the failing C++ test and wait for the full CI matrix to pass before requesting another review.

im-anishraj (approved):
Thanks for the fix. I rechecked the latest diff and CI status.

The NaN hashing invariant is now covered with positive NaN payload regression tests, C++ native tests are passing again, and the change remains scoped to `drop_duplicates()` plus coverage. The remaining Vercel status is an authorization status and not related to this C++ library change.

dynamo-pentester:
Thanks a lot for the detailed reviews and guidance throughout this PR, @im-anishraj 🙌

PR #1635, opened 2026-05-28
fix: pipeline dry_run validates against intermediate transformed frame

### LINKED ISSUE #1627: Bug: pipeline dry_run validates later steps against the original frame
I hit this while checking a multi-step pipeline with `dry_run=True` for GSSoC 2026. The dry run executes each step, but later steps still validate against the original `ArFrame` instead of the output of the previous step.

That makes dry-run validation misleading: a pipeline can fail in dry-run even though the real run succeeds, or pass dry-run even though an earlier transform would change the shape/schema before a later step.

Line: `arnio/pipeline.py:342`

A small example of the issue:

```python
import pandas as pd
import arnio as ar

frame = ar.from_pandas(pd.DataFrame({"old_name": [" Alice "]}))

steps = [
    ("rename_columns", {"old_name": "name"}),
    ("validate_columns_exist", {"columns": ["name"]}),
]

# This should validate the transformed output of step 1.
ar.pipeline(frame, steps, dry_run=True)
```

Current behavior:
The second step still sees the original frame with `old_name`, so it can report that `name` is missing.

Expected behavior:
`dry_run=True` should still walk through the transformed intermediate frame for validation purposes. It just should not return the transformed output as the final result.

Why it matters:
Dry-run is exactly what people will use before running a pipeline on real data. If it validates a different state than the real pipeline, it becomes hard to trust for CI/data contract checks.

Suggested fix:
Keep a temporary result during dry-run and advance it after each successful step, while preserving the public dry-run return behavior. Add a regression test with a schema-changing first step followed by a dependent second step.

Pcmhacker-piro:
## PR Description

Fixes #1627

## Summary

`pipeline(frame, steps, dry_run=True)` executed each step but validated all later steps against the **original** `ArFrame` instead of the transformed intermediate frame produced by prior steps.

This caused dry-run mode to incorrectly fail for valid multi-step pipelines where earlier steps modified the schema before later validation steps.

### Example

```python

steps = [

    ("rename_columns", {"old_name": "name"}),  # step 1 renames column

    ("validate_columns_exist", {"columns": ["name"]}),  # step 2 checks renamed column

]

ar.pipeline(frame, steps, dry_run=True)

```

Before this fix:

- step 2 still validated against the original frame

- `name` appeared missing

- dry-run falsely failed

Real execution (`dry_run=False`) succeeded correctly.

---

## Root Cause

The pipeline execution flow only advanced:

```python

result = step_result

```

when:

```python

dry_run == False

```

During dry-run mode, every later step re-read from the original frame instead of the intermediate transformed state.

---

## Fix

Introduced an internal `working_frame` variable that advances after every successful step execution, regardless of `dry_run`.

### Important behavior preserved

- `pipeline(..., dry_run=True)` still returns the original unmodified frame

- no external mutation behavior changed

- only internal validation state progression changed

This preserves the public API contract while making dry-run validation accurate.

---

## What changed

| File | Changes |

|---|---|

| `arnio/pipeline.py` | Added `working_frame` progression logic for dry-run execution |

| `tests/test_pipeline.py` | Added regression tests covering intermediate-frame validation behavior |

---

## Added regression coverage

Tests now cover:

- rename → validate flow

- drop_columns → validate flow

- chained schema-changing operations

- dry-run consistency with real execution

- metadata preservation

- external mutation safety

- Python/custom step execution paths

- cast/transform validation chains

---

## Why this fix is safe

The change is isolated to internal dry-run execution flow.

- No public API changes

- No behavior changes for normal execution

- No persistence/mutation changes

- Existing execution, streaming, and transform logic remain unchanged

The fix only ensures later dry-run steps validate against the correct intermediate state.

---

## Type of change

- [x] Bug fix

- [ ] New feature

- [ ] Breaking change

- [ ] Documentation update

---

## Validation

- Added dedicated regression tests

- Existing pipeline behavior preserved

- Multi-step dry-run pipelines now behave consistently with real execution

---

## Result

`dry_run=True` now correctly simulates step-by-step pipeline execution while still returning the original frame as intended.

### DIFF
--- a/arnio/pipeline.py
+++ b/arnio/pipeline.py
@@ -340,6 +340,7 @@ def pipeline(
     )
 
     result = frame
+    working_frame = frame
 
     step_timings: list[dict[str, Any]] = []
     applied_steps: list[str] = []
@@ -372,26 +373,29 @@ def pipeline(
             if name == "rename_columns" and (
                 "mapping" not in kwargs or not isinstance(kwargs["mapping"], dict)
             ):
-                step_result = fn(result, mapping=kwargs)
+                step_result = fn(working_frame, mapping=kwargs)
 
                 if not dry_run:
                     result = step_result
+                working_frame = step_result
 
             elif name == "cast_types" and (
                 "mapping" not in kwargs or not isinstance(kwargs["mapping"], dict)
             ):
-                step_result = fn(result, kwargs)
+                step_result = fn(working_frame, kwargs)
 
                 if not dry_run:
                     result = step_result
+                working_frame = step_result
 
             else:
-                target_frame = result
+                target_frame = working_frame
 
                 step_result = fn(target_frame, **kwargs)
 
                 if not dry_run:
                     result = step_result
+                working_frame = step_result
 
             elapsed_sec = perf_counter() - started_at
             elapsed_ms = elapsed_sec * 1000
@@ -434,7 +438,7 @@ def pipeline(
 
             fn = python_step_registry[name]
 
-            df = to_pandas(result)
+            df = to_pandas(working_frame)
 
             # Isolate genuine custom steps from internal core library functions
             is_builtin = _is_builtin_python_step(name, fn)
@@ -469,6 +473,7 @@ def pipeline(
             step_result = from_pandas(returned)
             if not dry_run:
                 result = step_result
+            working_frame = step_result
 
             elapsed_sec = perf_counter() - started_at
             elapsed_ms = elapsed_sec * 1000
--- a/tests/test_pipeline.py
+++ b/tests/test_pipeline.py
@@ -840,6 +840,176 @@ class TestPipeline:
         assert row_counts[0]["dry_run"] is True
 
 
+class TestPipelineDryRunIntermediateFrame:
+    """Regression tests for dry_run validating against intermediate frame."""
+
+    def test_dry_run_rename_then_validate(self):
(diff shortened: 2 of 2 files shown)


im-anishraj (changes requested):
Thanks for the clear write-up and the regression coverage. The dry-run intermediate-frame direction makes sense, but I can’t merge this PR as-is because it adds a new `uv.lock` file with ~1,900 unrelated dependency-lock lines.

Please remove `uv.lock` from this PR and keep the change focused on:

- `arnio/pipeline.py`
- the focused regression tests in `tests/test_pipeline.py`

After that, I’ll re-run the pipeline tests and review the implementation again. The Vercel failure is not relevant to this library change, but the unrelated lockfile needs to be cleaned up before merge.

im-anishraj (approved):
Reviewed the focused dry-run pipeline fix.

Verification:

- Clean merge into current `origin/main` locally.
- `git diff --check` passed.
- `ruff check arnio/pipeline.py tests/test_pipeline.py` passed.
- `black --check arnio/pipeline.py tests/test_pipeline.py` passed.
- `py_compile` passed.
- GitHub Actions CI, Benchmark, and Examples Smoke passed on the latest head.

Local pytest could not run on this machine because the native `_arnio_cpp` extension is not installed in the review checkout, so I used the completed GitHub CI matrix as the runtime verification source.

PR #1736, opened 2026-05-28
Fix: Do not apply encoding twice for text file-like inputs

### LINKED ISSUE #1663: Bug: CSV readers apply encoding twice for text file-like inputs
## Summary

`read_csv()` and `read_csv_chunked()` support text file-like objects by copying their already-decoded text into a temporary UTF-8 CSV file. However, the rest of the read path still treats that temporary file as if it were encoded with the user-provided `encoding`.

This means text streams can fail or be misinterpreted when `encoding` is not UTF-8, even though text file-like input has already been decoded by the caller.

## Why it matters

Users often pass `io.StringIO`, web framework upload wrappers, or already-decoded text streams into CSV readers. For those inputs, `encoding` should not be applied a second time to the temporary UTF-8 file.

## Evidence

- `arnio/io.py::_materialize_csv_input()` writes file-like text to a temporary file with `encoding="utf-8"`.
- `arnio/io.py::read_csv()` later calls `_utf8_csv_path(path, encoding, ...)` using the original user-supplied `encoding`.
- `arnio/io.py::read_csv_chunked()` follows the same materialize-then-transcode path.
- For text streams, that can decode the UTF-8 temp file as another encoding.

## Reproduction or trigger

```python
import io
import arnio as ar

csv = io.StringIO("name\ncafé\n")
ar.read_csv(csv, encoding="utf-16")
```

Current behavior raises:

```text
CsvReadError: Could not decode '<temp>.csv' using encoding 'utf-16'
```

## Expected behavior

For text file-like inputs, Arnio should treat the materialized temporary file as UTF-8 because the stream has already produced Python strings.

## Actual behavior

The original `encoding` is applied again to the UTF-8 temp file.

## Suggested direction

Have `_materialize_csv_input()` return enough metadata to distinguish original filesystem paths from materialized text streams. When the source is a text file-like object, use UTF-8 for subsequent temp-file validation/transcoding regardless of the caller's `encoding`.

Add tests for:

- `io.StringIO` with `encoding="utf-16"`
- `io.StringIO` with non-ASCII text
- `read_csv_chunked()` with the same file-like scenario
- filesystem path input still honoring `encoding`
- temporary file cleanup on decoding errors

## References

- `arnio/io.py::_materialize_csv_input`
- `arnio/io.py::read_csv`
- `arnio/io.py::read_csv_chunked`
- `arnio/io.py::_utf8_csv_path`

Exodus2004:
Fixes #1663

### DIFF
--- a/arnio/io.py
+++ b/arnio/io.py
@@ -347,10 +347,10 @@ def _validate_on_bad_lines(on_bad_lines: str) -> str:
 
 def _materialize_csv_input(
     source: str | os.PathLike[str] | io.TextIOBase,
-) -> tuple[str, bool]:
+) -> tuple[str, bool, bool]:
     """Convert supported CSV inputs into a filesystem path."""
     if isinstance(source, (str, os.PathLike)):
-        return os.fspath(source), False
+        return os.fspath(source), False, False
     if isinstance(source, io.StringIO) or (
         hasattr(source, "read") and callable(source.read)
     ):
@@ -372,7 +372,7 @@ def _materialize_csv_input(
                     )
                 tmp.write(chunk)
             tmp.close()
-            return tmp.name, True
+            return tmp.name, True, True
         except Exception:
             tmp.close()
             os.unlink(tmp.name)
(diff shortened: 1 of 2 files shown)


im-anishraj (approved):
Thanks @Exodus2004. I verified the focused text-stream encoding fix locally. python -m pytest tests/test_csv.py -q passes with 321 tests, and git diff --check origin/main...HEAD is clean. The only failing check shown is the external Vercel authorization gate, so this is safe to merge.

PR #1739, opened 2026-05-28
fix(pandas): add max_errors passthrough to ArnioPandasAccessor.validate

### LINKED ISSUE #1396: Feature: expose max_errors on pandas validation accessor
## Summary
The pandas accessor method `df.arnio.validate(schema)` does not expose the `max_errors` option supported by the core validation API.

## Why it matters
Users who work from pandas can already call `ar.validate(frame, schema, max_errors=...)` to cap validation output for large or very dirty datasets. The pandas accessor is meant to make Arnio easier to use alongside pandas, but this missing parameter forces users to drop back to the lower-level conversion path.

## Evidence
- `arnio/schema.py` supports `validate(frame, schema, max_errors=...)` and tests cover max-error behavior.
- `arnio/integrations/pandas.py::ArnioPandasAccessor.validate()` currently accepts only `schema` and calls `validate(self.to_arframe(), schema)`.
- `tests/test_pandas_accessor.py` does not cover a `max_errors` passthrough.

## Reproduction or trigger
```python
import pandas as pd
import arnio as ar

pdf = pd.DataFrame({"age": [-1, -2, -3]})
schema = ar.Schema({"age": ar.Int64(min=0)})

# Core API supports this:
result = ar.validate(ar.from_pandas(pdf), schema, max_errors=1)

# Pandas accessor currently has no equivalent:
pdf.arnio.validate(schema, max_errors=1)  # TypeError today
```

## Expected behavior
`ArnioPandasAccessor.validate()` should accept `max_errors: int | None = None` and pass it through to the core validation API.

## Actual behavior
The accessor signature does not include `max_errors`, so pandas-first users cannot use the error cap through `df.arnio.validate(...)`.

## Suggested direction
- Add `max_errors: int | None = None` to `ArnioPandasAccessor.validate()`.
- Forward it to `validate(self.to_arframe(), schema, max_errors=max_errors)`.
- Add a focused pandas accessor regression test showing the result respects the cap.
- Update docs/API reference if the accessor section documents method signatures.

## References
- `arnio/integrations/pandas.py::ArnioPandasAccessor.validate`
- `arnio/schema.py::validate`
- `tests/test_schema.py` max-error validation tests
- `tests/test_pandas_accessor.py`

ayushlad9108:
# Title

fix(pandas): add max_errors passthrough to ArnioPandasAccessor.validate

# Description

Fixes #1396

## What

`ArnioPandasAccessor.validate()` was missing the `max_errors` parameter already supported by the core `ar.validate()` API. Because of this, pandas-first users had to manually convert DataFrames just to limit validation output.

## Changes

* `arnio/integrations/pandas.py`

  * added `max_errors: int | None = None` as a keyword-only argument to `ArnioPandasAccessor.validate()`
  * forwarded the value to the core `validate()` call
  * updated the docstring

* `tests/test_pandas_accessor.py`

  * added `test_pandas_accessor_validate_respects_max_errors`
  * verifies that `max_errors=1` caps the output to one issue while the uncapped call still returns all three validation failures

## Before / After

```python
# Before — TypeError
pdf.arnio.validate(schema, max_errors=1)

# After — works, mirrors ar.validate()
pdf.arnio.validate(schema, max_errors=1)
```

## Scope

Two files only. No C++ changes and no schema validation logic changes.

The core `validate()` implementation already supported `max_errors`; this PR simply exposes the existing functionality through the pandas accessor layer.

### DIFF
--- a/arnio/integrations/pandas.py
+++ b/arnio/integrations/pandas.py
@@ -134,6 +134,21 @@ class ArnioPandasAccessor:
 
         return to_pandas(result)
 
-    def validate(self, schema: Schema | dict[str, Any]) -> ValidationResult:
-        """Validate the DataFrame against an Arnio schema."""
-        return validate(self.to_arframe(), schema)
+    def validate(
+        self,
+        schema: Schema | dict[str, Any],
+        *,
+        max_errors: int | None = None,
+    ) -> ValidationResult:
+        """Validate the DataFrame against an Arnio schema.
+
+        Parameters
+        ----------
+        schema : Schema or dict[str, Field]
+            Schema to validate against.
+        max_errors : int or None, default None
+            Maximum number of validation issues to collect. Mirrors the
+            ``max_errors`` parameter of ``ar.validate()``. When None all
+            issues are collected.
+        """
+        return validate(self.to_arframe(), schema, max_errors=max_errors)
--- a/tests/test_pandas_accessor.py
+++ b/tests/test_pandas_accessor.py
@@ -160,3 +160,18 @@ def test_pandas_accessor_auto_clean_return_report_and_explain():
     assert list(result["name"]) == ["Alice", "Bob"]
     assert isinstance(report, ar.DataQualityReport)
     assert isinstance(explanation, ar.CleanExplanation)
+
+
+def test_pandas_accessor_validate_respects_max_errors():
+    # max_errors=1 should cap the result at one issue even when multiple rows fail.
+    df = pd.DataFrame({"age": [-1, -2, -3]})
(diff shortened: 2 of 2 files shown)


ayushlad9108:
@im-anishraj Done with the changes! Now there are no merge conflicts!

im-anishraj (approved):
Thanks @ayushlad9108. This is a clean replacement for the older accessor PR. I verified it locally with python -m pytest tests/test_pandas_accessor.py -q (14 passed), git diff --check origin/main...HEAD is clean, and the GitHub Actions matrix is green. The only failing check is Vercel authorization. Safe to merge.

PR #1743, opened 2026-05-28
fix: preserve deferred schema inference for all-null chunked columns

### LINKED ISSUE #1731: Bug: read_csv_chunked silently corrupts column dtype when first chunk is all-null
### What happened?

When reading a CSV with `read_csv_chunked()`, if a column contains only null
values in the first chunk, its schema is permanently locked to `string`
before later chunks are processed.

If integers or floats appear in subsequent chunks, they are silently stored as
strings instead of numeric types. No warning or exception is raised, values
appear visually correct when printed, but numeric operations either fail or
produce incorrect results.

Root cause: in `CsvChunkReader::next_chunk()` (`cpp/src/csv_reader.cpp`),
after inferring types from chunk 1, any column still at `NULL_TYPE` is
immediately promoted to `STRING` and `schema_locked_` is set to `true`.

This happens before chunk 2 is processed, so later inference paths such as:

```cpp
promote_type(NULL_TYPE, INT64) = INT64
```

never execute for those columns, even when subsequent chunks contain valid
numeric data.

### Area

### Minimal reproduction

```python
import tempfile, os
import arnio as ar
import pandas as pd

csv = (
    "order_id,customer,revenue\n"
    "1001,alice,\n"       # chunk 1 — revenue is null
    "1002,bob,\n"         # chunk 1 — revenue is null
    "1003,carol,4500\n"   # chunk 2 — revenue is integer
    "1004,dave,7200\n"    # chunk 2 — revenue is integer
)

with tempfile.NamedTemporaryFile(mode="w", suffix=".csv", delete=False) as f:
    f.write(csv)
    path = f.name

try:
    chunks = list(ar.read_csv_chunked(path, chunksize=2))
    df = pd.concat([ar.to_pandas(c) for c in chunks], ignore_index=True)

    print("chunk 2 dtype:", chunks[1].dtypes["revenue"])
    # prints 'string', expected 'int64'

    print("sum:", df["revenue"].sum())
    # returns '45007200' instead of 11700 due to string concatenation

finally:
    os.unlink(path)
```

### Expected behavior

```text
chunks[1].dtypes["revenue"] == "int64"
df["revenue"].sum() == 11700
````

The schema for a column should not be locked until at least one non-null value
has been observed.

A column that is entirely null in chunk 1 should remain at `NULL_TYPE`
(meaning "no opinion yet") so that chunk 2 can correctly infer the real type via:

```cpp id="mwblyy"
promote_type(NULL_TYPE, INT64) = INT64
```

This is consistent with how `read_csv()` handles the same file in a single pass.

### Actual behavior

```text
chunk 2: {'revenue': 'string', 'customer': 'string', 'order_id': 'int64'}
````

`df["revenue"].sum()` returns:

```text
'45007200'
```

instead of:

```text
11700
```

because values `4500` and `7200` are stored as strings (`'4500'`, `'7200'`)
rather than numeric values.

This results in silent string concatenation instead of numeric aggregation.

No warning, exception, or schema mismatch error is raised.

### Impact

Produces incorrect data

### Arnio version

1.18.0

### Install method

pip install -e ".[dev]" --break-system-packages

### Operating System

Linux

### Python version

3.13.11

### Environment details

```shell
┌──(kali㉿kali)-[~/arnio]
└─$ pip show arnio
Name: arnio
Version: 1.18.0
Summary: C++ accelerated data preparation for pandas and the Python data stack
Home-page: https://github.com/im-anishraj/arnio
Author: 
Author-email: Anish Raj <anishrajyadav97@gmail.com>
License: MIT
Location: /home/kali/.local/lib/python3.13/site-packages
Editable project location: /home/kali/arnio
Requires: numpy, pandas
Required-by:
```

### Pre-flight checks

### Additional context

Validated with a dedicated proof-of-bug script before and after patching.

### Before fix

```text
chunk 2 dtype → string
row value → '4500'
sum → '45007200'
````

Image

### After fix

```text
chunk 2 dtype → int64
row value → 4500
sum → 11700
```

Image

Confirmed against v1.18.0 with the attached before/after reproducer.

After fix:

* all checks pass
* `sum == 11700`
* chunk 2 dtype resolves to `int64` (see image 1)

Before fix:

* dtype locked to `string`
* sum returns `'45007200'` due to string concatenation (see image 2)

### Root cause

`CsvChunkReader::next_chunk()` in `cpp/src/csv_reader.cpp` promotes any
remaining `NULL_TYPE` column to `STRING` and sets `schema_locked_ = true`
at the end of chunk 1, before chunk 2 has been read.

### Proposed fix

Defer `schema_locked_ = true` until every selected column has a concrete type.

`NULL_TYPE` columns (all-null so far) should remain unlocked so subsequent
chunks can infer the real type via:

```cpp
promote_type(NULL_TYPE, INT64) = INT64
```

A PR with the fix and regression tests is ready.

dynamo-pentester:
Closes #1731 

## Problem

`read_csv_chunked()` silently corrupts column dtypes when a column's first
chunk contains only null values.

In `CsvChunkReader::next_chunk()`, after processing chunk 1, any column
still at `NULL_TYPE` was immediately promoted to `STRING` and
`schema_locked_ = true` was set before chunk 2 was ever processed.

This meant inference paths such as:

```cpp
promote_type(NULL_TYPE, INT64) = INT64
```

never had a chance to execute for those columns. Integers and floats in
later chunks were silently stored as strings with no warning or error.

### Before fix

```python
chunks = list(ar.read_csv_chunked("data.csv", chunksize=2))

# column 'revenue' is all-null in chunk 1,
# then contains integers in chunk 2

chunks[1].dtypes["revenue"]
# → 'string'  ✗

df["revenue"].sum()
# → '45007200'  ✗ (string concatenation)
```

---

#### `next_chunk()`

Replaced the unconditional `NULL_TYPE → STRING` promotion and immediate
schema lock with a conditional lock.

The schema is now locked only after every selected column has resolved to
a concrete type.

Columns still at `NULL_TYPE` (meaning "all-null so far") remain unlocked
so later chunks can continue type inference correctly.

#### `build_frame()`

Introduced a temporary `effective_type` fallback.

If a column is still `NULL_TYPE` at emit time, the temporary emitted
`Column` is constructed as `STRING` so the current chunk can still be
materialized safely.

This is safe because the emitted values are null-only regardless of
storage type.

Crucially, `col_types_[ci]` itself is **not** mutated here, so the next
chunk still begins inference from `NULL_TYPE` and can correctly promote
later numeric values.

### After fix

```python
chunks[1].dtypes["revenue"]
# → 'int64'  ✓

df["revenue"].sum()
# → 11700  ✓
```

---

## Tests

Added six regression tests to
`tests/test_csv_chunked.py`
under:

```python
TestCsvChunkedNullColumnSchemaInference
```

| Test | Coverage |
|---|---|
| `test_integer_column_all_null_in_first_chunk` | Core regression for front-loaded null integer columns |
| `test_float_column_all_null_in_first_chunk` | Same scenario for float64 |
| `test_null_first_chunk_values_are_null_not_string` | Ensures null rows remain `pd.NA`, not empty strings |
| `test_schema_consistent_across_all_chunks` | Verifies dtype consistency once resolved |
| `test_genuinely_all_null_column_becomes_string` | Confirms EOF fallback remains correct |
| `test_full_dataframe_matches_read_csv_with_null_first_chunk` | Verifies values + resolved dtypes match `read_csv()` |

All existing tests pass after the change:

```text
2150 passed, 45 skipped
```

---

## Notes

- Only affects `read_csv_chunked()`
- Does **not** affect `read_csv()` or `scan_csv()`
- Columns that remain null across the entire file still correctly resolve
  to `string` at EOF
- `promote_type(NULL_TYPE, X) = X` was already correct — the bug was
  that inference never reached those paths for front-loaded null columns


im-anishraj (approved):
Thanks @dynamo-pentester. I rebuilt the native extension locally with Visual Studio, then verified the branch with python -m pytest tests/test_csv_chunked.py -q (24 passed). git diff --check origin/main...HEAD is clean. The earlier lint red was caused by main formatting from a just-merged CSV encoding PR, which I fixed on main separately. This focused chunked inference fix is safe to merge.

dynamo-pentester:
Thanks @im-anishraj — really appreciate the thorough verification and quick review 🙌

PR #1749, opened 2026-05-28
fix(io): add skiprows alias to read_csv_chunked for API consistency

### LINKED ISSUE #1293: Bug: read_csv and read_csv_chunked use inconsistent skip-row parameter names

ayushlad9108:
Fixes #1293

## What
read_csv() uses `skiprows` (matching pandas convention) but read_csv_chunked()
only accepted `skip_rows`, causing a confusing TypeError when users switched
between the two functions.

## Changes
- `arnio/io.py`: added `skiprows: int | None = None` as a keyword-only alias
  in `read_csv_chunked()`. `skip_rows` is fully preserved for backward compat.
- `tests/test_csv.py`: added `TestReadCsvChunkedSkiprowsAlias` with 8 tests.

## Conflict resolution
| Inputs | Behaviour |
|---|---|
| Only `skiprows` | used as the row-skip count |
| Only `skip_rows` | existing behaviour unchanged |
| Both, same value | no error |
| Both, different values | `ValueError` with clear message |

## Scope
Two files only. No C++ changes. `skip_rows` semantics are unchanged —
both parameters skip data rows *after* the header in chunked mode.

### DIFF
--- a/arnio/io.py
+++ b/arnio/io.py
@@ -596,7 +596,10 @@ def read_csv(
     try:
         effective_encoding = "utf-8" if is_materialized_text else encoding
         with _utf8_csv_path(
-            path, effective_encoding, encoding_errors=encoding_errors, delimiter=delimiter
+            path,
+            effective_encoding,
+            encoding_errors=encoding_errors,
+            delimiter=delimiter,
         ) as native_path:
             cpp_frame, bad_rows = reader.read(native_path, on_bad_lines)
 
@@ -627,6 +630,7 @@ def read_csv_chunked(
     usecols: list[str] | None = None,
     nrows: int | None = None,
     skip_rows: int = 0,
+    skiprows: int | None = None,
     encoding: str = "utf-8",
     trim_headers: bool = True,
     decimal_separator: str = ".",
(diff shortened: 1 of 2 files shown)


im-anishraj (approved):
@ayushlad9108 thank you, this is a clean API consistency fix. I verified the new `skiprows` alias behavior locally with the focused alias tests and the full CSV test file, and the existing `skip_rows` behavior remains backward-compatible.

Local verification:
- `python -m pytest tests/test_csv.py::TestReadCsvChunkedSkiprowsAlias -q` -> 8 passed
- `python -m pytest tests/test_csv.py -q` -> 329 passed
- `python -m black --check arnio/io.py tests/test_csv.py` -> passed

GitHub Actions are green for the native tests, lint, examples, notebook smoke, benchmark, pyarrow jobs, and the OS/Python matrix except the recurring Vercel authorization check and one macOS 3.13 runner that is queued rather than failing. I am approving this for merge.

PR #1757, opened 2026-05-28
fix: make drop_columns empty list behavior return a new frame (#1654)

### LINKED ISSUE #1654: Bug: make drop_columns empty-list behavior return a new frame
## Summary
The public helper `ar.drop_columns(frame, [])` returns the original `ArFrame` object unchanged, while the method-level `frame.drop_columns([])` returns a new frame copy. This makes the helper inconsistent with its own docs and with the method API.

## Why it matters
Arnio cleaning helpers generally behave like functional transformations: they return a result frame rather than the same object. Returning the same object for an empty drop list can surprise callers and can leak metadata mutations through shared identity.

## Evidence
- `arnio/cleaning.py`: `drop_columns()` has `if len(requested_columns) == 0: return frame`.
- `arnio/frame.py`: `ArFrame.drop_columns()` handles empty input with `return ArFrame(self._frame.select_columns(self.columns))`, producing a new frame.
- The helper docstring says it returns a `New frame with the requested columns removed`.

## Reproduction or trigger
```python
import pandas as pd
import arnio as ar

frame = ar.from_pandas(pd.DataFrame({"a": [1]}))
result = ar.drop_columns(frame, [])

print(result is frame)  # True
```

## Expected behavior
The helper should match the method API and return a new no-op frame for an empty column list, or the docs should explicitly document identity preservation as intentional.

## Actual behavior
The helper returns the original `ArFrame` object for empty input.

## Suggested direction
Change the empty-list path to return a copy/new frame, likely by delegating to `frame.drop_columns([])` or selecting all columns through the existing frame method.

Add a regression test that asserts:

- `ar.drop_columns(frame, [])` preserves data and column order
- the returned object is not the same object as the input
- method and helper behavior are consistent

## References
- `arnio/cleaning.py`
- `arnio/frame.py`
- `tests/test_cleaning.py`
- `tests/test_frame.py`

Exodus2004:
Resolves #1654

### What was done:
- Updated the public helper \r.drop_columns(frame, [])\ to delegate to the method-level \rame.drop_columns([])\ when the list of columns to drop is empty.
- This ensures consistent 'new frame' copy behavior and satisfies the docstring statement that it returns a new frame.
- Updated the test suite in \	ests/test_cleaning.py\ to verify:
  - The returned frame is not the same object as the input (\
esult_helper is not frame\).
  - The returned frame preserves data and matches the input.
  - Method and helper behaviors are consistent.

Ready for review!

### DIFF
--- a/arnio/cleaning.py
+++ b/arnio/cleaning.py
@@ -247,7 +247,7 @@ def drop_columns(frame: ArFrame, columns: Sequence[str]) -> ArFrame:
         ),
     )
     if len(requested_columns) == 0:
-        return frame
+        return frame.drop_columns([])
     if len(requested_columns) == len(frame.columns):
         raise ValueError("drop_columns cannot remove all columns from the frame")
 
--- a/tests/test_cleaning.py
+++ b/tests/test_cleaning.py
@@ -683,9 +683,12 @@ class TestDropColumns:
     def test_drop_columns_allows_empty_input_as_no_op(self, sample_csv):
         frame = ar.read_csv(sample_csv)
 
-        result = ar.drop_columns(frame, [])
+        result_helper = ar.drop_columns(frame, [])
+        result_method = frame.drop_columns([])
 
-        assert result is frame
+        assert result_helper is not frame
+        assert result_helper == frame
+        assert result_helper == result_method
 
     def test_drop_columns_rejects_missing_columns(self, sample_csv):
         frame = ar.read_csv(sample_csv)
--- a/tests/test_pipeline.py
+++ b/tests/test_pipeline.py
@@ -402,7 +402,9 @@ class TestPipeline:
         frame = ar.read_csv(sample_csv)
         result = ar.pipeline(frame, [("drop_columns", {"columns": []})])
 
-        assert result is frame
+        assert result is not frame
+        assert result.columns == frame.columns
+        assert len(result) == len(frame)
 
     def test_pipeline_drop_columns_rejects_missing_columns(self, sample_csv):
         import pytest


im-anishraj (changes requested):
@Exodus2004 thanks for the focused fix. The helper behavior change itself is sensible and the cleaning tests you updated pass locally, but this PR is not ready to merge yet because it leaves the pipeline expectation inconsistent.

Verified locally:
- `python -m pytest tests/test_cleaning.py::TestDropColumns -q` -> 6 passed
- `python -m pytest tests/test_cleaning.py -q` -> 389 passed
- `python -m black --check arnio/cleaning.py tests/test_cleaning.py` -> passed
- `python -m pytest tests/test_pipeline.py::TestPipeline::test_pipeline_drop_columns_allows_empty_columns -q` -> fails

GitHub Actions is failing for the same reason:
`tests/test_pipeline.py::TestPipeline::test_pipeline_drop_columns_allows_empty_columns` still asserts `result is frame`, but after this fix the pipeline also returns a new equivalent frame.

Please update the pipeline regression test to match the new contract, for example asserting `result is not frame` plus equality/data preservation, and then rerun the suite. After that this should be mergeable.

Exodus2004:
I have updated the pipeline regression test to assert that it returns a new frame instead of the exact same object reference, preserving all data and columns. This aligns the test suite with the new contract. Thanks for catching that!

im-anishraj (approved):
@Exodus2004 thanks for updating the pipeline expectation. I verified the current merge ref locally:

- `black --check arnio/cleaning.py tests/test_cleaning.py tests/test_pipeline.py` passed
- `ruff check arnio/cleaning.py tests/test_cleaning.py tests/test_pipeline.py` passed
- `pytest tests/test_cleaning.py::TestDropColumns tests/test_pipeline.py::TestPipeline::test_pipeline_drop_columns_allows_empty_columns -q` -> 7 passed

This now consistently returns a new equivalent frame for the empty-list no-op path across helper, method, and pipeline usage.

PR #1758, opened 2026-05-28
ci: smoke test sdist install in release dry run

### LINKED ISSUE #1643: CI: smoke-test sdist installs in release dry runs
## Summary
The release dry-run workflow builds and checks the source distribution with Twine, but it does not install the sdist and smoke-test `import arnio` the way the real publish workflow does.

## Why it matters
Arnio ships a native extension, so an sdist can pass `twine check` while still failing at install/import time because of build-backend, native-source, or packaging configuration problems. The dry-run workflow should catch release-blocking sdist import failures before the actual publish path runs.

## Evidence
- `.github/workflows/release-dry-run.yml`: `build_sdist` runs `pipx run build --sdist` and `twine check dist/*.tar.gz` only.
- `.github/workflows/build_wheels.yml`: `build_sdist` installs `dist/*.tar.gz` and runs an isolated `python -c "import arnio"` smoke test.
- The dry-run summary reports `sdist build + twine` as safe, but that is weaker than the real publish workflow's sdist install/import gate.

## Reproduction or trigger
1. Run the `Release Dry-Run Checklist` workflow.
2. Inspect the `build_sdist` job.
3. Notice that it never installs the generated sdist or imports Arnio from an isolated working directory.

## Expected behavior
The dry-run release workflow should mirror the real publish gate closely enough to catch sdist install/import failures.

## Actual behavior
The dry-run workflow only validates sdist metadata with Twine.

## Suggested direction
Add an sdist install/import smoke step to `.github/workflows/release-dry-run.yml`, matching the isolated import check already used in `.github/workflows/build_wheels.yml`.

## References
- `.github/workflows/release-dry-run.yml`
- `.github/workflows/build_wheels.yml`

eshaanag:
## Summary

Fixes #1643

This PR strengthens the release dry-run sdist check so it mirrors the publish workflow more closely. The dry-run job already built the sdist and ran `twine check`; it now also installs the generated `dist/*.tar.gz` and verifies `import arnio` from a temporary directory with `PYTHONPATH` cleared.

This is for GSSoC 2026 and stays scoped to the assigned release dry-run workflow issue.

## What changed

- Added `python -m pip install dist/*.tar.gz` after the Twine metadata check.
- Added an isolated import smoke test using a temporary working directory.
- Updated the dry-run summary row from `sdist build + twine` to `sdist build + install smoke` so the reported gate matches the actual workflow.

## Workflow command path covered

The release dry-run sdist job now exercises:

```bash
pipx run build --sdist
pip install twine
twine check dist/*.tar.gz
python -m pip install dist/*.tar.gz
python -c "import arnio"  # from outside the checkout, with PYTHONPATH cleared
```

## Validation

- `git diff --check`
- YAML syntax load check with Ruby/Psych for `.github/workflows/release-dry-run.yml`

### DIFF
--- a/.github/workflows/release-dry-run.yml
+++ b/.github/workflows/release-dry-run.yml
@@ -45,6 +45,26 @@ jobs:
           pip install twine
           twine check dist/*.tar.gz && echo "✅ Twine check passed" || { echo "❌ Twine check failed"; exit 1; }
 
+      - name: Install package from sdist
+        run: python -m pip install dist/*.tar.gz
+
+      - name: Smoke test sdist import
+        run: |
+          python - <<'PY'
+          import os
+          import subprocess
+          import sys
+          import tempfile
+
+          with tempfile.TemporaryDirectory() as tmp:
+              subprocess.run(
+                  [sys.executable, "-c", "import arnio"],
+                  cwd=tmp,
+                  check=True,
+                  env={**os.environ, "PYTHONPATH": ""},
+              )
+          PY
+
       - uses: actions/upload-artifact@v6
         with:
           name: sdist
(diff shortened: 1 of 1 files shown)


im-anishraj (approved):
@eshaanag thanks, this is a well-scoped CI hardening PR. I checked it against issue #1643 and the existing publish workflow: the new dry-run sdist install/import smoke test mirrors the current `build_wheels.yml` sdist gate closely.

Verification performed:
- `git diff --check origin/main...HEAD` -> passed
- parsed `.github/workflows/release-dry-run.yml` with PyYAML -> passed
- compared the new install/import smoke path with `.github/workflows/build_wheels.yml`

GitHub only reports the recurring Vercel authorization failure for this PR; there are no repo CI failures attached to the workflow-only change. Approving for merge.

PR #1762, opened 2026-05-28
fix: quote date/datetime-like strings in schema_to_yaml to prevent YAML timestamp coercion

### LINKED ISSUE #1730: Bug: quote date-like strings in schema_to_yaml
## Summary
`schema_to_yaml()` emits date-looking string values without quotes, so YAML parsers can interpret them as timestamps instead of strings.

## Why it matters
Schema export should preserve data contract values exactly. If an allowed/default string such as `"2026-05-28"` is emitted unquoted, downstream YAML tooling may load it as a date object instead of a string, changing the schema semantics during round-trip or review.

## Evidence
- `arnio/schema_export.py`: `_emit_scalar()` quotes numeric-looking strings, booleans, null-like strings, strings with colons, and several special prefixes, but it does not quote ISO date-like strings.
- Runtime probe on current main:
  ```python
  ar.schema_to_yaml({"field": {"type": "string", "allowed": ["2026-05-28"]}})
  ```
  produces:
  ```yaml
  fields:
    field:
      allowed:
        - 2026-05-28
      type: string
  ```

## Reproduction or trigger
1. Export a schema containing a string value that looks like `YYYY-MM-DD`.
2. Load the YAML with a YAML parser that follows timestamp resolution rules.
3. Observe that the scalar may become a date/timestamp instead of a string.

## Expected behavior
Date-looking string values should be quoted so they stay strings after YAML round-trip.

## Actual behavior
`schema_to_yaml()` emits `YYYY-MM-DD` strings as plain scalars.

## Suggested direction
Extend `_emit_scalar()` to quote date-like strings, at minimum values matching `^\d{4}-\d{2}-\d{2}$`. Add regression tests for allowed/default string values that look like dates.

## References
- `arnio/schema_export.py`

dynamo-pentester:
Closes #1730

## Problem

`_emit_scalar()` in `arnio/schema_export.py` quoted numeric strings,
booleans, null-like strings, and strings with special characters — but
not ISO date/datetime patterns. YAML 1.1 parsers silently resolve bare
scalars like `2026-05-28` to timestamp objects, breaking the schema
data contract on round-trip.

```python
# Before fix
ar.schema_to_yaml(
    {"field": {"type": "string", "allowed": ["2026-05-28"]}}
)

# emits:
#   allowed:
#     - 2026-05-28
#
# YAML parsers read this as a timestamp object,
# not a string.
```

## Fix — `arnio/schema_export.py`

Added `import re` and a module-level compiled pattern:

```python
_DATE_RE = re.compile(
    r"^\d{4}-\d{2}-\d{2}([ T]\d{2}:\d{2}(:\d{2})?.*|Z)?$"
)
```

Added one condition to the `needs_quoting` block in `_emit_scalar()`:

```python
or bool(_DATE_RE.match(value))
```

Patterns covered:

| Pattern | Example | YAML 1.1 risk |
|---|---|---|
| `YYYY-MM-DD` | `2026-05-28` | timestamp |
| `YYYY-MM-DDZ` | `2026-05-28Z` | timestamp |
| `YYYY-MM-DDThh:mm` | `2026-05-28T10:30` | timestamp |
| `YYYY-MM-DDThh:mm:ss` | `2026-05-28T10:30:00` | timestamp |
| `YYYY-MM-DDThh:mm:ss+tz` | `2026-05-28T10:30:00+05:30` | timestamp |
| `YYYY-MM-DD hh:mm:ss` | `2026-05-28 10:30:00` | timestamp |

```python
# After fix
ar.schema_to_yaml(
    {"field": {"type": "string", "allowed": ["2026-05-28"]}}
)

# emits:
#   allowed:
#     - "2026-05-28"
#
# YAML parsers now preserve the value as a string.
```

## Tests — `tests/test_schema_export.py`

13 regression tests added under `TestDateLikeStringQuoting`:

| Test | Covers |
|---|---|
| `test_bare_date_in_allowed_is_quoted` | Core case — `allowed` list |
| `test_bare_date_as_default_is_quoted` | `default` field |
| `test_multiple_date_allowed_values_all_quoted` | All values in a list quoted |
| `test_iso_datetime_T_in_allowed_is_quoted` | `T`-separated datetime |
| `test_bare_date_with_z_suffix_is_quoted` | `2026-05-28Z` UTC suffix |
| `test_iso_datetime_T_no_seconds_is_quoted` | `hh:mm` without seconds |
| `test_space_separated_datetime_is_quoted` | Space-separated datetime |
| `test_round_trip_date_stays_string` | PyYAML round-trip — value remains `str` |
| `test_round_trip_datetime_stays_string` | Same for datetime |
| `test_year_only_string_not_quoted_as_date` | `"2026"` — not a false positive |
| `test_partial_date_mm_dd_not_treated_as_date` | `"05-28"` — not a false positive |
| `test_normal_string_unaffected` | Plain strings remain unchanged |
| `test_existing_numeric_quoting_still_works` | Pre-existing behaviour unchanged |


im-anishraj (changes requested):
@dynamo-pentester thanks for the detailed write-up. The date/datetime YAML quoting issue is valid, but I cannot merge this PR in its current form because the branch contains unrelated changes outside the stated fix.

The current diff includes:
- C++ CSV parser performance changes in `cpp/include/arnio/csv_reader.h`, `cpp/src/csv_reader.cpp`, and `cpp/tests/test_csv_parser.cpp`
- a notebook formatting change in `examples/ignite_arnio_cleaning.ipynb`
- an older `schema_to_dict()` normalization change in `arnio/schema_export.py`
- the actual YAML date/datetime quoting fix and tests

Please rebase from latest `main` and keep this PR scoped only to issue #1730. Ideally the final diff should only touch:
- `arnio/schema_export.py`

Once the unrelated commits are removed, I can review the YAML quoting behavior directly and run the schema export tests.

PR #1771, opened 2026-05-28
fix: gate Catch2 behind native test option

### LINKED ISSUE #1715: Build: wheel builds fetch Catch2 from GitHub even when tests are not being built
I noticed this while looking at the native build path for GSSoC 2026. The C++ test dependency is wired into `cpp/CMakeLists.txt` unconditionally, so a normal package build can end up doing a network fetch for Catch2 even though Catch2 is only needed for native tests.

This is a classic build-isolation footgun: the runtime extension should not depend on being able to reach GitHub for a test framework.

Relevant lines:

- `cpp/CMakeLists.txt:1` includes `FetchContent`
- `cpp/CMakeLists.txt:3-8` declares and downloads Catch2 from `https://github.com/catchorg/Catch2.git`
- `CMakeLists.txt:14` always runs `add_subdirectory(cpp)` during the main build

What breaks:

```bash
python -m build --wheel
# or any scikit-build-core wheel/sdist build in a restricted CI environment
```

If the build runner has no network access, GitHub is unavailable, or the environment intentionally blocks arbitrary dependency downloads, the package build can fail before it even reaches the Arnio extension build. That is surprising because Catch2 is a test-only dependency and should not be required for installing or building the runtime package.

Expected behavior: normal wheel/sdist builds should build `arnio_core` and the Python extension without downloading Catch2. Native C++ tests can opt into Catch2 separately.

Current behavior: Catch2 is fetched as soon as `cpp/CMakeLists.txt` is configured, before there is any test-build guard.

A practical fix would be to gate the Catch2 setup and `arnio_test(...)` targets behind a CMake option, probably something like:

```cmake
option(ARNIO_BUILD_CPP_TESTS "Build Arnio C++ tests" OFF)

if(ARNIO_BUILD_CPP_TESTS)
  include(FetchContent)
  FetchContent_Declare(
      Catch2
      GIT_REPOSITORY https://github.com/catchorg/Catch2.git
      GIT_TAG        v3.5.3
  )
  FetchContent_MakeAvailable(Catch2)
  enable_testing()
  # arnio_test(...) targets here
endif()
```

For CI jobs that run native C++ tests, the workflow can pass `-DARNIO_BUILD_CPP_TESTS=ON`. For package builds, the default should stay offline-friendly and runtime-only.

Requested labels: `bug`, `type:devops`, `gssoc`, `gssoc:level-3`, `level:advanced`, `difficulty:advanced`, `area:ci-packaging`, `backend`, `priority:medium`, `size:m`.

saurabhhhcodes:
## Summary
- add ARNIO_BUILD_CPP_TESTS with a default OFF value so wheel/default native builds do not fetch Catch2
- keep Catch2 FetchContent and C++ test targets behind that explicit opt-in
- update the native C++ CI job to enable ARNIO_BUILD_CPP_TESTS when tests are meant to run

Closes #1715

## Verification
- python -m build --wheel
- default CMake configure/build without Catch2 FetchContent dirs
- cmake -DARNIO_BUILD_CPP_TESTS=ON opt-in native build
- ctest --test-dir build-cpp-tests --output-on-failure (4/4 passed)
- git diff --check origin/main...HEAD

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -58,6 +58,7 @@ jobs:
         run: |
           cmake -S . -B build \
             -DCMAKE_BUILD_TYPE=Debug \
+            -DARNIO_BUILD_CPP_TESTS=ON \
             -DCMAKE_PREFIX_PATH=$(python3 -c "import pybind11; print(pybind11.get_cmake_dir())")
       - name: Build
         run: cmake --build build --parallel
--- a/cpp/CMakeLists.txt
+++ b/cpp/CMakeLists.txt
@@ -1,12 +1,3 @@
-include(FetchContent)
-
-FetchContent_Declare(
-    Catch2
-    GIT_REPOSITORY https://github.com/catchorg/Catch2.git
-    GIT_TAG        v3.5.3
-)
-FetchContent_MakeAvailable(Catch2)
-
 add_library(arnio_core STATIC
     src/column.cpp
     src/frame.cpp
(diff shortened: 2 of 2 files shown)


saurabhhhcodes:
Validation update: the full GitHub matrix is green now. Lint, examples, benchmark, manylinux/windows/macOS wheel smoke tests, C++ native tests, notebook smoke, and Python 3.9-3.14 on Ubuntu/Windows/macOS all passed. The only remaining red status is Vercel project authorization, which is maintainer-side. Ready for review/merge.

im-anishraj (approved):
@saurabhhhcodes thanks, this is a solid CI/build fix. I verified both paths locally and checked the GitHub matrix.

Local verification on Windows/MSVC:
- `git diff --check origin/main...HEAD` -> passed
- default CMake configure/build with `ARNIO_BUILD_CPP_TESTS=OFF` -> passed and produced no `_deps` Catch2 directory
- opt-in CMake configure/build with `-DARNIO_BUILD_CPP_TESTS=ON` -> passed
- `ctest` for the opt-in native build -> 4/4 C++ tests passed

GitHub Actions are green for lint, native C++ tests, wheel smoke tests, examples, benchmark, and the OS/Python matrix. Vercel is the recurring authorization-only failure. Approving for merge.

PR #1794, opened 2026-05-28
fix: reject non-null push_back into NULL_TYPE columns

### LINKED ISSUE #1595: Bug: Column::push_back allows NULL_TYPE variant state corruption and null-mask desynchronization
### What happened?

While auditing the native C++ column internals, I discovered that `Column::push_back` allows non-null values to be inserted into `NULL_TYPE` columns without promoting or allocating the underlying variant storage.

The null mask grows correctly, but the internal variant storage remains `std::monostate`.

This creates a corrupted internal state where:
- the column reports a valid non-null value
- the backing vector does not exist
- null-mask state becomes desynchronized from variant storage

As a result, subsequent access operations may:
- return invalid values
- expose corrupted state
- throw runtime exceptions
- or trigger crashes during downstream operations

### Area

C++ cleaning engine

### Minimal reproduction

```cpp
#include <arnio/column.hpp>

arnio::Column col("test", arnio::DType::NULL_TYPE);

col.push_back(int64_t{10});

// Internal state now becomes inconsistent
// null_mask_ contains a valid value
// but underlying variant storage is still std::monostate
```

### Expected behavior

`NULL_TYPE` columns should either:
- reject non-null insertions
- or promote storage safely before insertion

The internal variant storage and null-mask state should always remain synchronized.

### Actual behavior

`Column::push_back` appends to the null mask even when the backing variant storage is still `std::monostate`.

This creates an invalid internal state where:
- null-mask length increases
- no backing vector exists
- non-null access paths become inconsistent

Subsequent operations may expose corrupted behavior or runtime failures.

### Impact

Produces incorrect data

### Arnio version

1.1.0

### Install method

Built from source

### Operating System

Linux

### Python version

3.12.4

### Environment details

```shell
Observed during deep validation of the native C++ column storage layer.

Affected path:
- `cpp/src/column.cpp`
- `Column::push_back`

The current implementation allows:
- `null_mask_` growth
- without corresponding variant-vector allocation

when the column dtype remains `NULL_TYPE`.

This creates an out-of-sync state between:
- variant storage
- null tracking
- logical column validity
```

### Pre-flight checks

### Additional context

This appears to be a core structural integrity issue in the native column engine.

Potential safe fix direction:

Validate insertions into `NULL_TYPE` columns before mutating the null mask.

Example direction:

```cpp
if (dtype_ == DType::NULL_TYPE && !is_null_val) {
    throw std::invalid_argument(
        "Cannot push non-null values into NULL_TYPE column"
    );
}
```

This would:
- preserve variant consistency
- prevent null-mask desynchronization
- protect downstream access logic
- maintain internal storage invariants

abhinavuser:
fixes #1595 
`Column::push_back` was mutating `null_mask_` before validating the column dtype, so non-null values would grow the mask while the backing variant storage stayed `std::monostate` -- leaving null tracking and data storage permanently out of sync

### root cause
the guard-before-mutate ordering was missing. `null_mask_.push_back(is_null_val)` ran unconditionally on line 174 before the `PushbackVisitor` dispatch, and since `PushbackVisitor::operator()(std::monostate)` is a no-op for monostate storage, non-null values just silently corrupted the column

### fix
added a dtype check in `push_back` that throws `std::invalid_argument` **before** any internal state is touched when a non-null value targets a NULL_TYPE column. null insertions (`std::monostate`) still work fine, consistent with `push_null` behavior

### changes
- `cpp/src/column.cpp` - NULL_TYPE validation guard before `null_mask_.push_back`
- - `cpp/tests/test_column.cpp` - 3 regression tests
### test coverage
- non-null int64/string/double/bool into NULL_TYPE -> all throw `std::invalid_argument`
- - column size stays 0 after rejected push (no state mutation)
- - dtype and variant storage remain consistent after rejection
- - `push_back(std::monostate{})` and `push_null()` still work on NULL_TYPE

### DIFF
--- a/cpp/src/column.cpp
+++ b/cpp/src/column.cpp
@@ -171,6 +171,16 @@ struct PushbackVisitor {
 void Column::push_back(const CellValue& value) {
     assert_type_consistency();
     bool is_null_val = std::holds_alternative<std::monostate>(value);
+
+    // Guard: reject non-null insertions into NULL_TYPE columns before
+    // mutating any internal state.  NULL_TYPE storage is std::monostate
+    // and has no backing vector, so allowing a non-null value would grow
+    // null_mask_ without a corresponding data element, leaving the
+    // variant storage and null tracking permanently out of sync.
+    if (dtype_ == DType::NULL_TYPE && !is_null_val) {
+        throw std::invalid_argument("Cannot push non-null value into a NULL_TYPE column");
+    }
+
     null_mask_.push_back(is_null_val);
 
     PushbackVisitor pv{data_};
--- a/cpp/tests/test_column.cpp
+++ b/cpp/tests/test_column.cpp
@@ -48,6 +48,51 @@ TEST_CASE("Column NULL_TYPE has zero size", "[column]") {
     REQUIRE(col.size() == 0);
 }
 
+TEST_CASE("NULL_TYPE column rejects non-null push_back", "[column]") {
+    Column col("test", DType::NULL_TYPE);
+
+    SECTION("int64 value is rejected") {
+        REQUIRE_THROWS_AS(col.push_back(int64_t{10}), std::invalid_argument);
+    }
+    SECTION("string value is rejected") {
+        REQUIRE_THROWS_AS(col.push_back(std::string("hi")), std::invalid_argument);
+    }
+    SECTION("double value is rejected") {
+        REQUIRE_THROWS_AS(col.push_back(double{3.14}), std::invalid_argument);
+    }
+    SECTION("bool value is rejected") {
+        REQUIRE_THROWS_AS(col.push_back(bool{true}), std::invalid_argument);
+    }
+}
+
+TEST_CASE("NULL_TYPE column state unchanged after rejected push_back", "[column]") {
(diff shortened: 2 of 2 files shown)


im-anishraj (changes requested):
@abhinavuser thanks for the solid native regression coverage here. The implementation and tests are on the right track, and the C++/Python test matrix is passing, but the lint job is failing on clang-format for cpp/src/column.cpp at the new 	hrow std::invalid_argument(...) block. Please run clang-format-17 -i cpp/src/column.cpp cpp/tests/test_column.cpp or otherwise apply the repo formatting, push the update, and I can re-review quickly.

abhinavuser:
@im-anishraj  hey can you check now? vercel is waiting for your authorization to run the preview build
lmk if any changes

im-anishraj (approved):
@abhinavuser thanks for the quick formatting fix. I rechecked the updated PR: clang-format/lint is now green, the C++ native tests pass, and the Python matrix/smoke/wheel checks are successful. The guard is placed before state mutation and the regression tests cover rejection plus null insertion behavior. Approving and merging.

PR #1798, opened 2026-05-28
Fix: treat empty/whitespace strings as null in required_if validation

### LINKED ISSUE #1659: Bug: required_if validation ignores empty string values
## Summary

`required_if` currently checks only `series.isna()` when a condition is met. For string columns, Arnio already treats empty strings and whitespace-only strings as null-like values in nullable validation, but `required_if` does not reuse that null mask.

As a result, a conditionally required string field can pass validation even when the value is `""` or only whitespace.

## Why it matters

Conditional required fields are commonly used for production data contracts, for example requiring `tax_id` when `account_type == "business"`. Empty strings should not satisfy that requirement.

This can let invalid records pass schema validation silently.

## Evidence

- `arnio/schema.py`: `_validate_column()` builds `is_null_mask` that treats empty/whitespace strings as null-like.
- `arnio/schema.py`: the `required_if` branch later uses `series.isna()` directly instead of `is_null_mask`.
- Runtime reproduction shows empty and whitespace values pass when the condition is true.

## Reproduction or trigger

```python
import pandas as pd
import arnio as ar

    "kind": ["business", "personal", "business"],
    "tax_id": ["", None, "   "],
}))

schema = ar.Schema({
    "kind": ar.String(),
    "tax_id": ar.String(required_if=("kind", "business")),
})

print(result.to_dict())
```

Current output:

```python
{'passed': True, 'row_count': 3, 'issue_count': 0, 'bad_rows': [], 'issues': []}
```

## Expected behavior

Rows where `kind == "business"` and `tax_id` is empty or whitespace-only should produce `required_if` validation issues.

## Actual behavior

The result passes because `required_if` only checks `series.isna()`.

## Suggested direction

Use the existing `is_null_mask` when evaluating `required_if` instead of `series.isna()`. Add tests for:

- empty string under a triggered condition
- whitespace-only string under a triggered condition
- real `None` / `pd.NA`
- non-triggered rows remaining valid

## References

Exodus2004:
Fixes #1659

### DIFF
--- a/arnio/cleaning.py
+++ b/arnio/cleaning.py
@@ -247,7 +247,7 @@ def drop_columns(frame: ArFrame, columns: Sequence[str]) -> ArFrame:
         ),
     )
     if len(requested_columns) == 0:
-        return frame
+        return frame.drop_columns([])
     if len(requested_columns) == len(frame.columns):
         raise ValueError("drop_columns cannot remove all columns from the frame")
 
--- a/arnio/schema.py
+++ b/arnio/schema.py
@@ -2205,7 +2205,7 @@ def _validate_column(
             )
         else:
             trigger_mask = df[condition_column] == expected_value
-            invalid = series[trigger_mask & series.isna()]
+            invalid = series[trigger_mask & is_null_mask]
 
             issues.extend(
                 _row_issues(
--- a/tests/test_cleaning.py
+++ b/tests/test_cleaning.py
@@ -683,9 +683,12 @@ class TestDropColumns:
     def test_drop_columns_allows_empty_input_as_no_op(self, sample_csv):
         frame = ar.read_csv(sample_csv)
 
-        result = ar.drop_columns(frame, [])
+        result_helper = ar.drop_columns(frame, [])
+        result_method = frame.drop_columns([])
 
-        assert result is frame
+        assert result_helper is not frame
+        assert result_helper == frame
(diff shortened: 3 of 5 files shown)


im-anishraj (changes requested):
@Exodus2004 thanks for updating the required_if handling. The core idea in `schema.py` looks aligned with the issue, but I cannot merge this branch yet because it includes unrelated files and fails diff hygiene.

Current diff includes unrelated changes outside the required_if/schema scope:
- `arnio/cleaning.py`
- `tests/test_cleaning.py`
- `tests/test_pipeline.py`

It also adds `tests/test_issue_1659.py` with trailing whitespace:

```text
tests/test_issue_1659.py:18: trailing whitespace.
tests/test_issue_1659.py:23: trailing whitespace.
```

Please clean/rebase the branch so it contains only the required_if schema fix and its focused tests, then run `git diff --check`, Black, Ruff, and the relevant schema tests. After that I can re-review and test it properly.

Exodus2004:
i have created a new PR please verify it #1844

im-anishraj:
@Exodus2004 quick update: the clean focused replacement PR #1844 has been reviewed, approved, and merged into `main`. This older PR is now superseded by #1844, so please feel free to close this branch/PR when you get a chance to keep the queue tidy.

Thanks again for fixing the required_if empty/whitespace handling.

PR #1802, opened 2026-05-28
fix: respect system color-scheme preference before manual toggle

### LINKED ISSUE #1769: Website: respect system color-scheme preference before manual toggle
## Summary
The website theme script does not actually respect the visitor's system color-scheme preference before they manually choose a theme.

## Why it matters
A visitor who prefers dark mode at the OS/browser level will still get the light theme on first visit. This is a small accessibility and UX issue because the page appears to support system preference changes, but the implementation prevents that path from working.

## Evidence
- `website/js/theme.js`: `getPreferred()` returns `LIGHT` whenever `localStorage` does not already contain `arnio-theme`.
- `website/js/theme.js`: `apply(getPreferred())` runs immediately and writes that light value into `localStorage`.
- `website/js/theme.js`: the `prefers-color-scheme` listener only applies system changes when `!localStorage.getItem(STORAGE_KEY)`, but the initial `apply()` call has already populated that key.

## Reproduction or trigger
1. Open the website in a browser/profile with no existing `arnio-theme` localStorage entry.
2. Set the OS/browser color scheme to dark.
3. Load the website.
4. Observe the initial theme and then change the OS/browser color scheme again.

## Expected behavior
Before a manual theme toggle is used, the website should follow `prefers-color-scheme`. After a manual toggle, the explicit stored preference should take priority.

## Actual behavior
The site defaults to light mode, stores that as an explicit preference immediately, and system preference changes no longer apply.

## Suggested direction
Separate detecting a preferred theme from persisting a manual choice:

- read `localStorage` first
- if absent, use `window.matchMedia('(prefers-color-scheme: dark)').matches`
- apply the theme without writing to storage during initial auto-detection
- only write to storage when the user clicks the theme toggle

## References
- `website/js/theme.js`

AMAN194701:
## Summary
Fix system color-scheme preference being ignored on first visit.

## Linked issue
Fixes #1769

## Type of change
- [x] Bug fix

## Area
- [x] Docs / website

## What changed in `website/js/theme.js`
1. `getPreferred()` now falls back to `prefers-color-scheme` instead of hardcoding `LIGHT`
2. Added `applyWithoutSaving()` — applies theme on load without writing to localStorage
3. Initial page load uses `applyWithoutSaving()` so system preference stays active until user manually toggles

## Testing
No build required — open website in browser with OS set to dark mode,
confirm dark theme loads on first visit without manual toggle.

## Contributor checklist
- [x] My PR title uses Conventional Commits (`fix:`).

### DIFF
--- a/website/js/theme.js
+++ b/website/js/theme.js
@@ -12,8 +12,12 @@
   function getPreferred() {
     const stored = localStorage.getItem(STORAGE_KEY);
     if (stored === DARK || stored === LIGHT) return stored;
-    // Default to light — only use dark if user explicitly toggles
-    return LIGHT;
+    // Fall back to system preference
+    return window.matchMedia && window.matchMedia('(prefers-color-scheme: dark)').matches ? DARK : LIGHT;
+  }
+  function applyWithoutSaving(theme) {
+    document.documentElement.setAttribute('data-theme', theme);
+    updateToggleIcon(theme);
   }
 
   function apply(theme) {
@@ -35,7 +39,7 @@
   }
 
   // Apply immediately to prevent flash
-  apply(getPreferred());
+  applyWithoutSaving(getPreferred());
 
   // Listen for system preference changes
   if (window.matchMedia) {
--- a/website/js/theme.js
+++ b/website/js/theme.js
@@ -45,7 +45,7 @@
   if (window.matchMedia) {
     window.matchMedia('(prefers-color-scheme: dark)').addEventListener('change', function (e) {
       if (!localStorage.getItem(STORAGE_KEY)) {
-        apply(e.matches ? DARK : LIGHT);
+        applyWithoutSaving(e.matches ? DARK : LIGHT);
       }
     });
   }


im-anishraj (changes requested):
@AMAN194701 thanks for the focused fix. The first-load path is almost right, but one behavior still needs correction before merge: the `matchMedia('(prefers-color-scheme: dark)')` change listener still calls `apply(...)` when there is no stored manual theme. `apply()` writes to `localStorage`, so simply changing the OS/browser theme will create a persisted manual preference and stop future system changes from being followed.

Please update that listener to use the non-persisting path as well, for example `applyWithoutSaving(e.matches ? DARK : LIGHT)`, so only the toggle button writes `arnio-theme` to localStorage.

CI is otherwise green apart from the Vercel authorization status, so this should be a small follow-up.

AMAN194701:
@im-anishraj  I Updated the matchMedia listener to use applyWithoutSaving as requested, but idk why checks are failing but before this all checks were passing.

im-anishraj (approved):
@AMAN194701 thanks for the update. I rechecked the current PR head against latest `origin/main`:

- PR diff is limited to `website/js/theme.js`
- `git diff --check origin/main...refs/remotes/pull/1802/head` passed
- `node --check website/js/theme.js` passed

The `matchMedia` listener now uses `applyWithoutSaving(...)`, so only the manual toggle writes to `localStorage`. The GitHub CI failure shown on the PR came from the older base before the main example-test repair, not from this website change.

AMAN194701:
@im-anishraj  Thank you

PR #1844, opened 2026-05-28
Fix: treat empty/whitespace strings as null in required_if validation

### LINKED ISSUE #1659: Bug: required_if validation ignores empty string values

Exodus2004:
Fixes #1659

This PR updates the 'required_if' validation logic in 'arnio/schema.py' to use 'is_null_mask' instead of 'series.isna()'. This ensures that empty strings and whitespace-only strings are correctly treated as null-like values (and thus fail the requirement) when the 'required_if' condition is met, matching the behavior of 'nullable' validation.

### Changes:
- Modified '_validate_column' in 'arnio/schema.py' to use the existing 'is_null_mask' for 'required_if' checks.
- Added regression tests in 'tests/test_issue_1659.py' covering:
    - Empty strings under triggered condition (fails)
    - Whitespace strings under triggered condition (fails)
    - 'None' values under triggered condition (fails)
    - Non-triggered rows (passes)

### DIFF
--- a/arnio/cleaning.py
+++ b/arnio/cleaning.py
@@ -247,7 +247,7 @@ def drop_columns(frame: ArFrame, columns: Sequence[str]) -> ArFrame:
         ),
     )
     if len(requested_columns) == 0:
-        return frame
+        return frame.drop_columns([])
     if len(requested_columns) == len(frame.columns):
         raise ValueError("drop_columns cannot remove all columns from the frame")
 
--- a/arnio/schema.py
+++ b/arnio/schema.py
@@ -2205,7 +2205,7 @@ def _validate_column(
             )
         else:
             trigger_mask = df[condition_column] == expected_value
-            invalid = series[trigger_mask & series.isna()]
+            invalid = series[trigger_mask & is_null_mask]
 
             issues.extend(
                 _row_issues(
--- a/tests/test_cleaning.py
+++ b/tests/test_cleaning.py
@@ -683,9 +683,12 @@ class TestDropColumns:
     def test_drop_columns_allows_empty_input_as_no_op(self, sample_csv):
         frame = ar.read_csv(sample_csv)
 
-        result = ar.drop_columns(frame, [])
+        result_helper = ar.drop_columns(frame, [])
+        result_method = frame.drop_columns([])
 
-        assert result is frame
+        assert result_helper is not frame
+        assert result_helper == frame
+        assert result_helper == result_method
 
     def test_drop_columns_rejects_missing_columns(self, sample_csv):
         frame = ar.read_csv(sample_csv)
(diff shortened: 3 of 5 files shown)


im-anishraj (approved):
@Exodus2004 thanks for opening the clean focused branch. I verified this version fixes the required_if empty/whitespace behavior without the unrelated files from #1798.

Local verification passed:
- `python -m black --check arnio/schema.py tests/test_issue_1659.py`
- `python -m ruff check arnio/schema.py tests/test_issue_1659.py`
- `python -m pytest tests/test_issue_1659.py tests/test_schema.py -k "required_if or issue_1659" -q` -> 14 passed

GitHub CI is green except the known Vercel authorization check. Approving and merging this one.

Exodus2004:
Hi @im-anishraj, could you please update the labels on this PR to `level:intermediate` / `gssoc:level-2`? The resolved issue #1659 is labeled as intermediate, so updating this would help align the GSSoC scoring correctly with the issue's difficulty. Thank you!

PR #1856, opened 2026-05-28
Reject unsupported object dtype casts in astype

Sarthak-SethiNsd:
## Summary
Fixes #1005

Fixes silent fallback behavior when calling `ArFrame.astype(object)`.

Previously object dtype casts appeared to succeed but were silently converted back to inferred Arnio dtypes during `from_pandas()` conversion.

This PR:
- rejects unsupported whole-frame `object` casts
- rejects per-column dict object casts
- adds regression tests for both cases

Validation performed:
- pytest tests passed
- ruff checks passed
- black formatting applied

Screenshots of pytest, ruff, and black outputs have also been attached.
image


im-anishraj (changes requested):
@Sarthak-SethiNsd thanks for the PR. I ran the local checks and the added tests pass, but the fix is incomplete because it only rejects the literal `object` type.

Verified locally:
- `git diff --check origin/main...HEAD` passed.
- `black --check arnio/frame.py tests/test_frame.py` passed.
- `ruff check arnio/frame.py tests/test_frame.py` passed.
- `pytest tests/test_frame.py -k "astype or Astype or object" -q` passed: 5 tests.

Blocking behavior gap:
The same unsupported object cast still succeeds through common aliases:

```python
frame.astype("object")
frame.astype(np.object_)
frame.astype(np.dtype("O"))
frame.astype({"a": "object"})
frame.astype({"a": np.object_})
frame.astype({"a": np.dtype("O")})
```

I verified those on this branch and they are accepted. Please normalize dtype targets before casting and reject all object dtype forms, not only `target_dtype is object`. Also please keep unrelated test annotation churn out if possible so the behavior change stays easy to review.

im-anishraj (changes requested):
@Sarthak-SethiNsd thanks for the update. The new alias handling covers the object-dtype forms I called out earlier, but this PR still is not merge-ready.

Remaining blockers:
- GitHub reports the branch as conflicting with main; please rebase or merge latest main and resolve conflicts.
- The diff still contains a large amount of unrelated test annotation churn in 	ests/test_frame.py. Please remove that unrelated noise so the PR stays focused on stype(object) rejection and the regression tests.

After those cleanup items, I can re-check the final diff.

im-anishraj:

PR #1858, opened 2026-05-28
fix: validate nested JSONL values and include line/field context

### LINKED ISSUE #1709: Bug: include line context for nested read_jsonl values
## Summary

`read_jsonl()` does not add JSONL-specific context when a row contains nested values. It collects records, builds a pandas DataFrame, and then lets `from_pandas()` raise a generic `TypeError` without the JSONL line number.

## Why it matters

Nested objects and arrays are common in log/event JSONL files. Arnio may intentionally require flat records, but users need a line number and field name to fix the malformed row quickly.

## Evidence

- `arnio/io.py`: `read_jsonl()` validates that each line decodes to a top-level JSON object, but does not validate nested field values before appending the record.
- `arnio/convert.py`: `from_pandas()` later rejects nested `list` and `dict` values without JSONL line context.
- Runtime behavior raises `TypeError: Column 'tags' contains unsupported nested value...` instead of `JsonlReadError` with a line number.

## Reproduction or trigger

```python
from pathlib import Path
import arnio as ar

path = Path("events.jsonl")
path.write_text('{"id": 1, "tags": ["a"]}\n')

ar.read_jsonl(path)
```

## Expected behavior

`read_jsonl()` should raise `JsonlReadError` that includes the line number and field name, for example `Nested JSON value in column 'tags' on line 1`.

## Actual behavior

The function raises a generic `TypeError` from the pandas conversion layer without the JSONL line number.

## Suggested direction

Validate each parsed object before appending it to `records`. Reject nested list/dict values with `JsonlReadError` that includes `path`, line number, column/key, and nested type. Add tests for nested list and nested object values.

## References

- `arnio/io.py`
- `arnio/convert.py`
- `tests/test_jsonl.py`

Nicks-19:
## Summary

When I read JSONL files with structures like lists or dictionaries the parser used to create a Pandas DataFrame. Then it would let the conversion layer throw an error with no details, about where the problem occurred.

This PR update changes the `read_jsonl` function in `arnio/io.py`. Now it checks each objects values as it reads the lines. If it finds a problem it  immediately raises a descriptive `JsonlReadError` that explains what went wrong. The error message includes the key/column ,the type of value that caused the issue, the line number and the file path. I also updated the functions description to explain how it handles errors.

## Linked issue
Fixes #1709

## Type of change
- [x] Bug fix
- [x] Documentation
- [x] Tests

## Area
- [x] CSV parser / I/O

## Testing
- [x] `make lint` (or run separately:
  ```powershell
  python -m ruff check .
  python -m black --check .

### DIFF
--- a/arnio/io.py
+++ b/arnio/io.py
@@ -1040,7 +1040,8 @@ def read_jsonl(
         ``nrows`` is not a non-negative integer.
     JsonlReadError
         If the file is empty (no data rows), or if a line contains invalid
-        JSON.  The error message includes the 1-based line number.
+        JSON or unsupported nested values. The error message includes the
+        1-based line number.
 
     Examples
     --------
@@ -1049,7 +1050,7 @@ def read_jsonl(
     """
     import json
 
-    from .convert import from_pandas
+    from .convert import _is_nested, from_pandas
 
     if not isinstance(encoding, str):
         raise TypeError(f"encoding must be a string, got {type(encoding).__name__!r}")
@@ -1100,6 +1101,13 @@ def read_jsonl(
                         f"Expected a JSON object on line {lineno} of {path!r}, "
                         f"got {type(obj).__name__}"
                     )
+                for key, val in obj.items():
+                    if _is_nested(val):
+                        raise JsonlReadError(
+                            f"Column {key!r} contains unsupported nested value "
+                            f"of type {type(val).__name__!r} on line {lineno} of {path!r}. "
+                            "Convert nested objects to strings or flatten them first."
+                        )
                 records.append(obj)
     except OSError as exc:
         raise JsonlReadError(str(exc)) from exc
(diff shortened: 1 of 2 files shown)


im-anishraj (approved):
@Nicks-19 thanks for the focused JSONL error handling improvement. I reviewed and tested the branch, and the behavior is much clearer now for nested JSONL values.

Verified locally:
- `git diff --check origin/main...HEAD` passed.
- `black --check arnio/io.py tests/test_jsonl.py` passed.
- `ruff check arnio/io.py tests/test_jsonl.py` passed.
- `pytest tests/test_jsonl.py -q` passed: 45 tests.
- Manual repro with a nested dict now raises `JsonlReadError` with the field name, nested type, line number, and path.

The Vercel failure is the known authorization-only status and is unrelated to this Python I/O change. Approving and merging.

PR #1862, opened 2026-05-28
fix(pipeline): prevent custom steps from overwriting Python-backed built-in steps

### LINKED ISSUE #1658: Bug: prevent custom steps from overwriting Python-backed built-ins
## Summary

`register_step(..., overwrite=True)` can currently replace Python-backed built-in pipeline steps such as `filter_rows`, `safe_divide_columns`, `replace_values`, and `standardize_missing_tokens`.

The protection in `register_step()` only checks the native `_STEP_REGISTRY`, so built-ins stored in `_PYTHON_STEP_REGISTRY` can be overwritten by user code. That makes pipeline behavior depend on previous custom registrations and can silently replace official Arnio behavior.

## Why it matters

Pipeline step names are part of Arnio's public API. Built-in step names should be stable and predictable, especially for reusable cleaning pipelines, examples, and GSSoC contributor work.

If a custom step accidentally reuses a built-in Python-backed name, future pipeline calls may execute unrelated code without a clear warning.

## Evidence

- `arnio/pipeline.py`: `register_step()` protects names only when `name in _STEP_REGISTRY`.
- `arnio/pipeline.py`: Python-backed built-ins are stored separately in `_PYTHON_STEP_REGISTRY` / `_BUILTIN_PYTHON_STEP_REGISTRY`.
- The runtime behavior allows replacing a built-in Python-backed step name when `overwrite=True` is passed.

## Reproduction or trigger

1. Register a custom step using the name of a Python-backed built-in:

   ```python
   import pandas as pd
   import arnio as ar
   from arnio.pipeline import register_step, reset_steps

   def fake_step(frame, **kwargs):
       print("fake step called")
       return frame

   register_step("filter_rows", fake_step, overwrite=True)
   frame = ar.from_pandas(pd.DataFrame({"value": [1, 2, 3]}))
   ar.pipeline(frame, [("filter_rows", {"column": "value", "op": ">", "value": 1})])
   reset_steps()
   ```

2. The custom step runs instead of the official `filter_rows` implementation.

## Expected behavior

Official built-in pipeline step names should not be overwritten accidentally. If overwriting built-ins is intentionally supported, it should require an explicit maintainer-only/unsafe API and be documented clearly.

## Actual behavior

`overwrite=True` can replace Python-backed built-in names because only native `_STEP_REGISTRY` names are protected.

## Suggested direction

Update `register_step()` so it treats `_BUILTIN_PYTHON_STEP_REGISTRY` names as protected built-ins too. Add tests for:

- refusing to overwrite native built-ins
- refusing to overwrite Python-backed built-ins
- still allowing overwrite of previously registered custom steps when `overwrite=True`
- keeping `reset_steps()` behavior unchanged

## References

- `arnio/pipeline.py`
- Existing built-in Python-backed steps: `filter_rows`, `safe_divide_columns`, `replace_values`, `standardize_missing_tokens`

Exodus2004:
Resolves #1658. This PR fixes the issue where \
egister_step\ would allow developers to mistakenly overwrite Python-backed built-in steps (like \ilter_rows\ or \safe_divide_columns\) when \overwrite=True\ is passed. It updates the internal check to respect \_BUILTIN_PYTHON_STEP_REGISTRY\ as a protected namespace, and includes a comprehensive suite of unit tests to prevent future regressions.

### DIFF
--- a/arnio/pipeline.py
+++ b/arnio/pipeline.py
@@ -56,6 +56,7 @@ _PYTHON_STEP_REGISTRY: dict[str, Callable] = {
     "coalesce_columns": cleaning.coalesce_columns,
     "normalize_whitespace": cleaning.normalize_whitespace,
 }
+_BUILTIN_PYTHON_STEP_REGISTRY: dict[str, Callable] = {}
 
 
 @dataclass(frozen=True)
@@ -139,6 +140,11 @@ def register_step(name: str, fn: Callable, overwrite: bool = False):
         if name in _STEP_REGISTRY:
             raise ValueError(
                 f"Cannot register '{name}': conflicts with built-in C++ step. "
+                f"Use a different name."
+            )
+        if name in _BUILTIN_PYTHON_STEP_REGISTRY:
+            raise ValueError(
+                f"Cannot register '{name}': conflicts with built-in Python step. "
                 f"Use a different name."
             )
         if name in _DEPRECATED_STEP_ALIASES:
@@ -542,7 +548,7 @@ register_step("filter_rows", cleaning.filter_rows)
 register_step("drop_columns_matching", cleaning.drop_columns_matching)
 register_step("safe_divide_columns", cleaning.safe_divide_columns)
 register_step("replace_values", cleaning.replace_values)
-_BUILTIN_PYTHON_STEP_REGISTRY = dict(_PYTHON_STEP_REGISTRY)
+_BUILTIN_PYTHON_STEP_REGISTRY.update(_PYTHON_STEP_REGISTRY)
 
 
 def reset_steps() -> None:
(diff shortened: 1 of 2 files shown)


im-anishraj (approved):
@Exodus2004 thanks for the focused fix. I verified the diff, ran Black/Ruff, executed the new safety tests plus the existing pipeline suite locally, and confirmed manually that Python-backed built-in steps cannot be overwritten even with overwrite=True. CI is clean aside from the known Vercel authorization status, so this is ready to merge.

Exodus2004:
Hi @im-anishraj, could you please update the labels on this PR to `level:intermediate` / `gssoc:level-2`? The resolved issue #1658 is labeled as intermediate, so updating this would help align the GSSoC scoring correctly with the issue's difficulty. Thank you!

PR #1866, opened 2026-05-28
fix(io): validate parquet compression argument type

### LINKED ISSUE #1812: Bug: validate write_parquet compression type before codec lookup
## Summary

`write_parquet()` does not validate the `compression` argument type before checking membership in the supported codec set.

## Why it matters

The public docs describe `compression` as a string option. Passing an invalid non-string value should produce a clear Arnio-facing validation error, not a raw Python error such as `TypeError: unhashable type: 'list'`.

## Evidence

- `arnio/io.py`: `write_parquet()` checks `if compression not in _VALID_COMPRESSIONS` without first validating that `compression` is a string.
- `tests/test_parquet.py` covers unknown string compression values and `row_group_size` type validation, but does not cover non-string compression inputs.

## Reproduction or trigger

```python
import pandas as pd
import arnio as ar

frame = ar.from_pandas(pd.DataFrame({"a": [1]}))
ar.write_parquet(frame, "out.parquet", compression=["snappy"])
```

## Expected behavior

A clear `TypeError`, for example `compression must be a string`, before codec validation.

## Actual behavior

A raw Python error is raised:

```text
TypeError: unhashable type: 'list'
```

Other non-string values such as `None`, `123`, or `True` currently produce `ValueError: Unknown compression codec`, which is also less precise than a type error.

## Suggested direction

Add explicit type validation for `compression` before the membership check, and add tests for non-string values such as `None`, `123`, `True`, and `['snappy']`.

## References

- `arnio/io.py`
- `tests/test_parquet.py`

sat-06:
## Summary

Adds explicit type validation for the `compression` argument in `write_parquet()` before codec membership validation.

## Problem

`write_parquet()` previously checked:

```python
if compression not in _VALID_COMPRESSIONS:
```

without first validating that `compression` was a string.

This caused invalid non-string values such as:

```python
None
123
True
["snappy"]
```

to raise inconsistent or raw Python errors like:

```txt
TypeError: unhashable type: 'list'
```

instead of a clear Arnio-facing validation error.

## Changes made

- Added explicit type validation for the `compression` argument
- Non-string compression values now raise:
  ```python
  TypeError("compression must be a string")
  ```
- Preserved existing behavior for unsupported string codecs:
  ```python
  compression="lz4"
  ```
  still raises:
  ```python
  ValueError("Unknown compression codec")
  ```
- Added parametrized regression tests covering:
  - `None`
  - `123`
  - `True`
  - `["snappy"]`

## Result

`write_parquet()` now produces clearer and more consistent validation errors for invalid compression argument types while preserving existing codec validation behavior.

### DIFF
--- a/arnio/io.py
+++ b/arnio/io.py
@@ -1417,7 +1417,7 @@ def write_parquet(
 
     if not isinstance(compression, str):
         raise TypeError("compression must be a string")
-    
+
     if compression not in _VALID_COMPRESSIONS:
         raise ValueError(
             f"Unknown compression codec: {compression!r}. "
--- a/arnio/io.py
+++ b/arnio/io.py
@@ -1415,6 +1415,9 @@ def write_parquet(
             "write_parquet only supports .parquet and .pq files."
         )
 
+    if not isinstance(compression, str):
+        raise TypeError("compression must be a string")
+    
     if compression not in _VALID_COMPRESSIONS:
         raise ValueError(
             f"Unknown compression codec: {compression!r}. "
--- a/tests/test_parquet.py
+++ b/tests/test_parquet.py
@@ -198,6 +198,18 @@ class TestWriteParquetErrors:
         with pytest.raises(ValueError, match="Unknown compression codec"):
             ar.write_parquet(frame, tmp_path / "out.parquet", compression="lz4")
 
+    @pytest.mark.parametrize("compression", [None, 123, True, ["snappy"]])
+    def test_non_string_compression_raises_type_error(self, tmp_path, compression):
+        # Compression type validation happens before codec validation and pyarrow import.
+        frame = ar.from_pandas(pd.DataFrame({"a": [1]}))
+
+        with pytest.raises(TypeError, match="compression must be a string"):
+            ar.write_parquet(
+                frame,
+                tmp_path / "out.parquet",
+                compression=compression,
+            )
+
     def test_missing_pyarrow_raises_import_error(self, tmp_path):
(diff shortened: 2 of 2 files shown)


im-anishraj (changes requested):
@sat-06 thanks for the focused validation fix. The behavior itself looks good locally: `python -m pytest tests/test_parquet.py -q` passes with 32 tests, and targeted Ruff/Black on my local environment did not find issues.

However, I cannot merge while the repository CI lint job is red. The failing GitHub job is `CI / Lint (Black, Ruff, clang-format)`, and the log shows Black 26.5.1 would reformat `arnio/io.py` because of whitespace around the new compression validation block:

```
black --check --diff .
would reformat /home/runner/work/arnio/arnio/arnio/io.py
```

Please run the repo-level formatter, preferably:

```
python -m black arnio/io.py tests/test_parquet.py
python -m ruff check arnio/io.py tests/test_parquet.py
python -m pytest tests/test_parquet.py -q
```

After the CI lint check is green, this should be straightforward to merge. The Vercel authorization status is unrelated; the blocker is the failing Black check.

sat-06:
Hi @im-anishraj 
Thanks for clarifying.
I have updated the pr, all the tests are now passing locally and the cl matrix is green.

im-anishraj (approved):
@sat-06 Thanks for the update. I rechecked the current head locally: Black passes, Ruff passes, and pytest tests/test_parquet.py -q passes with 32 tests. The compression type validation now fails early with a clear TypeError, and the regression coverage is focused. Approved for merge.

PR #1910, opened 2026-05-29
perf: skip csv inference when dtypes cover selected columns

### LINKED ISSUE #1710: Performance: read_csv still does a full inference pass when dtype is already supplied
I noticed this while timing a larger CSV import for GSSoC 2026 where I had already supplied the schema through `dtype`. I expected that to skip type inference, but the native reader still walks the whole file once just to infer column types, then applies the explicit dtype mapping afterwards and opens the file again for the real parse.

File + line: `cpp/src/csv_reader.cpp:708`

At line 708 the reader starts PASS 1 for full-file type inference. The `config.dtype` override is not applied until around line 832, after that inference pass has already completed. PASS 2 then reopens the same file at line 857 and streams rows into the typed columns.

So for a call like this:

```python
import arnio as ar

frame = ar.read_csv(
    "large_events.csv",
    dtype={
        "event_id": "int64",
        "user_id": "int64",
        "amount": "float64",
        "active": "bool",
    },
)
```

if the dtype mapping covers the selected columns, the reader still pays for full inference before using the user-provided types.

Why this matters:
- large CSVs do twice the I/O even when the schema is already known
- every field is parsed once for inference and then again for final conversion
- this makes explicit dtypes less useful for production/CI imports where schemas are fixed
- it also makes performance harder to reason about because `dtype=` looks like an optimization but does not avoid the expensive pass

Current behavior: explicit dtype is applied only after the full inference pass.

Expected behavior: when the dtype mapping fully covers the selected output columns, the reader should be able to skip value type inference and go directly to the streaming parse path while still preserving header validation, `usecols`, `nrows`, and bad-row handling.

Suggested direction: detect whether `config.dtype` fully covers `col_indices` after the header/usecols are known. If it does, initialize `col_types` from the mapping and avoid the inference loop. Partial mappings can keep the current inference path for unspecified columns. This should come with a regression/performance test that proves `dtype` avoids the double scan for fully specified schemas.

Requested labels: `bug`, `type:performance`, `gssoc`, `gssoc:level-3`, `level:advanced`, `difficulty:advanced`, `area:csv-parser`, `area:python-api`, `backend`, `priority:medium`, `size:m`.

eshaanag:
## Summary

Fixes #1710

This PR skips the full CSV type-inference scan when `dtype` fully covers the selected output columns. The reader still resolves the header/usecols first, validates the dtype mapping, and then goes directly to the streaming parse path with the explicit column types.

This is for GSSoC 2026 and stays scoped to the assigned CSV parser performance issue.

## What changed

- Added shared helpers for resolving selected CSV columns and applying explicit dtype mappings.
- Detects the fully-covered dtype case after header/usecols resolution.
- Skips value type inference only when every selected output column has an explicit dtype.
- Keeps partial dtype mappings on the existing inference path for unspecified columns.
- Preserves `on_bad_lines` behavior by collecting/raising bad rows in the streaming pass when the inference pass is skipped.
- Added regression coverage for fully explicit dtype + `usecols`, bad-line errors, and bad-line warnings.

## Validation

- `.venv-codex/bin/python -m pip install -e '.[dev]'`
- `.venv-codex/bin/python -m pytest tests/test_csv.py -q -k "dtype or bad_line or on_bad_lines or nrows or usecols"` - 21 passed
- `.venv-codex/bin/python -m pytest tests/test_csv.py -q` - 333 passed
- `.venv-codex/bin/python -m pytest -q` - 2340 passed, 47 skipped
- `.venv-codex/bin/ruff check tests/test_csv.py`
- `.venv-codex/bin/black --check tests/test_csv.py`
- `.venv-codex/bin/clang-format --dry-run --Werror cpp/src/csv_reader.cpp`
- `git diff --check`

### DIFF
--- a/cpp/src/csv_reader.cpp
+++ b/cpp/src/csv_reader.cpp
@@ -334,6 +334,70 @@ void validate_row_width(size_t row_number, size_t expected, size_t actual) {
                              std::to_string(expected));
 }
 
+void generate_synthetic_header(std::vector<std::string>& header, size_t column_count) {
+    header.clear();
+    header.reserve(column_count);
+    for (size_t i = 0; i < column_count; ++i) header.push_back("col_" + std::to_string(i));
+    validate_header(header);
+}
+
+std::vector<size_t> resolve_col_indices(const std::vector<std::string>& header,
+                                        const CsvConfig& config) {
+    std::vector<size_t> col_indices;
+    if (config.usecols.has_value()) {
+        for (const auto& name : config.usecols.value()) {
+            auto it = std::find(header.begin(), header.end(), name);
+            if (it == header.end()) throw std::runtime_error("Column not found: " + name);
+            col_indices.push_back(static_cast<size_t>(std::distance(header.begin(), it)));
+        }
+    } else {
+        for (size_t i = 0; i < header.size(); ++i) col_indices.push_back(i);
+    }
+    return col_indices;
+}
+
+struct ExplicitDTypeResult {
+    std::vector<bool> explicit_columns;
+    bool covers_selected_columns = false;
+};
+
+ExplicitDTypeResult apply_explicit_dtypes(const CsvConfig& config,
+                                          const std::vector<std::string>& header,
+                                          const std::vector<size_t>& col_indices,
+                                          std::vector<DType>& col_types) {
+    ExplicitDTypeResult result{std::vector<bool>(col_types.size(), false), false};
+
+    if (!config.dtype.has_value()) {
+        return result;
+    }
+
+    const auto& dtype = config.dtype.value();
+    for (const auto& [column_name, dtype_name] : dtype) {
+        auto header_it = std::find(header.begin(), header.end(), column_name);
+        if (header_it == header.end()) {
+            throw std::runtime_error("Column not found in dtype mapping: " + column_name);
+        }
+        size_t column_index = static_cast<size_t>(std::distance(header.begin(), header_it));
+        bool selected =
+            std::find(col_indices.begin(), col_indices.end(), column_index) != col_indices.end();
+        if (!selected) {
+            throw std::runtime_error("dtype specified for non-selected column: " + column_name);
+        }
+        col_types[column_index] = string_to_dtype(dtype_name);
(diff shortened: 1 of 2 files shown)


im-anishraj:
@eshaanag thanks for the detailed and well-scoped PR. I reviewed the diff and the change is focused on the CSV parser performance path for fully explicit `dtype` mappings, with useful regression coverage around `usecols` and `on_bad_lines` behavior.

I added the matching GSSoC labels (`gssoc`, `level:advanced`, `type:performance`, `type:testing`, plus the CSV/C++ area labels). I am not merging yet because CI currently has a failing lint check and several jobs are still pending. Please check the lint job once the logs are available, fix the formatting/lint issue, and let the full CI finish green. After that this can go through final maintainer review.

im-anishraj:
@eshaanag thanks for the quick follow-up. I rechecked the latest commit and the GitHub Actions side is now green, including lint, C++ native tests, wheel smoke tests, benchmark, examples, notebook smoke, and the full Python matrix.

The only remaining failing status is Vercel authorization, which looks maintainer-side rather than a code issue in this PR. I am holding the merge until that external status is resolved, but no further code changes are requested from you right now.

eshaanag:
Thanks for rechecking. I confirmed the latest CI status as well: the GitHub Actions side is green now, and the remaining Vercel status looks like the external authorization issue you mentioned.

No further code changes from my side right now. I’ll leave this PR as-is and wait for the Vercel authorization/final maintainer review.

im-anishraj (approved):
@eshaanag I rechecked this PR and it is approved from code/test review.

Verification:
- `black --check tests/test_csv.py` passed
- `ruff check tests/test_csv.py` passed
- `C:\Users\anish\anaconda3\python.exe -m pytest tests/test_csv.py -q -k 'fully_explicit_dtype or dtype_with_usecols or dtype_non_selected_usecols_column'` passed: 5 selected tests passed
- GitHub Actions are green across lint, C++ native tests, wheel smoke tests, examples, notebook smoke, benchmark, and the Python matrix

Approved. I am not merging while Vercel still reports the external deployment authorization failure.

PR #1978, opened 2026-05-29
fix: return zero-column ArFrame when select_dtypes matches no columns

### LINKED ISSUE #1954: Bug: select_dtypes should preserve row count for empty matches
## Summary

`ArFrame.select_dtypes()` raises `ValueError` when no columns match the dtype filter.

This differs from pandas-style selection behavior and from Arnio’s current zero-column frame support. A dtype selection that matches no columns should be able to return an empty-column `ArFrame` while preserving the original row count.

## Why it matters

`select_dtypes()` is useful in generic pipelines where the available column types may vary by dataset. Raising on an empty match forces users to add defensive pre-checks before every dtype selection, even though an empty result is a valid and useful outcome.

Arnio already supports zero-column frames in other paths, so this behavior is unnecessarily restrictive.

## Evidence

- `arnio/frame.py`: `select_dtypes()` builds a `matched` list and raises `ValueError` when it is empty.
- `arnio/frame.py`: `select_columns()` also rejects empty selections, so `select_dtypes()` cannot currently delegate to it for a zero-column result.
- Existing zero-column frame tests show Arnio can preserve row counts for empty-column frames.

## Reproduction or trigger

```python
import pandas as pd
import arnio as ar

frame = ar.from_pandas(pd.DataFrame({"age": [1, 2], "score": [3.5, 4.0]}))

result = frame.select_dtypes(include="string")

assert result.shape == (2, 0)  # currently raises ValueError
```

Another trigger:

```python
frame.select_dtypes(exclude=["int64", "float64"])
# currently raises ValueError instead of returning a zero-column frame
```

## Expected behavior

When the dtype filter is valid but matches no columns, return an `ArFrame` with:

- the original row count
- zero columns
- no data loss in the original frame

## Actual behavior

`select_dtypes()` raises `ValueError: No columns match the dtype selection`.

## Suggested direction

Add a zero-column construction path for `select_dtypes()` when the filter itself is valid but `matched` is empty. Add focused tests for:

- `include` with an absent dtype
- `exclude` removing all columns
- row count preservation
- interaction with attrs metadata, if supported by the final implementation

## References

- `arnio/frame.py`
- `tests/test_frame.py`

taniy8:
Fixes #1954

## What changed
- `select_dtypes()` no longer raises `ValueError` when a valid dtype filter matches no columns
- Returns a zero-column `ArFrame` preserving the original row count instead

## Tests added
- `include` with an absent dtype
- `exclude` removing all columns
- row count preservation
- `attrs` metadata copied but not shared

### DIFF
--- a/arnio/frame.py
+++ b/arnio/frame.py
@@ -649,9 +649,7 @@ class ArFrame:
             matched.append(col)
 
         if not matched:
-            raise ValueError(
-                f"No columns match the dtype selection. Frame dtypes: {col_dtypes}."
-            )
+            return ArFrame(_Frame(len(self)), attrs=self._attrs.copy())
 
         return self.select_columns(matched)
 
--- a/arnio/frame.py
+++ b/arnio/frame.py
@@ -581,8 +581,8 @@ class ArFrame:
         ------
         ValueError
             If neither *include* nor *exclude* is provided, if *include*
-            and *exclude* overlap, if an unrecognised dtype string is
-            passed, or if no columns match the filter.
+            and *exclude* overlap, or if an unrecognised dtype string is
+            passed.
         TypeError
             If *include* or *exclude* is not a string, list, or tuple of
             strings.
--- a/tests/test_frame.py
+++ b/tests/test_frame.py
@@ -1314,3 +1314,35 @@ class TestFilterRows:
             match="cannot compare column 'name' with value 18 using operator '>'",
         ):
             ar.filter_rows(frame, column="name", op=">", value=18)
+
+
+# ── select_dtypes zero-column ─────────────────────────────────────────────────
+
+
+def test_select_dtypes_include_absent_dtype_returns_zero_col_frame():
+    frame = ar.from_pandas(pd.DataFrame({"age": [1, 2], "score": [3.5, 4.0]}))
+    result = frame.select_dtypes(include="string")
+    assert result.shape == (2, 0)
+    assert result.columns == []
+
+
+def test_select_dtypes_exclude_all_columns_returns_zero_col_frame():
+    frame = ar.from_pandas(pd.DataFrame({"age": [1, 2], "score": [3.5, 4.0]}))
+    result = frame.select_dtypes(exclude=["int64", "float64"])
+    assert result.shape == (2, 0)
+    assert result.columns == []
+
+
+def test_select_dtypes_zero_col_preserves_row_count():
+    frame = ar.from_pandas(pd.DataFrame({"age": [1, 2, 3, 4, 5]}))
+    result = frame.select_dtypes(include="string")
+    assert result.shape[0] == 5
+    assert result.shape[1] == 0
+
+
+def test_select_dtypes_zero_col_preserves_attrs():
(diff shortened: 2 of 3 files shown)


im-anishraj (changes requested):
@taniy8 thank you for the clean, focused patch and for adding coverage for the zero-column return case. I tested this locally and the new focused tests pass:

- `python -m black --check arnio/frame.py tests/test_frame.py` passed
- `python -m ruff check arnio/frame.py tests/test_frame.py` passed
- `python -m pytest tests/test_frame.py -q -k "select_dtypes"` passed

However, I cannot merge this yet because the existing dedicated `select_dtypes` contract tests now fail, and the remote CI is failing for the same reason. Reproduction:

```bash
python -m pytest tests/test_select_dtypes.py -q
```

Current failures:

- `tests/test_select_dtypes.py::test_select_dtypes_null_dtype_is_valid`
- `tests/test_select_dtypes.py::test_select_dtypes_no_match_raises_value_error`
- `tests/test_select_dtypes.py::test_select_dtypes_exclude_all_raises_value_error`

Those tests still expect `ValueError` when no dtype columns match. Please update this PR so the behavior is internally consistent: either adjust the existing `tests/test_select_dtypes.py` expectations if we are intentionally changing the public contract, or narrow the code change if only specific cases should return an empty frame. Also please update the `select_dtypes()` docstring because it still says no matching columns raises `ValueError`.

Once those are aligned and CI is green, I can re-review quickly.

taniy8:
Hi @im-anishraj, I've made the requested changes:

- Updated the three conflicting tests in `tests/test_select_dtypes.py` to expect a zero-column `ArFrame` instead of `ValueError` on empty match
- Updated the `select_dtypes()` docstring to remove the "no columns match" raise condition

CI should be green now. Please re-review when you get a chance.

Labels from the issue #1954 : `priority:medium`, `area:python-api`, `difficulty:intermediate`, `gssoc`, `gssoc:level-2`, `type:bug`

im-anishraj:
@taniy8 thanks for the update. I fixed the PR tracking labels to match #1954 (`level:intermediate` / `gssoc:level-2`) and see your re-review request.

I will keep this in the review queue for a proper check of the updated tests/docstring before changing the review status.

im-anishraj (approved):
@taniy8 thank you for the quick update. I re-reviewed the latest head and verified the previous blocker is resolved.

Local verification:

- `python -m black --check arnio/frame.py tests/test_frame.py tests/test_select_dtypes.py` passed
- `python -m ruff check arnio/frame.py tests/test_frame.py tests/test_select_dtypes.py` passed
- `python -m pytest tests/test_frame.py tests/test_select_dtypes.py -q -k "select_dtypes"` passed with 32 selected tests
- Extra smoke check confirmed `select_dtypes()` now returns a zero-column frame with the original row count and copied attrs when no dtype columns match

Approving from code review. I also aligned the difficulty label to `level:intermediate` based on the linked issue metadata. I will wait for the remaining GitHub check to finish before merging.

Tiny non-blocking cleanup for later: two updated test names still include `raises_value_error` even though they now assert zero-column returns.

PR #1980, opened 2026-05-29
fix: reject invalid tokens for forced numeric CSV dtypes

### LINKED ISSUE #1923: Bug: forced numeric CSV dtypes silently coerce invalid tokens to null
## Summary
`read_csv(..., dtype={...})` silently converts invalid non-null numeric tokens to null values when a column is forced to `int64` or `float64`.

## Why it matters
Explicit dtype mappings are commonly used to make CSV ingestion deterministic. If a value like `"abc"` appears in a forced numeric column, silently turning it into `None` can hide upstream data quality problems and change downstream aggregates without warning.

## Evidence
- `cpp/src/csv_reader.cpp`: in `CsvParser::parse_value`, the `DType::INT64` and `DType::FLOAT64` branches return `std::monostate{}` when parsing fails.
- `arnio/io.py`: `read_csv()` exposes `dtype` but does not expose an `errors=` or `on_invalid_dtype_value=` policy for explicit dtype parsing.

## Reproduction or trigger
1. Create a CSV:
   ```csv
   age,score
   1,1.2
   abc,bad
   3,4.5
   ```
2. Read it with forced numeric dtypes:
   ```python
   import arnio as ar

   frame = ar.read_csv("num.csv", dtype={"age": "int64", "score": "float64"})
   print(frame.dtypes)
   print(frame.to_dict())
   ```
3. Current output:
   ```python
   {'age': 'int64', 'score': 'float64'}
   {'age': [1, None, 3], 'score': [1.2, None, 4.5]}
   ```

## Expected behavior
Invalid non-null tokens in forced numeric columns should be reported clearly, or Arnio should provide an explicit documented coercion option so users opt into nulling invalid values.

## Actual behavior
Invalid non-null numeric tokens are silently converted to null values.

## Suggested direction
Add explicit error handling for forced dtype parse failures. A safe first step is to raise `CsvReadError` by default and add regression tests for invalid int64 and float64 tokens. If coercion is desired later, expose a documented option rather than making it implicit.

## References
- `cpp/src/csv_reader.cpp`
- `arnio/io.py`
- `tests/test_csv.py`

Exodus2004:
Resolves #1923

### Description
This PR modifies `CsvParser::parse_value` to reject invalid non-null tokens (like `"abc"`) when a column is explicitly forced to `int64` or `float64`. Rather than silently converting these tokens to nulls (`std::monostate{}`), the parser now throws a `CsvReadError`. This aligns forced dtype behavior with user expectations for fail-fast validation.

- Updated `CsvReader` and `CsvChunkReader` to resolve explicit columns and pass the `is_forced` flag to `parse_value`.
- Added regression tests in `tests/test_csv_numeric_parsing.py` for `int64` and `float64` forced types.

### Related issue
Fixes #1923

### DIFF
--- a/arnio/io.py
+++ b/arnio/io.py
@@ -648,6 +648,7 @@ def read_csv_chunked(
     path: str | os.PathLike[str] | io.TextIOBase,
     *,
     chunksize: int = 10_000,
+    dtype: dict[str, str] | None = None,
     delimiter: str | None = None,
     has_header: bool = True,
     usecols: list[str] | None = None,
@@ -815,6 +816,9 @@ def read_csv_chunked(
 
         if null_values is not None:
             config.null_values = _validate_null_values(null_values)
+
+        if dtype is not None:
+            config.dtype = _validate_dtype_mapping(dtype)
 
         if usecols is not None:
             config.usecols = _validate_usecols(usecols)
--- a/cpp/include/arnio/csv_reader.h
+++ b/cpp/include/arnio/csv_reader.h
@@ -53,7 +53,7 @@ class CsvParser {
     bool is_null_sentinel(const std::string& value) const;
     DType infer_type(const std::string& value) const;
     static DType promote_type(DType current, DType incoming);
-    CellValue parse_value(const std::string& raw, DType dtype) const;
+    CellValue parse_value(const std::string& raw, DType dtype, bool is_forced = false) const;
 
    private:
     CsvConfig config_;
@@ -94,6 +94,7 @@ class CsvChunkReader {
     std::vector<std::string> header_;
     std::vector<size_t> col_indices_;
     std::vector<DType> col_types_;
+    std::vector<bool> explicit_dtype_columns_;
     std::optional<size_t> expected_cols_;
     size_t record_number_ = 0;
     size_t rows_read_total_ = 0;
--- a/cpp/src/csv_reader.cpp
+++ b/cpp/src/csv_reader.cpp
@@ -745,9 +745,8 @@ CellValue CsvParser::parse_value(const std::string& raw, DType dtype, bool is_fo
             int64_t value = 0;
             if (!try_parse_int64(cleaned, value)) {
                 if (is_forced) {
-                    throw std::runtime_error(
-                        "CsvReadError: Invalid token '" + raw +
-                        "' for forced int64 column");
+                    throw std::runtime_error("CsvReadError: Invalid token '" + raw +
+                                             "' for forced int64 column");
                 }
                 return std::monostate{};
             }
@@ -758,9 +757,8 @@ CellValue CsvParser::parse_value(const std::string& raw, DType dtype, bool is_fo
             double value = 0.0;
             if (!try_parse_float64(cleaned, value)) {
                 if (is_forced) {
-                    throw std::runtime_error(
-                        "CsvReadError: Invalid token '" + raw +
-                        "' for forced float64 column");
+                    throw std::runtime_error("CsvReadError: Invalid token '" + raw +
+                                             "' for forced float64 column");
                 }
                 return std::monostate{};
             }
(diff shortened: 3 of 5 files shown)


im-anishraj (changes requested):
@Exodus2004 thanks for the focused fix for forced numeric dtype validation. I added the GSSoC classification labels and did an initial local/CI review.

This cannot merge yet because the lint job is still failing. I reproduced the current Python lint failure locally:

```bash
python -m ruff check tests/test_csv_numeric_parsing.py
```

Ruff reports `I001` for the local imports inside both new tests. Please move these imports to the module top, for example:

```python
import pytest

from arnio import read_csv
from arnio.exceptions import CsvReadError
```

Then remove the repeated `import pytest` / `from arnio.exceptions import CsvReadError` lines inside the test functions.

I also tried to rebuild the editable package locally before trusting the C++ behavior tests, but this Windows environment is missing `nmake`, so the local C++ rebuild could not complete. Please rely on the GitHub CI matrix after fixing lint; once the lint job and test matrix are green, I can re-review the parser behavior and merge if everything checks out.

im-anishraj (changes requested):
@Exodus2004 thanks for the follow-up fixes. The previous Python import/lint issue is addressed locally:

- `python -m black --check tests/test_csv.py tests/test_csv_numeric_parsing.py` passed
- `python -m ruff check tests/test_csv.py tests/test_csv_numeric_parsing.py` passed

I still need to request changes because the `read_csv_chunked()` path can bypass this new forced-dtype validation.

In `CsvChunkReader::next_chunk()`, after `apply_explicit_dtypes()` sets `explicit_dtype_columns_`, schema inference still runs for every selected column:

```cpp
DType inferred = parser_.infer_type(row[ci]);
col_types_[ci] = CsvParser::promote_type(col_types_[ci], inferred);
```

That means a forced `int64`/`float64` column can be promoted to `string` before `build_frame()` calls `parse_value()`. Since `parse_value()` only raises for forced numeric dtypes when the effective dtype is `INT64` or `FLOAT64`, an invalid token can be parsed as a string in the chunked path instead of raising `CsvReadError`.

Please mirror the non-chunked fix in `CsvChunkReader::next_chunk()` by skipping type inference/promotion for `explicit_dtype_columns_[ci]`, and add chunked regression coverage such as:

```python
with pytest.raises(ar.CsvReadError, match="Invalid token 'abc' for forced int64 column"):
    list(ar.read_csv_chunked(path, chunksize=1, dtype={"a": "int64"}))
```

It would also be useful to add one non-chunked partial-dtype regression, for example `dtype={"a": "int64"}` with another inferred column, so both fully-explicit and partially-explicit mappings are covered.

Once that chunked path is covered and CI is green, I can re-review quickly.

Exodus2004:
@im-anishraj All requested changes have been implemented:

1. Skipped type inference and promotion for forced dtype columns in CsvChunkReader::next_chunk() to mirror the non-chunked CsvReader behavior.
2. Added the dtype parameter to the Python signature of 
ead_csv_chunked() and mapped it to the reader configuration.
3. Added chunked and partial-dtype validation regression tests in 	ests/test_csv_numeric_parsing.py.
4. Resolved all clang-format formatting violations.

All CI checks (Lint, C++ Native Tests, Notebook Smoke Tests, scikit-learn Integration, and the Python matrix across all OSes) are fully green and passing. Ready for your review and merge!

im-anishraj:
@Exodus2004 thanks for the detailed update. I see the new chunked-reader and formatting fixes, and I’ll keep PR #1980 in the review queue for a proper diff/test check before changing the review status.

im-anishraj (approved):
Approved, @Exodus2004. The latest commit addresses the requested chunked-reader gap: explicit dtype columns now skip inference/promotion in `CsvChunkReader::next_chunk()`, and the new tests cover forced numeric failures, partial dtype mappings, and chunked reads.

I rechecked the Python formatting/lint locally:

- `black --check tests/test_csv.py tests/test_csv_numeric_parsing.py arnio/io.py`
- `ruff check tests/test_csv.py tests/test_csv_numeric_parsing.py arnio/io.py`

Those all pass. The local targeted pytest run cannot be used as authoritative here because this Windows checkout does not have the updated C++ extension rebuilt; the GitHub CI matrix did rebuild the extension and all non-Vercel checks are green. The remaining Vercel failure is the known authorization-limit issue, so it is not blocking this parser fix.

PR #1992, opened 2026-05-29
fix: make code block copy buttons visible on keyboard focus

### LINKED ISSUE #1925: Accessibility: reveal code copy buttons on keyboard focus
## Summary
Code-block copy buttons are keyboard-focusable but remain visually hidden when focused from the keyboard.

## Why it matters
Keyboard users can tab onto an invisible control. That makes the website harder to navigate and fails the expected focus-visible behavior for interactive elements.

## Evidence
- `website/js/code.js`: dynamically inserts a real `<button class="copy-btn">Copy</button>` into code blocks, so the control is part of the tab order.
- `website/css/components.css`: `.copy-btn` starts with `opacity: 0`.
- `website/css/components.css`: opacity is only raised for `.code-block:hover .copy-btn`, `pre:hover .copy-btn`, and `.copy-btn.copied`; there is no `.copy-btn:focus` or `.copy-btn:focus-visible` rule.
- The existing closed install-command keyboard issue (#1641) covered `.hero-install`; this is a separate focus-visibility problem for generated code-block buttons.

## Reproduction or trigger
1. Open any website page with a code block, such as `website/docs.html`.
2. Use the keyboard Tab key to navigate through the page.
3. When focus reaches the generated copy button inside a code block, the button can be focused while still invisible unless the mouse is hovering over the code block.

## Expected behavior
The copy button should become visible when it receives keyboard focus, ideally through `.copy-btn:focus-visible` styling.

## Actual behavior
The copy button opacity remains `0` for keyboard focus, so users can land on a hidden interactive control.

## Suggested direction
Add focus-visible CSS for the generated copy button, for example making `.copy-btn:focus-visible` visible and giving it a clear outline. Consider also revealing the button on `.code-block:focus-within` so the surrounding code block behaves consistently for keyboard users.

## References
- `website/js/code.js`
- `website/css/components.css`
- Related but distinct closed issue: #1641

CoderAnshuman:
## Summary

This PR improves keyboard accessibility for generated code-block copy buttons by making them visible when focused via keyboard navigation. A clear focus indicator has also been added to ensure users can easily identify the focused control without relying on mouse hover.

Fixes #1925

## Changes

* Added `.copy-btn:focus-visible` styling.
* Made copy buttons visible on keyboard focus.
* Added a visible focus outline for accessibility.

## Manual Testing

1. Open a page containing code blocks.
2. Navigate using the `Tab` key.
3. Verify the copy button becomes visible when focused.
4. Verify a clear focus indicator is displayed.
5. Confirm existing hover and copy functionality remain unchanged.

### DIFF
--- a/website/css/components.css
+++ b/website/css/components.css
@@ -445,6 +445,12 @@ pre:hover .copy-btn {
   opacity: 1;
 }
 
+.copy-btn:focus-visible {
+  opacity: 1;
+  outline: 2px solid currentColor;
+  outline-offset: 2px;
+}
+
 .copy-btn:hover {
   background: var(--color-accent);
   color: #ffffff;


im-anishraj (approved):
Approved, @CoderAnshuman. This is a focused accessibility fix: the copy button now becomes visible on keyboard focus and gets a visible focus indicator without changing the existing hover/copied behavior.

I added the GSSoC labels and verified the patch locally with `git diff --check origin/main...HEAD`. The only failing status is the known Vercel authorization-limit check, so I am not treating that as blocking for this website CSS-only change.

PR #1993, opened 2026-05-29
fix(csv): preserve mid-field quotes

### LINKED ISSUE #1890: Bug: CSV parser drops mid-field quote characters
## Summary
The native CSV parser silently drops quote characters that appear in the middle of an unquoted field. For example, `ab"cd"` is parsed as `abcd`, which changes the user's data without raising an error.

## Why it matters
Silent data mutation is risky in CSV preprocessing. If a quote character is part of a dirty but valid-looking field, Arnio should either preserve it as data or reject the row clearly. Dropping it can corrupt identifiers, notes, product codes, and free-text columns.

## Evidence
- `cpp/src/csv_reader.cpp`: `CsvParser::parse_line()` enters quoted mode whenever it sees `"`, regardless of whether the quote appears at the beginning of a field.
- `tests/test_csv_fuzz.py`: `test_malformed_mid_field_quotes` currently allows either parsing or raising, but does not assert that quote characters are preserved if parsing succeeds.
- Local repro on current `origin/main` (`4b99b03`) shows Arnio changes `ab"cd"` to `abcd`.

## Reproduction or trigger
1. Create a CSV:

```csv
a,b
ab"cd",2
```

2. Read it with Arnio:

```python
import arnio as ar

frame = ar.read_csv("midfield_quote.csv")
print(ar.to_pandas(frame).to_dict("list"))
```

## Expected behavior
Arnio should preserve the quote characters as field data, or reject the malformed field with a clear `CsvReadError`.

## Actual behavior
Arnio parses the value as `abcd`, silently removing the embedded quote characters.

## Suggested direction
Make quote handling state-aware:

- only treat `"` as opening a quoted field when it appears at the start of a field, or
- reject mid-field quote characters in strict mode, and
- add regression tests for `ab"cd"`, `"ab"cd`, and properly escaped `"ab""cd"` cases.

## References
- `cpp/src/csv_reader.cpp`
- `tests/test_csv_fuzz.py`
- closed issue #113 covered general malformed quote diagnostics, but this is a distinct data-preservation regression for mid-field quotes.

saurabhhhcodes:
## Summary
- make CSV quote handling context-aware so quotes only enter quoted mode at field start
- preserve mid-field quote characters as literal CSV data instead of silently dropping them
- add C++ parser and Python regression coverage for mid-field quotes, escaped quotes, delimiter-adjacent quoted fields, and quoted fields followed by trailing text

## Verification
- `clang++ -std=c++17 -Icpp/include -c cpp/src/csv_reader.cpp -o /private/tmp/arnio_csv_reader.o`
- `clang++ -std=c++17 -Icpp/include /private/tmp/arnio_parser_smoke.cpp cpp/src/csv_reader.cpp cpp/src/frame.cpp cpp/src/column.cpp -o /private/tmp/arnio_parser_smoke`
- `/private/tmp/arnio_parser_smoke`
- `PYTHONPYCACHEPREFIX=/private/tmp/arnio-1890-pycache python3 -m compileall -q tests/test_csv.py tests/test_csv_fuzz.py`
- `git diff --check`

Could not run the CMake/Catch2 target locally because `cmake` is not installed in this environment. Could not run the Python regression tests locally because the editable extension is not built here and `pip install -e . --no-build-isolation` failed with missing `scikit_build_core`.

Closes #1890

### DIFF
--- a/cpp/src/csv_reader.cpp
+++ b/cpp/src/csv_reader.cpp
@@ -44,16 +44,34 @@ inline void strip_utf8_bom(std::string& s) {
     }
 }
 
-inline bool record_complete(const std::string& record) {
+inline bool record_complete(const std::string& record, char delimiter) {
     bool in_quotes = false;
+    bool at_field_start = true;
 
     for (size_t i = 0; i < record.size(); ++i) {
-        if (record[i] != '"') continue;
+        const char c = record[i];
 
-        if (in_quotes && i + 1 < record.size() && record[i + 1] == '"') {
-            ++i;
-        } else {
-            in_quotes = !in_quotes;
+        if (in_quotes) {
+            if (c == '"') {
+                if (i + 1 < record.size() && record[i + 1] == '"') {
+                    ++i;
+                } else {
+                    in_quotes = false;
+                    at_field_start = false;
+                }
+            }
+            continue;
+        }
+
+        if (c == '"') {
+            if (at_field_start) {
+                in_quotes = true;
+                at_field_start = false;
+            }
+        } else if (c == delimiter) {
+            at_field_start = true;
+        } else if (c != '\r') {
+            at_field_start = false;
         }
     }
 
@@ -137,7 +155,8 @@ class BufferedStreamReader {
 
 class RecordReader {
    public:
-    explicit RecordReader(std::istream& stream) : reader_(stream) {
+    explicit RecordReader(std::istream& stream, char delimiter)
+        : reader_(stream), delimiter_(delimiter) {
         record_.reserve(1024);
         line_.reserve(1024);
     }
@@ -161,13 +180,13 @@ class RecordReader {
             prev_line_ending_ = line_ending_;
             first = false;
 
-            if (record_complete(record_)) {
+            if (record_complete(record_, delimiter_)) {
                 out_record = record_;
                 return true;
             }
         }
 
-        if (!record_.empty() && !record_complete(record_)) {
+        if (!record_.empty() && !record_complete(record_, delimiter_)) {
             throw std::runtime_error("Unterminated quoted field starting at line " +
                                      std::to_string(record_start_line));
         }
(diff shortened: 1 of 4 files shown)


saurabhhhcodes:
Thanks for the fast CI run. All GitHub checks are green or finishing green on this parser fix; the only visible red status is Vercel authorization, which needs a team member and is not code-actionable from my fork.

Small scoring-label request: issue #1890 is marked difficulty:advanced / gssoc:level-3 and type:bug. Could you please mirror the advanced/level-3 bug labels on this PR as well?

im-anishraj:
Thanks for the parser fix, @saurabhhhcodes. I added the missing GSSoC classification labels and did an initial static/local pass:

- `black --check tests/test_csv.py tests/test_csv_fuzz.py`
- `ruff check tests/test_csv.py tests/test_csv_fuzz.py`

Those are clean. I also reviewed the C++ diff around `record_complete()` / `parse_line()`. I cannot use this local MinGW toolchain as an authoritative C++ compile here because it fails on the standard `<optional>` header, so I am waiting for the rebuilt GitHub Actions matrix. At the moment, all non-Vercel checks are green except one Windows Python job that is still running. I’ll hold review/merge until that finishes.

im-anishraj:
@saurabhhhcodes thanks for flagging the tracking mismatch. I updated PR #1993 to mirror #1890 with `level:advanced` and `gssoc:level-3`; the parser bug/type labels are already present.

im-anishraj (approved):
@saurabhhhcodes thanks for the careful fix and the regression coverage around mid-field quote handling. I rechecked the full GitHub CI matrix: all non-Vercel jobs are green, including the C++ native tests and Python test matrix. The remaining Vercel failure is the known repository auth-limit issue, not a code failure.

Local Python formatting/static checks are clean. My local targeted pytest run cannot validate the new C++ behavior because this Windows workspace is still importing the previously installed `_arnio_cpp` wheel, and rebuilding the native extension here is blocked by missing NMake/MSVC. The rebuilt CI wheels cover the native path, so this is ready from my side.

PR #1996, opened 2026-05-29
fix: avoid returning original frame from drop_empty_columns zero-row path

### LINKED ISSUE #1950: Bug: avoid returning original frame from drop_empty_columns zero-row path
## Summary

`drop_empty_columns()` returns the original `ArFrame` when the input has zero rows.

That preserves the current shape, but it also aliases the original wrapper and mutable `_attrs` metadata instead of returning a fresh transformed frame.

## Why it matters

Cleaning helpers should have predictable copy/transform semantics. Returning the original object only on the zero-row path makes this helper behave differently from non-empty inputs and can cause accidental mutation of the caller’s original frame metadata.

The docstring also describes the return value as a new frame, so this edge path does not match the public contract.

## Evidence

- `arnio/cleaning.py`: `drop_empty_columns()` returns `frame` directly when `frame.shape[0] == 0`.
- `arnio/cleaning.py`: the docstring says the function returns a “New frame without fully empty columns.”
- `tests/test_cleaning.py`: zero-row schema preservation is tested, but object identity and `_attrs` aliasing are not tested.

## Reproduction or trigger

```python
import pandas as pd
import arnio as ar

frame = ar.from_pandas(pd.DataFrame({"a": pd.Series(dtype="object")}))
frame._attrs = {"nested": {"x": 1}}

result = ar.drop_empty_columns(frame)

assert result is not frame  # currently fails

result._attrs["nested"]["x"] = 2
assert frame._attrs["nested"]["x"] == 1  # currently fails
```

## Expected behavior

Zero-row inputs should still return a fresh `ArFrame` with the same columns and row count, without sharing mutable metadata with the original frame.

## Actual behavior

The zero-row branch returns the original frame object directly.

## Suggested direction

Replace the zero-row early return with a fresh zero-row frame construction path that preserves columns, dtypes, shape, and a deep copy of `_attrs`. Add focused tests for:

- zero-row inputs return a distinct `ArFrame`
- `_attrs` metadata is not aliased
- columns/dtypes/shape are preserved
- non-empty behavior remains unchanged

## References

- `arnio/cleaning.py`
- `tests/test_cleaning.py`

Exodus2004:
Closes #1950. Preserves schema, shape, and deep copies _attrs on zero-row inputs to avoid wrapper and metadata aliasing.

### DIFF
--- a/arnio/cleaning.py
+++ b/arnio/cleaning.py
@@ -539,7 +539,15 @@ def drop_empty_columns(frame: ArFrame) -> ArFrame:
     from .convert import to_pandas
 
     if frame.shape[0] == 0:
-        return frame
+        attrs = copy.deepcopy(frame._attrs) if frame._attrs is not None else None
+        empty_columns_data: dict[str, list[object]] = {}
+        empty_dtype_hints: dict[str, _DType] = {}
+        for col_name in frame.columns:
+            empty_columns_data[col_name] = []
+            empty_dtype_hints[col_name] = frame._frame.column_by_name(col_name).dtype()
+        return ArFrame(
+            _Frame.from_dict(empty_columns_data, empty_dtype_hints, 0), attrs=attrs
+        )
 
     df = to_pandas(frame)
     empty_columns: list[str] = []
--- a/tests/test_cleaning.py
+++ b/tests/test_cleaning.py
@@ -930,6 +930,20 @@ class TestDropEmptyColumns:
         assert result.shape[0] == 0
         assert result.shape[1] == 3
 
+    def test_drop_empty_columns_zero_row_metadata_isolation(self):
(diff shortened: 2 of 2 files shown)


im-anishraj (approved):
@Exodus2004 thanks for the focused fix. I reviewed the diff and verified the zero-row `drop_empty_columns` path locally:

- `black --check arnio/cleaning.py tests/test_cleaning.py` passed
- `ruff check arnio/cleaning.py tests/test_cleaning.py` passed
- `pytest tests/test_cleaning.py -k "DropEmptyColumns or drop_empty_columns" -q` passed: 6 tests

All non-Vercel CI is green; the remaining Vercel failure is the known authorization-limit issue. This is ready to merge.

PR #2003, opened 2026-05-29
fix: guard unsafe float-to-int64 conversions

### LINKED ISSUE #1591: Bug: unsafe float-to-int64 casts in clip_numeric and coerce_value can trigger undefined behavior and silent numeric corruption
### What happened?

While auditing the native numeric cleaning kernels, I discovered that several float-to-int64 conversions perform unchecked `static_cast<int64_t>` operations on unbounded floating-point values.

If extremely large floating-point values are passed (for example `1e20`), the cast exceeds signed 64-bit integer limits and triggers platform-dependent undefined behavior.

This currently affects:
- `clip_numeric`
- `coerce_value`

As a result, valid datasets can silently corrupt numeric values, overflow into negative ranges, or produce inconsistent behavior across architectures/compiler builds.

### Area

C++ cleaning engine

### Minimal reproduction

```python
import arnio as ar
import pandas as pd

df = pd.DataFrame({
    "value": [10, 20, 30]
})

frame = ar.from_pandas(df)

result = ar.pipeline(
    frame,
    [
        (
            "clip_numeric",
            {
                "lower": -1e20,
                "upper": 1e20
            }
        )
    ]
)

print(ar.to_pandas(result))
```

### Expected behavior

Numeric cleaning kernels should safely validate floating-point bounds before converting values into signed integer types.

Out-of-range values should:
- clamp safely
- raise validation errors
- or saturate deterministically

without invoking undefined behavior.

### Actual behavior

The native C++ cleaning kernels directly execute unchecked casts such as:

```cpp
static_cast<int64_t>(value)
```

on unbounded floating-point values.

When values exceed signed 64-bit integer limits:
- overflow occurs
- behavior becomes platform/compiler dependent
- negative sentinel values may appear
- numeric corruption becomes possible

### Impact

Produces incorrect data

### Arnio version

1.1.0

### Install method

Built from source

### Operating System

Linux

### Python version

3.12.4

### Environment details

```shell
Observed during deep validation of the C++ numeric cleaning engine.

Affected paths:
- `cpp/src/cleaning.cpp`
- `clip_numeric`
- `coerce_value`

Current logic includes unchecked conversions similar to:


static_cast<int64_t>(value)


without validating whether the floating-point value fits within:
- `INT64_MIN`
- `INT64_MAX`

This issue appears in multiple execution paths and may produce architecture-dependent behavior on x86_64 vs ARM systems.
```

### Pre-flight checks

### Additional context

This appears to be a numeric safety issue in the native C++ cleaning layer.

Potential safe fix direction:

Validate bounds using `std::numeric_limits<int64_t>` before any float-to-int conversion occurs.

Example approach:

```cpp
if (value > static_cast<double>(std::numeric_limits<int64_t>::max())) {
    value = static_cast<double>(std::numeric_limits<int64_t>::max());
}
```

This would:
- eliminate undefined behavior
- preserve deterministic numeric handling
- improve cross-platform safety
- prevent silent corruption in numeric cleaning operations

eshaanag:
Fixes #1591

This keeps the numeric-safety fix scoped to the two native paths called out in the issue: `coerce_value()` and the int64 branch of `clip_numeric()`.

The problem was that both paths validated part of the input shape, then still reached `static_cast<int64_t>` with doubles that could be outside the representable int64 range. That is the unsafe part: non-finite values, huge integral-looking doubles like `1e20`, and values at/above `2^63` should never reach the cast.

What changed:

- Added a small local `checked_double_to_int64()` helper in `cpp/src/cleaning.cpp`.
- Requires double values to be finite, integral, and inside the safe int64 conversion range before casting.
- Reused the helper from `coerce_value()` and from int64 `clip_numeric()` bounds.
- Preserved existing valid int64 fill behavior and ordinary valid clip bounds.
- Added native C++ regression coverage for out-of-range doubles, `inf`, `-inf`, `NaN`, and int64 boundary fill values.
- Added Python API regression coverage for out-of-range `fill_nulls()` and `clip_numeric()` inputs.

This PR is for GSSoC 2026 and stays focused on the assigned numeric-safety bug.

Validation:

- `cmake -S . -B /tmp/arnio-build-1591 -DARNIO_BUILD_CPP_TESTS=ON -Dpybind11_DIR=/home/kali/arnio/.venv-codex/lib/python3.13/site-packages/pybind11/share/cmake/pybind11`
- `cmake --build /tmp/arnio-build-1591`
- `ctest --test-dir /tmp/arnio-build-1591 --output-on-failure` — 4 passed
- `.venv-codex/bin/python -m pip install -e '.[dev]'`
- `.venv-codex/bin/python -m pytest tests/test_cleaning.py -q -k "clip_numeric or fill_nulls_validation_lossy"` — 23 passed
- `.venv-codex/bin/python -m pytest tests/test_cleaning.py -q` — 410 passed
- `.venv-codex/bin/python -m pytest -q` — 2466 passed, 43 skipped
- `.venv-codex/bin/python -m black --check tests/test_cleaning.py`
- `.venv-codex/bin/python -m ruff check tests/test_cleaning.py`
- `git diff --check`

### DIFF
--- a/cpp/src/cleaning.cpp
+++ b/cpp/src/cleaning.cpp
@@ -174,6 +174,18 @@ static uint64_t hash_row(const Frame& frame, size_t row, const std::vector<size_
     return h;
 }
 
+static int64_t checked_double_to_int64(double value, const char* context) {
+    constexpr double int64_min = static_cast<double>(std::numeric_limits<int64_t>::min());
+    constexpr double int64_max_exclusive = 9223372036854775808.0;  // 2^63
+
+    if (!std::isfinite(value) || value != std::floor(value) || value < int64_min ||
+        value >= int64_max_exclusive) {
+        throw std::invalid_argument(std::string(context) +
+                                    " must be a finite integral value within int64 range.");
+    }
+    return static_cast<int64_t>(value);
+}
+
 static CellValue coerce_value(const CellValue& value, DType target) {
     if (std::holds_alternative<std::monostate>(value)) {
         return std::monostate{};
@@ -190,12 +202,13 @@ static CellValue coerce_value(const CellValue& value, DType target) {
         }
         if (std::holds_alternative<double>(value)) {
             double d = std::get<double>(value);
-            if (std::isnan(d) || std::isinf(d) || d != std::floor(d)) {
+            try {
+                return checked_double_to_int64(d, "Numeric fill value");
+            } catch (const std::invalid_argument&) {
                 throw std::invalid_argument(
                     "Lossy or non-finite numeric fill values are not permitted for integer "
                     "columns.");
             }
-            return static_cast<int64_t>(d);
         }
         if (std::holds_alternative<std::string>(value)) {
             const auto& s = std::get<std::string>(value);
@@ -752,9 +765,9 @@ Frame clip_numeric(const Frame& frame, std::optional<double> lower, std::optiona
         if (src.dtype() == DType::INT64) {
             const auto& vec = std::get<std::vector<int64_t>>(src.data());
             Column col(src.name(), DType::INT64);
-            const int64_t lo = lower.has_value() ? static_cast<int64_t>(lower.value())
+            const int64_t lo = lower.has_value() ? checked_double_to_int64(lower.value(), "lower")
                                                  : std::numeric_limits<int64_t>::min();
-            const int64_t hi = upper.has_value() ? static_cast<int64_t>(upper.value())
+            const int64_t hi = upper.has_value() ? checked_double_to_int64(upper.value(), "upper")
                                                  : std::numeric_limits<int64_t>::max();
             for (size_t r = 0; r < src.size(); ++r) {
                 if (src.is_null(r)) {
--- a/cpp/tests/test_cleaning.cpp
+++ b/cpp/tests/test_cleaning.cpp
@@ -2,6 +2,7 @@
 #include <cmath>    // std::isnan
 #include <cstdint>  // uint64_t
 #include <cstring>  // std::memcpy
+#include <limits>
 
 #include "arnio/cleaning.h"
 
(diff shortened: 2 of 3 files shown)


im-anishraj (changes requested):
@eshaanag thank you for the focused numeric-safety patch. I added the GSSoC classification labels and reviewed the diff/CI. The implementation direction looks reasonable, but this PR is not merge-ready yet because the non-Vercel lint check is failing on clang-format.\n\nCurrent blocker:\n- cpp/tests/test_cleaning.cpp needs clang-format. CI reports formatting violations around the new Frame min_result, Frame max_result, and Frame result assignments in the numeric-safety tests. Please run clang-format-17 on the touched C++ test file or apply the repository formatter and push the formatted result.\n\nLocal checks I could run: git diff --check, Black, and Ruff all passed. The targeted Python pytest for the new native behavior fails in my local Windows checkout because it imports the already-installed _arnio_cpp extension from site-packages instead of rebuilding this PR's C++ changes; that is the same known local limitation, so I am not treating that as your blocker. Please fix clang-format and wait for the full non-Vercel CI matrix to go green, then I can re-review.

im-anishraj (approved):
@eshaanag thanks for the quick formatting fix. I re-reviewed the latest head (c88b35f2) and the clang-format blocker is resolved. The numeric-safety helper now guards finite/integral/range conditions before any int64 cast, and the new C++ plus Python regression tests cover out-of-range doubles, non-finite values, and boundary-safe cases.\n\nLocal checks: git diff --check, Black, and Ruff passed. The targeted Python pytest still fails in my Windows checkout because it imports the stale installed _arnio_cpp extension rather than this PR's rebuilt C++ code, but GitHub Actions rebuilt the native extension and all non-Vercel checks are green, including lint, native C++ tests, wheel smoke tests, and the full Python matrix. Approved.

PR #2006, opened 2026-05-29
fix: preserve row count in normalize_unicode for zero-column frames (…

### LINKED ISSUE #1956: Bug: normalize_unicode loses row count for zero-column frames
## Summary

`normalize_unicode()` loses row count when called on an `ArFrame` with zero columns and one or more rows.

For example, an input shape of `(3, 0)` becomes `(0, 0)`.

## Why it matters

Zero-column frames are valid in Arnio and can appear after column-selection or column-dropping operations. A text-normalization helper should not change the row count when there are no columns to process.

This is a data-integrity issue for pipelines that preserve rows while progressively reducing columns.

## Evidence

- `arnio/cleaning.py`: `normalize_unicode()` rebuilds the native frame with `_Frame.from_dict(new_columns, dtype_hints)`.
- When `new_columns` is empty, no row count is passed to `_Frame.from_dict()`.
- Other helpers already use `_Frame.from_dict({}, {}, frame.shape[0])` to preserve row count for zero-column results.

## Reproduction or trigger

```python
import pandas as pd
import arnio as ar

result = ar.normalize_unicode(frame)

assert frame.shape == (3, 0)
assert result.shape == (3, 0)  # currently fails; result is (0, 0)
```

## Expected behavior

`normalize_unicode()` should preserve the original row count when the input has zero columns.

## Actual behavior

The result loses the row count and reports `(0, 0)`.

## Suggested direction

Pass the source row count when constructing the replacement native frame, especially when `new_columns` is empty:

```python
new_cpp_frame = _Frame.from_dict(new_columns, dtype_hints, frame.shape[0])
```

Add focused tests for:

- zero-column frames with non-zero rows
- zero-column empty frames
- existing non-empty string-column behavior
- attrs preservation on the zero-column path

## References

- `arnio/cleaning.py`
- `tests/test_cleaning.py`
- Related broader issue: #941

CoderAnshuman:
Here is a concise and professional summary you can copy and paste directly into your Pull Request description on GitHub:

***

## Description

This PR resolves a data-integrity issue where `normalize_unicode()` loses the row count when called on an `ArFrame` with zero columns and one or more rows (for example, reducing a shape of `(3, 0)` down to `(0, 0)`). 

### Key Changes
- **Bug Fix (`arnio/cleaning.py`)**: Updated `normalize_unicode()` to explicitly pass the source frame's row count (`frame.shape[0]`) to `_Frame.from_dict()` when rebuilding the native C++ frame.
- **Testing (`tests/test_cleaning.py`)**: Added a dedicated test method `test_normalize_unicode_zero_columns` to thoroughly verify:
  - Non-empty zero-column frames (e.g., shape `(3, 0)`) preserve their row count.
  - Empty zero-column frames (shape `(0, 0)`) work correctly.
  - Existing non-empty string-column normalization behavior remains unchanged (no regressions).
  - Metadata `attrs` preservation and deep-copy isolation behave correctly on the zero-column path.

Closes #1956

### DIFF
--- a/arnio/cleaning.py
+++ b/arnio/cleaning.py
@@ -1112,7 +1112,7 @@ def normalize_unicode(
         else:
             new_columns[name] = col.to_python_list()
             dtype_hints[name] = dtype
-    new_cpp_frame = _Frame.from_dict(new_columns, dtype_hints)
+    new_cpp_frame = _Frame.from_dict(new_columns, dtype_hints, frame.shape[0])
     return ArFrame(
         new_cpp_frame,
         attrs=copy.deepcopy(frame._attrs) if frame._attrs is not None else None,
--- a/tests/test_cleaning.py
+++ b/tests/test_cleaning.py
@@ -1735,7 +1735,6 @@ class TestNormalizeUnicode:
         assert frame_3_0_attrs._attrs["key"] == "value"
 
 
-
 class TestParseBoolStrings:
     def test_parse_basic_bool_strings(self):
         import pandas as pd
--- a/tests/test_cleaning.py
+++ b/tests/test_cleaning.py
@@ -1699,6 +1699,42 @@ class TestNormalizeUnicode:
         result._attrs["meta"]["key"] = "mutated"
         assert frame._attrs["meta"]["key"] == "value"
 
+    def test_normalize_unicode_zero_columns(self):
+        import pandas as pd
+
+        import arnio as ar
+
+        # Non-empty zero-column frame
+        frame_3_0 = ar.from_pandas(pd.DataFrame(index=range(3)))
+        assert frame_3_0.shape == (3, 0)
+        result_3_0 = ar.normalize_unicode(frame_3_0)
+        assert result_3_0.shape == (3, 0)
+
+        # Empty zero-column frame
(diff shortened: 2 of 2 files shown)


im-anishraj (changes requested):
@CoderAnshuman thanks for the focused normalize_unicode fix. I reviewed the patch and ran the targeted behavior checks. The implementation direction looks good and the focused tests pass locally, but the PR cannot merge until formatting is fixed.

Local verification:
- `git diff --check` passed.
- Ruff passed for `arnio/cleaning.py tests/test_cleaning.py`.
- `pytest tests/test_cleaning.py -q -k "normalize_unicode"` passed (`13 passed`).
- Black failed: `python -m black --check arnio/cleaning.py tests/test_cleaning.py` would reformat `tests/test_cleaning.py`.

Please run Black on the touched files and push the formatting update, then wait for the non-Vercel CI checks to pass. After that this should be a quick re-review.

im-anishraj (approved):
@CoderAnshuman thanks for the quick cleanup. I re-ran the targeted checks locally: diff whitespace, Black, Ruff, and pytest tests/test_cleaning.py -q -k normalize_unicode all pass. The zero-column row-count fix is focused and the regression coverage is good, so this is ready to merge.

PR #2007, opened 2026-05-29
fix: reject zero-column Parquet export that drops row count

### LINKED ISSUE #1942: Bug: preserve row count when writing zero-column frames to Parquet
## Summary
`write_parquet()` silently loses row count when exporting an `ArFrame` that has rows but no columns.

## Why it matters
Zero-column frames can be produced by column-selection or drop-column workflows. If such a frame has a meaningful row count, exporting it to Parquet should either preserve that row count or raise a clear unsupported-case error. Writing a file that reads back as zero rows can hide data-shape bugs in downstream validation or pipeline smoke tests.

## Evidence
- `arnio/io.py`: `write_parquet()` converts the frame with `to_pandas(frame)` and then calls `DataFrame.to_parquet(...)`.
- `arnio/convert.py`: `to_pandas()` correctly creates `pd.DataFrame(index=pd.RangeIndex(cpp_frame.num_rows()))` when the frame has zero columns.
- With pandas/pyarrow, writing a zero-column `DataFrame` with a default `RangeIndex` reads back as an empty `(0, 0)` DataFrame, so the original row count is lost.

## Reproduction or trigger
```python
import pandas as pd
import arnio as ar

print(frame.shape)
# (3, 0)

ar.write_parquet(frame, "zero_columns.parquet")
roundtrip = pd.read_parquet("zero_columns.parquet")
print(roundtrip.shape)
# Actual today: (0, 0)
```

## Expected behavior
`write_parquet()` should either preserve the three-row shape when the file is read back, or explicitly reject non-empty zero-column frames with a clear `ValueError` explaining that this shape is not representable by the current Parquet export path.

## Actual behavior
The export succeeds, but reading the Parquet file back returns `(0, 0)`, silently dropping the original row count.

## Suggested direction
Add a focused guard in `write_parquet()` for zero-column frames with rows. Either:

- preserve row count using an explicit metadata/index strategy that round-trips through pyarrow/pandas, or
- reject the case clearly until row-count preservation is supported.

Add regression tests for:

- `(0, 0)` empty frames
- `(n, 0)` frames where `n > 0`
- normal non-empty frames to confirm existing behavior is unchanged

## References
- `arnio/io.py`
- `arnio/convert.py`
- Related but separate: #1648 tracks zero-column row-count preservation for `to_arrow()`.

shreyansh-tech21:
Fixes #1942

## Summary
- Add a guard in `write_parquet()` for `(n, 0)` frames where `n > 0`
- Keep `(0, 0)` export behavior unchanged
- Add regression tests for `(0, 0)`, `(n, 0)`, and normal frames

## Test plan
- [x] `pytest tests/test_parquet.py -v`
- [x] `make test`
- [x] `make lint`

### DIFF
--- a/arnio/io.py
+++ b/arnio/io.py
@@ -1456,8 +1456,6 @@ def write_parquet(
             f"Cannot write a zero-column ArFrame with {rows} rows to Parquet: the current export path cannot preserve row count without columns."
         )
 
-    """ If the frame has zero columns and a positive row count. """
-
     df = to_pandas(frame)
 
     kwargs: dict = {
--- a/arnio/io.py
+++ b/arnio/io.py
@@ -1450,6 +1450,14 @@ def write_parquet(
             "Install it with: pip install arnio[parquet]"
         ) from exc
 
+    rows, cols = frame.shape
+    if cols == 0 and rows > 0:
+        raise ValueError(
+            f"Cannot write a zero-column ArFrame with {rows} rows to Parquet: the current export path cannot preserve row count without columns."
+        )
+
+    """ If the frame has zero columns and a positive row count. """
+
     df = to_pandas(frame)
 
     kwargs: dict = {
--- a/tests/test_parquet.py
+++ b/tests/test_parquet.py
@@ -150,6 +150,46 @@ class TestWriteParquetCompression:
 
 
 @skip_without_pyarrow
+class TestWriteParquetZeroColumn:
+    def test_zero_by_zero_frame_round_trips_empty(self, tmp_path):
+        frame = ar.from_pandas(pd.DataFrame())
+        assert frame.shape == (0, 0)
+
+        out = tmp_path / "empty.parquet"
+
+        ar.write_parquet(frame, out)
+        df = pd.read_parquet(out, engine="pyarrow")
+
+        assert df.shape == (0, 0)
+        assert out.exists()
+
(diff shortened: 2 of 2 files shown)


im-anishraj (changes requested):
@shreyansh-tech21 thank you for the focused Parquet edge-case fix. I reviewed the patch and ran the local checks; the behavior and tests look good, but please make one small cleanup before merge.

Local verification passed:
- `git diff --check`
- `python -m black --check arnio/io.py tests/test_parquet.py`
- `python -m ruff check arnio/io.py tests/test_parquet.py`
- `pytest tests/test_parquet.py -q -k "ZeroColumn or write_parquet"` (`7 passed`)

Remaining cleanup:
- Please remove the standalone triple-quoted string in `arnio/io.py` after the new guard: `""" If the frame has zero columns and a positive row count. """`. It is a no-op runtime string, not a useful comment/docstring.

All non-Vercel CI checks are green, so after that cleanup this should be ready to merge.

im-anishraj (approved):
@shreyansh-tech21 thanks for removing the stray triple-quoted string. I re-ran the targeted local checks: diff whitespace, Black, Ruff, and `pytest tests/test_parquet.py -q -k "ZeroColumn or write_parquet"` all pass. The non-Vercel CI checks are green as well, so this is ready to merge.

PR #2008, opened 2026-05-29
fix: allow unregistering custom aliases of built-in cleaning steps

### LINKED ISSUE #1921: Bug: custom aliases of Arnio cleaning steps cannot be unregistered
## Summary
`unregister_step()` treats custom step names as built-ins when the registered function comes from `arnio.cleaning`, even if the custom name was user-defined.

## Why it matters
Users may reasonably register aliases or wrapped Arnio helpers while experimenting with pipelines. After registration, the custom step appears in `list_steps()`, but attempting to unregister it raises `UnknownStepError`. This makes the registry state confusing and can leave test suites or notebooks polluted until `reset_steps()` is called.

## Evidence
- `arnio/pipeline.py`: `_is_builtin_python_step()` returns true for any function whose module starts with `arnio.cleaning`.
- `arnio/pipeline.py`: `unregister_step()` blocks removal when `_is_builtin_python_step(name, fn)` returns true.
- The custom name can still be added to `_PYTHON_STEP_REGISTRY` through `register_step()` and appears in `list_steps()`.

## Reproduction or trigger
1. Run:
   ```python
   import arnio as ar

   ar.register_step("tmp_strip_alias", ar.normalize_whitespace)
   print("tmp_strip_alias" in ar.list_steps())
   ar.unregister_step("tmp_strip_alias")
   ```
2. Observe that the custom step is visible but cannot be unregistered.

## Expected behavior
A user-defined step name should be unregisterable unless the name itself belongs to a protected built-in step.

## Actual behavior
`unregister_step("tmp_strip_alias")` raises `UnknownStepError`, while the available-steps list still includes `tmp_strip_alias`.

## Suggested direction
Track protected built-in step names explicitly instead of inferring built-in status only from the registered function module. Add a regression test that registers an `arnio.cleaning` function under a custom name and verifies that `unregister_step()` removes it.

## References
- `arnio/pipeline.py`
- `tests/test_register_step_safety.py`

Arbaaz123676:
## Summary

Fixes #1921

Custom aliases registered using built-in `arnio.cleaning` functions were incorrectly treated as protected built-in steps during `unregister_step()`. As a result, user-defined aliases could not be removed even though they appeared in the step registry.

## Root Cause

Built-in step detection relied on the underlying function's module origin (`arnio.cleaning`) rather than the registered step name. This caused custom aliases pointing to built-in cleaning functions to be mistakenly identified as built-in steps.

## Changes

- Updated built-in step detection logic to distinguish actual built-in step names from user-defined aliases.
- Preserved protection for genuine built-in steps.
- Added regression tests covering:
  - Registration of a custom alias for a built-in cleaning function.
  - Successful unregistration of the alias.
  - Continued protection of actual built-in step names.

## Result

User-defined aliases of built-in cleaning steps can now be unregistered correctly, while built-in steps remain protected from removal.

### DIFF
--- a/arnio/pipeline.py
+++ b/arnio/pipeline.py
@@ -175,9 +175,13 @@ def unregister_step(name: str) -> None:
         if name not in _PYTHON_STEP_REGISTRY:
             available_steps = sorted(set(_STEP_REGISTRY) | set(_PYTHON_STEP_REGISTRY))
             raise UnknownStepError(name, available_steps)
-        fn = _PYTHON_STEP_REGISTRY[name]
 
-        if _is_builtin_python_step(name, fn):
+        # Protect only names that were registered as built-ins at module load
+        # time (i.e. present in _BUILTIN_PYTHON_STEP_REGISTRY).  Do NOT use
+        # _is_builtin_python_step here: that helper checks the function's
+        # __module__, which would wrongly block user-defined aliases whose
+        # underlying callable happens to live in arnio.cleaning.
+        if name in _BUILTIN_PYTHON_STEP_REGISTRY:
             available_steps = sorted(set(_STEP_REGISTRY) | set(_PYTHON_STEP_REGISTRY))
             raise UnknownStepError(name, available_steps)
         del _PYTHON_STEP_REGISTRY[name]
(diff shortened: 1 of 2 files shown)


im-anishraj (approved):
@Arbaaz123676 thanks for the focused fix. I reviewed the unregister logic and the regression tests: the guard now protects only actual built-in registered names, while allowing user-created aliases of built-in cleaning functions to be removed.

Local checks passed:
- `git diff --check`
- `python -m black --check arnio/pipeline.py tests/test_register_step_safety.py`
- `python -m ruff check arnio/pipeline.py tests/test_register_step_safety.py`
- `pytest tests/test_register_step_safety.py -q` (`8 passed`)

All non-Vercel CI checks are green; Vercel is only the known authorization gate. Approved.

PR #2017, opened 2026-05-29
fix: deterministic sort for mixed-type allowed values

### LINKED ISSUE #1682: Bug: serialize mixed-type schema allowed values deterministically
## Summary

`Schema.to_json()` can crash when a field's `allowed` set contains mixed scalar types, because the serializer tries to sort the set directly.

## Why it matters

`String(allowed=...)` is typed as accepting `Any` values, and mixed scalar allow-lists can be valid for messy CSV validation. Serialization should either preserve those schemas safely or reject them at construction time with a clear error. A raw sort comparison error makes schema export unreliable.

## Evidence

- `arnio/schema.py`: `_normalize_sequence()` converts sets with `sorted(value)`.
- Python cannot sort sets containing incomparable types such as `int` and `str`.
- `Schema.to_json()` calls `_field_to_json_dict()`, which eventually normalizes `field_def.allowed` through this helper.

## Reproduction or trigger

```python
import arnio as ar

schema = ar.Schema({
    "code": ar.String(allowed={1, "1"}),
})

schema.to_json()
```

## Expected behavior

Schema serialization should produce stable JSON for supported allowed values, or schema construction should reject mixed-type allowed values with a clear error.

## Actual behavior

`Schema.to_json()` raises a raw Python sorting error:

```text
TypeError: '<' not supported between instances of 'str' and 'int'
```

## Suggested direction

Use a JSON-safe deterministic ordering strategy that does not compare raw mixed types, for example sorting by `(type(value).__name__, repr(value))`, or explicitly validate allowed values to a narrower supported type set. Add tests for:

- homogeneous string allowed values
- homogeneous numeric allowed values
- mixed scalar allowed values
- round-tripping through `Schema.to_json()` / `Schema.from_json()`

## References

- `arnio/schema.py::_normalize_sequence`
- `arnio/schema.py::Schema.to_json`
- `tests/test_schema.py`

taniy8:
Fixes #1682

`_normalize_sequence()` called `sorted(value)` directly on sets, which raises `TypeError` when the set contains mixed scalar types like `{1, "1"}` since Python can't compare `int` and `str` with `<`.

Replaced with a type-safe key:

```python
sorted(value, key=lambda x: (type(x).__name__, repr(x)))
```

Tests added for homogeneous strings, homogeneous numerics, mixed scalars, and `to_json()` / `from_json()` round-trip.

### DIFF
--- a/arnio/schema.py
+++ b/arnio/schema.py
@@ -2665,11 +2665,15 @@ def _normalize_unique(
 def _normalize_sequence(value: Any) -> Any:
     """Convert sets and tuples to lists for JSON-serializable output.
 
-    Sets are sorted with a type-safe key so mixed scalar types (e.g. ``{1, "1"}``)
-    do not raise ``TypeError``.
+    Sets are sorted naturally when possible. If the set contains mixed
+    incomparable types (e.g. int and str), falls back to a type-safe key
+    so serialization never raises TypeError.
     """
     if isinstance(value, set):
-        return sorted(value, key=lambda x: (type(x).__name__, repr(x)))
+        try:
+            return sorted(value)
+        except TypeError:
+            return sorted(value, key=lambda x: (type(x).__name__, repr(x)))
 
     if isinstance(value, tuple):
         return list(value)
--- a/arnio/schema.py
+++ b/arnio/schema.py
@@ -2663,9 +2663,14 @@ def _normalize_unique(
 
 
 def _normalize_sequence(value: Any) -> Any:
-    """Convert sets and tuples to lists for JSON-serializable output."""
+    """Convert sets and tuples to lists for JSON-serializable output.
+
+    Sets are sorted with a type-safe key so mixed scalar types (e.g. ``{1, "1"}``)
+    do not raise ``TypeError``.
+    """
     if isinstance(value, set):
-        return sorted(value)
+        return sorted(value, key=lambda x: (type(x).__name__, repr(x)))
+
     if isinstance(value, tuple):
         return list(value)
     return value
--- a/tests/test_schema.py
+++ b/tests/test_schema.py
@@ -3586,3 +3586,37 @@ def test_validate_max_errors_zero_valid_data():
 
     with pytest.raises(ValueError, match="max_errors must be >= 1"):
         ar.validate(frame, schema, max_errors=0)
+
+
+def test_normalize_sequence_homogeneous_strings():
+    schema = ar.Schema({"status": ar.String(allowed={"active", "inactive", "pending"})})
+    payload = json.loads(schema.to_json())
+    assert payload["fields"]["status"]["allowed"] == ["active", "inactive", "pending"]
+
+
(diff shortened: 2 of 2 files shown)


im-anishraj:
@taniy8 thanks for the PR. It is linked to #1682 and the implementation scope matches the assigned deterministic allowed-values serialization fix.

I aligned the GSSoC tracking labels with the issue level: `gssoc:level-2` and `level:intermediate`. Please wait for the GitHub Actions checks to finish; the Vercel authorization failure is known external noise for fork PRs. I can do the full review once the Actions results are visible.

im-anishraj (changes requested):
@taniy8 thanks for the focused fix and the added schema tests. The mixed-type crash path is addressed, and the targeted local checks pass:

- `python -m black --check arnio/schema.py tests/test_schema.py`
- `python -m ruff check arnio/schema.py tests/test_schema.py`
- `pytest tests/test_schema.py -q -k "normalize_sequence or mixed_scalar_allowed or allowed"` -> 32 passed

One change is still needed before this is merge-ready. The new key is applied to every set, so homogeneous numeric allowed values no longer preserve the existing numeric sort behavior. For example, on this branch:

```python
json.loads(ar.Schema({"code": ar.Field(allowed={1, 2, 10})}).to_json())["fields"]["code"]["allowed"]
# [1, 10, 2]
```

Issue #1682 specifically asked to preserve existing homogeneous allowed-value behavior. Please update `_normalize_sequence()` to keep the normal `sorted(value)` path when the set is directly comparable, and only fall back to a type-safe key when Python raises `TypeError`, for example:

```python
try:
    return sorted(value)
except TypeError:
    return sorted(value, key=lambda x: (type(x).__name__, repr(x)))
```

Please also add/adjust a test such as `{1, 2, 10}` expecting `[1, 2, 10]`, and consider making the mixed round-trip assertion compare the restored schema/value set directly rather than only checking length.

taniy8:
Thanks for pointing that out! The `repr()`-based key breaks natural numeric ordering  `{1, 2, 10}` sorts as `[1, 10, 2]` instead of `[1, 2, 10]`. Switching to `try/except` keeps the natural sort for homogeneous sets and only falls back to the type-safe key when Python raises `TypeError`. Will update `_normalize_sequence()` and fix the tests accordingly.

im-anishraj (approved):
@taniy8 thanks for the quick update. This now preserves the homogeneous numeric ordering while still falling back safely for mixed scalar sets.

Verified locally on the updated branch:
- `python -m black --check arnio/schema.py tests/test_schema.py`
- `python -m ruff check arnio/schema.py tests/test_schema.py`
- `pytest tests/test_schema.py -q -k "normalize_sequence or mixed_scalar_allowed or allowed"` -> 32 passed
- manual checks for `{1, 2, 10}` serialization and `{1, "1"}` JSON round-trip

GitHub Actions is also green for the non-Vercel checks; the Vercel authorization failure is external repo noise. This is ready to merge.

PR #2021, opened 2026-05-29
fix: prevent benchmark scripts from deleting user files

### LINKED ISSUE #1933: Bug: benchmark scripts can overwrite or delete existing CSV files
## Summary
Some benchmark scripts can overwrite or delete CSV files that existed before the benchmark started.

## Why it matters
Benchmark scripts should never remove or overwrite user-created files unless the user explicitly opts into that behavior. This can cause accidental data loss for contributors running benchmarks from the repository root or passing a custom `--file` path.

## Evidence
- `benchmarks/benchmark_csv.py` sets `test_file = "large_benchmark.csv"`. `generate_test_csv()` skips generation when the file exists, but the `finally` block still removes `large_benchmark.csv` whenever it exists.
- Reproduced locally by creating a small `large_benchmark.csv`, running `py -3.13 benchmarks\benchmark_csv.py`, and observing that the pre-existing file was deleted.
- `benchmarks/benchmark_auto_clean_memory.py` accepts `--file`, but if the path already exists and `--reuse-file` is not passed, it overwrites the file, sets `generated = True`, and deletes it at the end unless `--keep-file` is passed.
- Reproduced locally by creating `auto_clean_sentinel.csv`, running `py -3.13 benchmarks\benchmark_auto_clean_memory.py --file auto_clean_sentinel.csv --rows 5 --repeat 1 --no-warmup`, and observing that the pre-existing file was overwritten and removed.

## Reproduction or trigger
1. From the repository root, create a file named `large_benchmark.csv`.
2. Run `python benchmarks/benchmark_csv.py`.
3. Observe that the original file is removed.
4. Or create `auto_clean_sentinel.csv` and run `python benchmarks/benchmark_auto_clean_memory.py --file auto_clean_sentinel.csv --rows 5 --repeat 1 --no-warmup`.
5. Observe that the file is overwritten and removed.

## Expected behavior
Benchmark scripts should only delete files created during the current run, and they should not overwrite user-provided paths without an explicit opt-in flag.

## Actual behavior
The scripts can remove or overwrite existing CSV files because cleanup/generation logic does not consistently track file ownership.

## Suggested direction
Use safer file ownership and temporary-file patterns:

- use `tempfile.NamedTemporaryFile` / `TemporaryDirectory` for generated benchmark data
- track whether the current run created each file before deleting it
- require an explicit `--overwrite` flag before overwriting a user-provided `--file`
- keep generated benchmark files under a dedicated ignored benchmark output directory

## References
- `benchmarks/benchmark_csv.py`
- `benchmarks/benchmark_auto_clean_memory.py`

nrynmish:
## Summary

Prevent benchmark scripts from overwriting or deleting pre-existing CSV files unless the current benchmark run explicitly owns or creates them.

Fixes #1933

## Changes

* Updated `benchmark_csv.py` to generate benchmark data in a temporary directory instead of using a fixed repository-root filename.
* Added explicit file ownership tracking in `benchmark_auto_clean_memory.py`.
* Added a new `--overwrite` flag for intentionally replacing an existing user-provided file.
* Prevented cleanup logic from deleting files that existed before the benchmark started.
* Preserved existing `--reuse-file` and `--keep-file` behavior.

## Verification

Verified the following scenarios manually:

### benchmark_csv.py

* Confirmed a pre-existing `large_benchmark.csv` is no longer deleted after running the benchmark.
* Confirmed benchmark data is generated and used from a temporary directory.

### benchmark_auto_clean_memory.py

* Confirmed a pre-existing file without `--reuse-file` or `--overwrite` raises `FileExistsError`.
* Confirmed a pre-existing file with `--reuse-file` is reused and preserved.
* Confirmed a pre-existing file with `--overwrite` is explicitly replaced.
* Confirmed cleanup only removes files created during the current benchmark run.

## Quality Checks

* `black` passed
* `ruff` passed
* `pre-commit` checks passed

### DIFF
--- a/benchmarks/benchmark_auto_clean_memory.py
+++ b/benchmarks/benchmark_auto_clean_memory.py
@@ -259,6 +259,12 @@ def main() -> None:
         help="Keep the generated CSV file after benchmark",
     )
 
+    parser.add_argument(
+        "--overwrite",
+        action="store_true",
+        help="Overwrite an existing --file path",
+    )
+
     args = parser.parse_args()
 
     csv_path = (
@@ -269,13 +275,21 @@ def main() -> None:
 
     warmup = not args.no_warmup
 
+    csv_existed_before = os.path.exists(csv_path)
     generated = False
 
-    if os.path.exists(csv_path) and args.reuse_file:
+    if csv_existed_before and args.reuse_file:
         print(f"Using existing dataset: {csv_path}")
 
+    elif csv_existed_before and not args.overwrite:
+        raise FileExistsError(
+            f"{csv_path} already exists. "
+            "Use --reuse-file to use it as-is or "
+            "--overwrite to replace it."
+        )
+
     else:
-        if os.path.exists(csv_path):
+        if csv_existed_before:
             print(f"Overwriting existing dataset: {csv_path}")
 
         generate_synthetic_data(
--- a/benchmarks/benchmark_csv.py
+++ b/benchmarks/benchmark_csv.py
@@ -1,14 +1,16 @@
 import csv
-import os
+import tempfile
 import time
+from pathlib import Path
 
 import arnio
 
 
 def generate_test_csv(filename, num_rows=200000, num_cols=10):
-    if not os.path.exists(filename):
+    filename = Path(filename)
+    if not filename.exists():
         print(f"Generating test CSV with {num_rows} rows...")
-        with open(filename, "w", newline="", encoding="utf-8") as f:
+        with filename.open("w", newline="", encoding="utf-8") as f:
             writer = csv.writer(f)
             header = [f"col_{i}" for i in range(num_cols)]
             writer.writerow(header)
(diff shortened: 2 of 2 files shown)


im-anishraj (approved):
@nrynmish reviewed and this is good to merge.

Verified locally:
- `python -m black --check benchmarks/benchmark_csv.py benchmarks/benchmark_auto_clean_memory.py`
- `python -m ruff check benchmarks/benchmark_csv.py benchmarks/benchmark_auto_clean_memory.py`
- safety smoke checks for #1933:
  - pre-existing `--file` without `--reuse-file`/`--overwrite` exits with `FileExistsError` and preserves the file;
  - pre-existing `--file` with `--reuse-file --keep-file` succeeds and preserves the file;
  - `benchmark_csv.py` runs from a directory containing a pre-existing `large_benchmark.csv` and preserves that file.

GitHub Actions is green for non-Vercel checks. The Vercel authorization failure is external repo noise.

PR #2026, opened 2026-05-29
fix: return fresh object for zero-row drop_constant_columns

### LINKED ISSUE #1949: Bug: avoid returning original object from drop_constant_columns empty-input path
## Summary

`drop_constant_columns()` returns the original input object when the frame has zero rows.

For pandas inputs this means the returned `DataFrame` is the same mutable object. For `ArFrame` inputs this means the same `ArFrame` and `_attrs` object are returned instead of a fresh result.

## Why it matters

Most cleaning helpers are expected to behave like transformations that return a new frame. Returning the original object on one edge path makes mutation/copy semantics inconsistent and can surprise users who treat cleaning outputs as independent results.

This is especially important because `drop_constant_columns()` documents the return value as a “New frame without constant columns.”

## Evidence

- `arnio/cleaning.py`: `drop_constant_columns()` converts the input, then immediately returns `frame` when `len(df.index) == 0`.
- `arnio/cleaning.py`: the docstring says the function returns a new frame.
- `tests/test_cleaning.py`: zero-row behavior is covered for shape/columns, but object identity and metadata aliasing are not tested.

## Reproduction or trigger

```python
import pandas as pd
import arnio as ar

df = pd.DataFrame({"a": pd.Series(dtype="int64")})
result = ar.drop_constant_columns(df)
result["b"] = []

assert list(df.columns) == ["a"]  # currently fails; df now also has "b"

frame = ar.from_pandas(pd.DataFrame({"a": pd.Series(dtype="int64")}))
frame._attrs = {"nested": {"x": 1}}

result = ar.drop_constant_columns(frame)
result._attrs["nested"]["x"] = 2

assert frame._attrs["nested"]["x"] == 1  # currently fails
```

## Expected behavior

Zero-row inputs should still return a fresh result object with the same columns and row count, without aliasing the original pandas object or mutable `ArFrame` metadata.

## Actual behavior

The zero-row branch returns the original object directly.

## Suggested direction

Use a zero-row copy for pandas inputs, and for `ArFrame` inputs return a fresh `ArFrame` converted from that copy while preserving the existing zero-row shape behavior. Add focused tests for:

- pandas zero-row input does not return/alias the original object
- `ArFrame` zero-row input returns a new wrapper
- mutable `_attrs` are not shared between input and output
- existing zero-row columns and shape remain unchanged

## References

- `arnio/cleaning.py`
- `tests/test_cleaning.py`

Code-007-maker:
Fixes #1949

### Changes

- Return a fresh result object for zero-row pandas inputs
- Return a fresh ArFrame wrapper for zero-row ArFrame inputs
- Deep copy `_attrs` metadata to prevent aliasing
- Added regression tests for:
  - pandas object identity
  - ArFrame object identity
  - `_attrs` metadata isolation

### DIFF
--- a/arnio/cleaning.py
+++ b/arnio/cleaning.py
@@ -507,7 +507,17 @@ def drop_constant_columns(
     frame, is_arframe = _validate_frame(frame, allow_pandas=True)
     df = to_pandas(frame) if is_arframe else frame
     if len(df.index) == 0:
-        return frame
+        result_df = df.copy(deep=True)
+
+        if is_arframe:
+            result = from_pandas(result_df)
+
+            if getattr(frame, "_attrs", None) is not None:
+                result._attrs = copy.deepcopy(frame._attrs)
+
+            return result
+
+        return result_df
 
     nunique_counts = df.nunique(dropna=False)
     constant_columns = nunique_counts[nunique_counts == 1].index.tolist()
--- a/tests/test_cleaning.py
+++ b/tests/test_cleaning.py
@@ -888,22 +888,18 @@ class TestDropConstantColumns:
         result = ar.drop_constant_columns(df)
 
         assert result is not df
-        assert result.shape == (0, 1)        
+        assert result.shape == (0, 1)
 
     def test_drop_constant_columns_zero_row_arframe_returns_new_object(self):
-        frame = ar.from_pandas(
-            pd.DataFrame({"a": pd.Series(dtype="int64")})
-        )
+        frame = ar.from_pandas(pd.DataFrame({"a": pd.Series(dtype="int64")}))
 
         result = ar.drop_constant_columns(frame)
 
         assert result is not frame
-        assert result.shape == (0, 1)    
+        assert result.shape == (0, 1)
 
     def test_drop_constant_columns_zero_row_attrs_not_shared(self):
-        frame = ar.from_pandas(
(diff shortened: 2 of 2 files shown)


im-anishraj:
@Code-007-maker thanks for opening the PR. I added the missing GSSoC tracking labels so this matches #1949.

Current blocker: `Lint (Black, Ruff, clang-format)` is failing at the `Black (check-only)` step on the latest head. GitHub will publish full logs after the run finishes, but the immediate fix should be to run Black on the touched Python files, especially `arnio/cleaning.py` and `tests/test_cleaning.py`, then push the formatted result. The Vercel authorization status can be ignored for this repo.

im-anishraj (approved):
@Code-007-maker reviewed after the formatting follow-up, and this is good to merge.

Verified locally:
- `python -m black --check arnio/cleaning.py tests/test_cleaning.py`
- `python -m ruff check arnio/cleaning.py tests/test_cleaning.py`
- `pytest tests/test_cleaning.py -q -k drop_constant_columns` -> 13 passed

The PR is focused on #1949, preserves the zero-row column shape, returns a fresh object, and deep-copies ArFrame attrs for the no-op path. GitHub Actions is green for non-Vercel checks; the Vercel authorization failure is external repo noise.

PR #2033, opened 2026-05-29
fix: ArFrame string output is blank for zero-column frames

### LINKED ISSUE #1939: Bug: ArFrame string output is blank for zero-column frames
## Summary

`ArFrame.__str__()` renders blank table lines for frames that have rows but zero columns.

This is separate from the existing `preview()` issue: `__str__()` has its own formatting path and currently produces a confusing multi-line output with empty header, separator, and data rows.

## Why it matters

Zero-column frames can occur after column selection, dropping all constant/empty columns, or constructing from a pandas DataFrame with an index but no columns. `str(frame)` is the default CLI/debugging representation, so it should clearly describe this valid state instead of printing blank rows.

## Evidence

- `arnio/frame.py`: `ArFrame.__str__()` only special-cases `rows == 0`.
- For `rows > 0` and `cols == 0`, `col_header`, `separator`, and each row render as empty strings.

## Reproduction or trigger

```python
import pandas as pd
import arnio as ar

frame = ar.from_pandas(pd.DataFrame(index=range(2)))
print(frame.shape)      # (2, 0)
print(str(frame))
```

Current output includes blank table lines:

```text
ArFrame: 2 rows × 0 columns
Columns: []
DTypes: {}
Memory: 104 bytes



```

## Expected behavior

`str(frame)` should explicitly report the state, for example:

```text
ArFrame: 2 rows × 0 columns
Columns: []
DTypes: {}
Memory: ... bytes
(no columns to display)
```

## Actual behavior

The representation prints empty header/separator/data lines, which looks like broken output.

## Suggested direction

Add a `cols == 0` branch in `ArFrame.__str__()` before building table rows. Add a regression test for a non-empty zero-column frame.

## References

- `arnio/frame.py`: `ArFrame.__str__()`
- Related but separate display path: #1401 covers `ArFrame.preview()`

Aditya8369:
Implemented the fix and regression test.

### Changes made
- Added a zero-column fast path in frame.py so non-empty frames with 0 columns no longer render blank table lines.
- The string output now includes an explicit terminal line, shown at frame.py: (no columns to display).
- Added regression test test_frame.py for a frame with 2 rows and 0 columns, asserting the shape line, columns, dtypes, and explicit no-columns message.

### Validation
- Ran targeted pytest for string tests in test_frame.
- Test run could not proceed because collection fails before tests execute: arnio._arnio_cpp is missing in this environment, so importing arnio fails.
- Error summary: ModuleNotFoundError for arnio._arnio_cpp, then ImportError from _core.py during test collection.

closes #1939

### DIFF
--- a/arnio/frame.py
+++ b/arnio/frame.py
@@ -711,6 +711,14 @@ class ArFrame:
         if rows == 0:
             return f"{header}\nColumns: {truncated_names}\n(empty frame)"
 
+        if cols == 0:
+            dtypes_line = f"DTypes: {self.dtypes}"
+            memory_line = f"Memory: {self.memory_usage()} bytes"
+            return (
+                f"{header}\nColumns: {truncated_names}\n{dtypes_line}\n"
+                f"{memory_line}\n(no columns to display)"
+            )
+
         actual_n = min(5, rows)
         col_data = [
             [self._frame.column_by_index(i).at(r) for r in range(actual_n)]
--- a/tests/test_frame.py
+++ b/tests/test_frame.py
@@ -723,6 +723,17 @@ def test_str_keeps_normal_column_names():
     assert "..." not in result
 
 
+def test_str_zero_columns_non_empty_rows_has_explicit_message():
+    frame = ar.from_pandas(pd.DataFrame(index=range(2)))
+
+    result = str(frame)
+
+    assert "ArFrame: 2 rows × 0 columns" in result
+    assert "Columns: []" in result
+    assert "DTypes: {}" in result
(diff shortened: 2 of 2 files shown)


im-anishraj:
@Aditya8369 thank you for the focused fix. I verified the diff, labels, and CI: Black/Ruff passed, the targeted string representation tests passed locally, the manual zero-column reproduction now shows (no columns to display), and all non-Vercel GitHub checks are green. The Vercel status is the known external authorization-limit failure, so I merged this PR and added the GSSoC approval/quality labels.

PR #2036, opened 2026-05-29
fix: handle malformed benchmark entries safely

### LINKED ISSUE #1908: Bug: benchmark summary crashes on malformed timing entries
## Summary

`benchmarks/generate_pr_summary.py` crashes when benchmark result or baseline entries are incomplete, or when the baseline execution time is zero.

## Why it matters

This script is intended to generate PR/release benchmark summaries. If a benchmark run writes a partial JSON file, a baseline is edited incorrectly, or an older baseline contains `arnio_exec_time: 0`, the summary generator crashes instead of producing a clear “not comparable” row. That makes benchmark automation fragile during release validation.

## Evidence

- `benchmarks/generate_pr_summary.py`: `generate_summary()` indexes `current["arnio_exec_time"]` and `baseline_case["arnio_exec_time"]` directly.
- The same function divides by `baseline_time` without checking for zero.
- `tests/test_generate_pr_summary.py` covers missing whole-baseline data and normal faster/slower cases, but not partial/malformed entries.

## Reproduction or trigger

```python
from benchmarks.generate_pr_summary import generate_summary

# Crashes with ZeroDivisionError
generate_summary(
    {"Case A": {"arnio_exec_time": 1.0}},
    {"Case A": {"arnio_exec_time": 0.0}},
)

# Crashes with KeyError
generate_summary(
    {"Case A": {}},
    {"Case A": {"arnio_exec_time": 1.0}},
)
```

## Expected behavior

The summary generator should not crash on malformed benchmark entries. It should mark that row as not comparable and continue generating the rest of the Markdown summary.

## Actual behavior

Malformed entries raise raw `ZeroDivisionError` or `KeyError`, aborting the whole summary.

## Suggested direction

Harden `generate_summary()` by validating each case before calculating the percent change:

- require numeric `arnio_exec_time` in both current and baseline case
- treat missing or non-positive baseline times as “No comparable baseline available” for that row
- add tests for zero baseline time, missing current metric, missing baseline metric, and mixed valid/invalid cases

## References

- `benchmarks/generate_pr_summary.py`
- `tests/test_generate_pr_summary.py`
- `benchmarks/baseline.json`

Ankitkr16:
## Description

This PR fixes benchmark tooling crashes caused by malformed benchmark entries and invalid baseline timings.

### Changes Made

* Added validation for missing benchmark metrics in `generate_pr_summary.py`
* Prevented crashes caused by missing `arnio_exec_time` fields
* Handled zero and non-positive baseline execution times safely
* Updated benchmark regression helpers in `benchmark_vs_pandas.py`
* Added test coverage for:

  * zero baseline timings
  * missing current metrics
  * missing baseline metrics
  * mixed valid and invalid benchmark entries

## Related Issue

Closes #1908

## Testing

* Added and executed benchmark summary tests
* Verified malformed benchmark data no longer causes crashes
* Confirmed valid benchmark comparisons continue to work correctly

### Test Result


im-anishraj (changes requested):
@Ankitkr16 thanks for the benchmark hardening PR. I reviewed the linked issue and ran local checks. The focused summary tests pass (`pytest tests/test_generate_pr_summary.py -q` -> 7 passed), and Ruff passes, but this is not merge-ready yet:

- `python -m black --check benchmarks/benchmark_vs_pandas.py benchmarks/generate_pr_summary.py tests/test_generate_pr_summary.py` reports all three touched files would be reformatted.
- Please also re-check the `run_case()` control flow in `benchmarks/benchmark_vs_pandas.py`: after the new invalid-baseline branch, the old regression calculation block still appears to run again. The invalid-baseline path should return/skip cleanly and must not leave `is_regression` or `regression_percent` in an unsafe state.

Please run Black on the touched files and add/adjust a focused test for the benchmark regression helper path with non-positive or missing baseline timing, not only `generate_summary()`. Also, #1908 was not assigned yet, so please wait for maintainer assignment/confirmation before treating it as GSSoC-approved work.

Ankitkr16:
Thanks for the review.

I addressed the formatting issues by running Black on the touched files and cleaned up the `run_case()` control flow so invalid/non-positive baselines no longer fall through into duplicate regression checks.

Please take another look when you have a chance.

im-anishraj:
@Ankitkr16 thanks for the update. I saw the new push and your note, but the latest CI run is still not ready for re-review: the `Lint (Black, Ruff, clang-format)` job failed again on the updated commit.

GitHub has not exposed the full job log yet while the workflow run is still finishing, so please open that lint job once logs are available and fix whatever Black/Ruff reports. I will do the code-level re-review after the lint job is green; the remaining Vercel status is still just the external authorization check.

im-anishraj (changes requested):
@Ankitkr16 thanks for the update. The formatting issue is fixed now, and I verified locally that `black --check`, `ruff check`, and `pytest tests/test_generate_pr_summary.py -q` all pass on the latest head. However, this still cannot be merged because the latest diff removes the executable script behavior from `benchmarks/generate_pr_summary.py`.

`.github/workflows/benchmark.yml` runs `python benchmarks/generate_pr_summary.py` and then reads `benchmark_summary.md`, but this PR deletes the `load_json()` helper, file constants, `main()`, and the `if __name__ == "__main__"` entry point. That means the workflow can run the script successfully while producing no `benchmark_summary.md`, breaking benchmark PR summaries.

Please keep the malformed-entry handling, but restore the CLI/script behavior: load `benchmark_results.json` and `benchmarks/baseline.json`, write `benchmark_summary.md`, and print the summary as before. A small regression test or smoke check for the script entry point/output file would be ideal. Once that is restored, I can re-review quickly.

Ankitkr16:
Hi @im-anishraj,

I restored the benchmark summary script behavior (`load_json`, CLI entry point, summary file generation, and output printing) while keeping the malformed benchmark handling improvements intact.

I also addressed the formatting issues and re-ran the relevant tests. All CI checks are now passing, with only the external Vercel authorization status remaining.

Please take another look when you have time. Thank you for the review and guidance.

im-anishraj (approved):
@Ankitkr16 thanks for the follow-up. I re-reviewed the latest diff and verified that the malformed benchmark handling now preserves the benchmark summary script behavior, including the CLI entry point and benchmark_summary.md output path.

Local checks on the verified head passed for the touched files:
- black --check benchmarks/benchmark_vs_pandas.py benchmarks/generate_pr_summary.py tests/test_generate_pr_summary.py
- ruff check benchmarks/benchmark_vs_pandas.py benchmarks/generate_pr_summary.py tests/test_generate_pr_summary.py
- python -m pytest tests/test_generate_pr_summary.py -q

All non-Vercel CI checks are green, and the remaining Vercel status is the known external authorization check. Approved.

PR #2079, opened 2026-05-29
test: add comprehensive tests for select_columns (#1814)

### LINKED ISSUE #1814: test : add tests for frame select_columns
## Summary
Add tests for arnio.frame.select_columns function
## Verification
pytest tests/test_frame.py -v -k select_columns

Nicks-19:
Resolves #1814

### Description

This change adds tests for the select_columns cleaning function in two places: `tests/test_cleaning.py` and `tests/test_frame.py`.

Here are the new tests we have:

1. **Handling values:** We make sure that `select_columns` keeps None, np.nan and pd.NA in the columns we choose.

2. **Testing columns:** We check that we can select one column from a frame with columns and from a frame with only one column.

3. **Changing column order:** We verify that when we select columns in an order the values are still correct and the new column order is kept.

4. **Using types:** We confirm that if we pass the types of containers like a dictionary or a single number we get a TypeError.

5. **Keeping data types:** We check that the types of data in the columns, like Int64, float64, boolean and string are the same after we select them.

All added tests were verified and are fully passing.

### DIFF
--- a/tests/test_cleaning.py
+++ b/tests/test_cleaning.py
@@ -3940,6 +3940,91 @@ class TestSelectColumns:
         with pytest.raises(ValueError):
             ar.select_columns(frame, ["id", "id"])
 
+    def test_select_columns_null_nan_handling(self):
+        df = pd.DataFrame(
+            {
+                "id": [1, None, 3],
+                "name": ["Alice", "Bob", None],
+                "score": [95.5, float("nan"), 80.0],
+            }
+        )
+        frame = ar.from_pandas(df)
+        selected = ar.select_columns(frame, ["id", "score"])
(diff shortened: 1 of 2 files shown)


im-anishraj:
@Nicks-19 I synced the GSSoC labels from #1814 onto this PR. GitHub Actions are green; the Vercel authorization failure is external and can be ignored here. This is now in the maintainer review queue.

im-anishraj (approved):
Thanks @Nicks-19, this adds useful coverage for both the functional and method-style select_columns paths. I reviewed the added cases and ran the targeted select_columns tests locally on Python 3.12; they pass.

PR #2085, opened 2026-05-29
Fix: short-circuit read_csv when nrows=0 to avoid unnecessary I/O

### LINKED ISSUE #2070: Bug: read_csv does not short-circuit when nrows=0, unlike read_jsonl which returns an empty frame without any I/O
**What happened?**

`read_jsonl` in `arnio/io.py` explicitly handles `nrows=0` as a short-circuit that returns an empty ArFrame without any file I/O:

```python
if nrows == 0:
    return from_pandas(pd.DataFrame())
```

`read_csv` has no equivalent handling. `_validate_nrows` only rejects negative values, passing `nrows=0` to the C++ reader. The C++ reader's behavior with `nrows=0` is unspecified.

**Area**

**Minimal reproduction**

```python
import arnio as ar
import tempfile, os

with tempfile.NamedTemporaryFile('w', suffix='.csv', delete=False) as f:
    f.write('not,valid\x00csv')
    path = f.name

try:
    result = ar.read_csv(path, nrows=0)  # behavior undefined
    # ar.read_jsonl(path, nrows=0) -> empty ArFrame, no error
finally:
    os.unlink(path)
```

Expected: `read_csv(path, nrows=0)` returns an empty ArFrame without reading the file.
Actual: behavior depends on C++ backend; inconsistent with `read_jsonl`.

**Environment**
Python 3.x, any OS

Xenon010101:
### Description
When 
rows=0 is passed to 
ead_csv, the function should return an empty DataFrame immediately without performing any file I/O. This matches the existing behavior of 
ead_jsonl, which already has a 
rows=0 short-circuit. Without this fix, 
ead_csv still materializes the input, validates the path, sets up the C++ reader config, and opens the file — all for zero rows.

### Related Issue
Fixes #2070

### Type of Change
- [x] Bug fix (non-breaking change which fixes an issue)

### Checklist
- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my code
- [x] My changes generate no new warnings

program: gssoc

### DIFF
--- a/arnio/io.py
+++ b/arnio/io.py
@@ -570,6 +570,13 @@ def read_csv(
     """
     path, should_cleanup, is_materialized_text = _materialize_csv_input(path)
 
+    if nrows is not None and nrows == 0:
+        if should_cleanup and os.path.exists(path):
+            os.unlink(path)
+        from .integrations.pandas import from_pandas
+        import pandas as pd
+        return from_pandas(pd.DataFrame())
+
     try:
         _validate_csv_path(path, encoding)
 


im-anishraj:
@Xenon010101 I added the matching GSSoC tracking labels, but this PR is not ready for maintainer review as a GSSoC assignment yet because the linked issue is still `status:needs-triage` and has not been assigned to you.

Please wait for maintainer triage/assignment before stacking implementation PRs. If the issue becomes `status:ready`, ask on the issue with a focused plan and confirm you have no active assigned work.

im-anishraj (changes requested):
@Xenon010101 thanks for the PR. I cannot merge this as submitted.

Blockers:
- The linked issue #2070 is not assigned and was still in maintainer triage, so this PR was opened before the issue was ready/assigned.
- The PR has no regression test for `nrows=0` behavior.
- The current implementation short-circuits after `_materialize_csv_input()` but before validating the path/input. For a missing file path, `read_csv(..., nrows=0)` would return an empty frame instead of surfacing the missing file, which changes error behavior.
- The returned empty frame also does not consider header/usecols/dtype expectations, so the intended API behavior needs maintainer confirmation first.

I moved the issue/PR to `status:needs-maintainer`. Please wait for the issue scope to be confirmed and assigned before continuing, then add focused tests that cover the agreed behavior.

Xenon010101:
Hi maintainer! This PR has been open for a while — just a friendly nudge for review when you get a chance. Thanks!

im-anishraj:
@Xenon010101 thanks for checking in. This PR is still blocked on the same items from the review: issue #2070 needs maintainer scoping/assignment first, and the implementation still needs focused regression coverage plus confirmation of the expected 
rows=0 error/header behavior. Please wait for the issue to be marked ready/assigned, then update the PR against that confirmed scope.

im-anishraj:
@Xenon010101, thank you for investigating the `nrows=0` path. I am closing this PR because issue #2070 was not assigned or scoped before implementation, the PR has no regression tests, and the current short-circuit changes established error behavior: a missing file can return an empty frame, and header/usecols/dtype expectations are not resolved.

If issue #2070 is later approved and marked `status:ready`, please request assignment first. A new implementation should follow the agreed API behavior and include focused tests for missing paths, headers, usecols, dtype, file-like inputs, and empty files.

PR #2092, opened 2026-05-30
fix: make custom pipeline step frames writable

### LINKED ISSUE #1657: Bug: custom pipeline steps receive read-only zero-copy buffers
## Summary
Custom Python pipeline steps receive a pandas DataFrame created with the default zero-copy `to_pandas()` path. Numeric columns can expose read-only NumPy buffers, so custom steps that mutate arrays in place fail with `ValueError: assignment destination is read-only`.

## Why it matters
`register_step()` is Arnio’s contributor/user extension point. Many pandas users write custom cleaning steps that mutate a DataFrame in place before returning it. Even if Arnio prefers functional steps, the current behavior is surprising because the custom step receives a pandas DataFrame that looks normal but contains read-only buffers for some columns.

## Evidence
- `arnio/pipeline.py`: custom Python steps use `df = to_pandas(result)` without `copy=True`.
- `arnio/convert.py`: `to_pandas(copy=False)` preserves zero-copy paths where supported.
- Runtime behavior on current source:

```python
import pandas as pd
import arnio as ar
from arnio.pipeline import register_step


def mutate_numpy(df):
    arr = df["x"].to_numpy()
    arr[0] = 99.0
    return df

register_step("mutate_numpy_probe", mutate_numpy, overwrite=True)
frame = ar.from_pandas(pd.DataFrame({"x": [1.0, 2.0]}))
ar.pipeline(frame, [("mutate_numpy_probe",)])
# PipelineStepError wrapping: ValueError: assignment destination is read-only
```

## Reproduction or trigger
1. Register a custom Python pipeline step that mutates a numeric column through `to_numpy()` or similar in-place operations.
2. Run `ar.pipeline(frame, [("step_name",)])` on a numeric frame.
3. Observe the read-only buffer error.

## Expected behavior
Custom pipeline steps should either receive a writable pandas DataFrame or the documentation should clearly state that custom steps must be functional/non-mutating and should create copies before in-place mutation.

## Actual behavior
The DataFrame appears like a normal pandas object, but some underlying arrays are read-only because they came from Arnio’s zero-copy conversion path.

## Suggested direction
Choose and document one contract:

- safer API: use `to_pandas(result, copy=True)` for custom Python pipeline steps
- performance-first API: document that custom steps may receive read-only zero-copy buffers and should avoid in-place mutation
- optional API: add a pipeline option such as `custom_step_copy=True` if maintainers want both modes

Add regression tests covering a custom step that mutates a numeric array in place.

## References
- `arnio/pipeline.py`
- `arnio/convert.py`
- `tests/test_pipeline_custom_step_return_type.py`

eshaanag:
Fixes #1657

This PR keeps the custom pipeline step fix scoped to the handoff between Arnio and user-registered Python steps.

What changed:

- Custom user steps now receive a defensive pandas frame instead of the default zero-copy view.
- The writable frame path is only used for genuine custom steps; built-in Python-backed Arnio steps keep the existing handoff.
- The custom-step frame uses a small internal pandas subclass so `Series.to_numpy()` returns a writable view even under pandas 3 copy-on-write behavior.
- Added a regression test for a custom step that mutates a numeric column through `df["x"].to_numpy()[0] = ...`.

Why this shape:

I first tried the obvious `to_pandas(..., copy=True)` path, but under pandas 3 the resulting `Series.to_numpy()` view is still marked read-only because copy-on-write is always active. The internal wrapper keeps the public custom-step contract the same while making the frame users receive behave like a normal mutable working DataFrame.

Validation:

- `.venv-codex/bin/python -m pytest tests/test_pipeline_custom_step_return_type.py -q` -> 19 passed
- `.venv-codex/bin/python -m pytest tests/test_pipeline.py tests/test_pipeline_custom_step_return_type.py -q` -> 145 passed
- `.venv-codex/bin/ruff check arnio/pipeline.py tests/test_pipeline_custom_step_return_type.py` -> passed
- `.venv-codex/bin/python -m black --check --diff arnio/pipeline.py tests/test_pipeline_custom_step_return_type.py` -> passed
- `.venv-codex/bin/python -m pytest -q` -> 2712 passed, 3 skipped
- `git diff --check` -> passed

This is for GSSoC 2026 and stays limited to the assigned custom pipeline step writable-buffer bug.

### DIFF
--- a/arnio/pipeline.py
+++ b/arnio/pipeline.py
@@ -5,6 +5,7 @@ Chained cleaning pipeline.
 
 from __future__ import annotations
 
+import copy as copylib
 import inspect
 import logging
 import warnings
@@ -68,6 +69,41 @@ class PipelineContext:
     step_index: int
     total_steps: int
     dry_run: bool
+
+
+class _WritablePipelineSeries(pd.Series):
+    @property
+    def _constructor(self):
+        return _WritablePipelineSeries
+
+    @property
+    def _constructor_expanddim(self):
+        return _WritablePipelineDataFrame
+
+    def to_numpy(self, *args, **kwargs):
+        values = super().to_numpy(*args, **kwargs)
+        try:
+            values.setflags(write=True)
+        except ValueError:
+            pass
+        return values
+
+
+class _WritablePipelineDataFrame(pd.DataFrame):
+    @property
+    def _constructor(self):
+        return _WritablePipelineDataFrame
+
+    @property
+    def _constructor_sliced(self):
+        return _WritablePipelineSeries
+
+
+def _to_writable_pipeline_dataframe(frame: ArFrame) -> pd.DataFrame:
+    base = to_pandas(frame, copy=True)
+    writable = _WritablePipelineDataFrame(base, copy=False)
+    writable.attrs = copylib.deepcopy(base.attrs)
+    return writable
 
 
 def _is_builtin_python_step(name: str, fn: Callable) -> bool:
@@ -473,10 +509,12 @@ def pipeline(
 
             fn = python_step_registry[name]
 
-            df = to_pandas(working_frame)
-
             # Isolate genuine custom steps from internal core library functions
             is_builtin = _is_builtin_python_step(name, fn)
+            if is_builtin:
+                df = to_pandas(working_frame)
+            else:
+                df = _to_writable_pipeline_dataframe(working_frame)
             signature = inspect.signature(fn)
             call_kwargs = dict(kwargs)
             if "context" in signature.parameters and "context" not in call_kwargs:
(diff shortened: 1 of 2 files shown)


im-anishraj (approved):
@eshaanag this is ready. The fix is scoped to the assigned custom-step writable-buffer issue, preserves built-in step handling, and adds focused regression coverage for in-place NumPy mutation inside a custom step.

Local checks run:
- git diff --check origin/main...HEAD -> passed
- black --check arnio/pipeline.py tests/test_pipeline_custom_step_return_type.py -> passed
- ruff check arnio/pipeline.py tests/test_pipeline_custom_step_return_type.py -> passed
- pytest tests/test_pipeline_custom_step_return_type.py -q -> 19 passed
- pytest tests/test_pipeline.py -q -> 126 passed

The GitHub matrix is green; the remaining Vercel status is only the external authorization gate.

PR #2124, opened 2026-05-30
fix: clarify HTML repr for non-empty zero-column frames

### LINKED ISSUE #1650: Bug: clarify HTML repr for non-empty zero-column frames
## Summary
`ArFrame._repr_html_()` renders non-empty zero-column frames as `(empty)`, even though the frame still has a positive row count.

## Why it matters
Zero-column frames can happen after legitimate cleaning operations such as dropping all-empty columns. In notebooks, the HTML representation is the primary user-facing display. Showing `(empty)` makes it look like rows were removed, even when the C++ frame still reports preserved row count.

## Evidence
- `arnio/frame.py`: `_repr_html_()` returns `summary + "<p><em>(empty)</em></p>"` when `num_cols == 0 or num_rows == 0`.
- `arnio/cleaning.py`: `drop_empty_columns()` can return `ArFrame(_Frame.from_dict({}, {}, frame.shape[0]), ...)`, preserving row count with zero columns.
- Runtime check on current source plus the compiled extension shows `zero.shape == (2, 0)` but `_repr_html_()` renders `(empty)`.

## Reproduction or trigger
1. Create a frame with rows where every column is empty/null.
2. Run `ar.drop_empty_columns(frame)`.
3. Display the result in a notebook, or call `result._repr_html_()` directly.

```python
import pandas as pd
import arnio as ar

frame = ar.from_pandas(pd.DataFrame({"a": [None, None]}))
zero = ar.drop_empty_columns(frame)
print(zero.shape)       # (2, 0)
print(zero._repr_html_())
```

## Expected behavior
The HTML output should distinguish zero rows from zero columns, for example by saying the frame has `2 rows × 0 cols` and no displayable columns.

## Actual behavior
The summary preserves the shape, but the body renders `<em>(empty)</em>`, which implies there are no rows.

## Suggested direction
Split the empty-frame branch into two cases:

- `num_rows == 0`: render `(empty frame)`.
- `num_cols == 0 and num_rows > 0`: render a clear message such as `(no columns to display; row count preserved)`.

Add a focused test for `_repr_html_()` on a zero-column frame with positive row count.

## References
- `arnio/frame.py`
- `arnio/cleaning.py`

Aditya8369:
Updated frame.py so `_repr_html_()` now treats zero rows and zero columns separately: zero rows still render “(empty)”, while non-empty zero-column frames render “(no columns to display)”. I also added a regression test in test_frame.py that builds a zero-column frame via `drop_empty_columns()` and checks that the row count is preserved in the HTML repr.

Validation: `python -m py_compile frame.py tests/test_frame.py` passed. 

closes #1650

### DIFF
--- a/arnio/frame.py
+++ b/arnio/frame.py
@@ -987,8 +987,10 @@ class ArFrame:
         )
 
         # ── empty-frame fast path ─────────────────────────────────────────
-        if num_cols == 0 or num_rows == 0:
+        if num_rows == 0:
             return summary + "<p><em>(empty)</em></p>"
+        if num_cols == 0:
+            return summary + "<p><em>(no columns to display)</em></p>"
 
         # ── column header ─────────────────────────────────────────────────
         th_style = (
--- a/tests/test_frame.py
+++ b/tests/test_frame.py
@@ -1356,6 +1356,18 @@ def test_repr_html_empty_frame_no_crash(tmp_path):
     assert len(out) > 0
 
 
+def test_repr_html_zero_columns_preserves_row_count():
+    frame = ar.from_pandas(pd.DataFrame({"a": [None, None]}))
+    zero = ar.drop_empty_columns(frame)
+
+    out = zero._repr_html_()
+
+    assert zero.shape == (2, 0)
+    assert "ArFrame [2 rows × 0 cols]" in out
+    assert "(no columns to display)" in out
+    assert "(empty)" not in out
+
+
 def test_repr_html_with_nulls_no_crash(csv_with_nulls):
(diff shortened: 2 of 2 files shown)


im-anishraj (approved):
Hi @Aditya8369, thanks for the focused fix. I reviewed the diff and verified that zero-row frames and non-empty zero-column frames now render distinct HTML messages while preserving the row-count summary. Local checks passed for git diff --check, Black, Ruff, and the focused 	ests/test_frame.py -k repr_html slice with 16 passing tests. The full GitHub Actions suite is passing apart from the known Vercel authorization status, so this is ready to merge.

PR #2125, opened 2026-05-30
fix: raise ValueError for empty columns in ArFrame.from_records with …

### LINKED ISSUE #1795: Bug: reject empty columns in ArFrame.from_records
## Summary
`ArFrame.from_records()` silently drops all rows when dict records are passed with `columns=[]`.

## Why it matters
`from_records()` is a public constructor. Passing an explicit empty `columns` list should either be rejected clearly or preserve the requested row count as an empty-column frame. Returning `(0, 0)` from non-empty input silently loses data and can hide caller mistakes.

## Evidence
- `arnio/frame.py::ArFrame.from_records()` accepts `columns` as a list/tuple of strings but does not reject an empty list for dict records.
- It delegates to `pandas.DataFrame.from_records(records, columns=columns)` and then `from_pandas()`.
- With non-empty dict records and `columns=[]`, pandas returns an empty `(0, 0)` frame, so Arnio loses the source row count.

## Reproduction or trigger
```python
import arnio as ar

frame = ar.ArFrame.from_records([{"a": 1}, {"a": 2}], columns=[])
print(frame.shape)
```

## Expected behavior
Either:

- reject `columns=[]` with a clear `ValueError`, or
- return an empty-column frame that still preserves the two source rows, if Arnio wants to support that shape.

## Actual behavior
The constructor returns an empty frame with no rows:

```text
(0, 0)
```

## Suggested direction
Reject empty `columns` for dict records unless empty-column frames with preserved row counts are fully supported here. Add tests for dict records with `columns=[]` and for normal explicit column subsets.

## References
- `arnio/frame.py::ArFrame.from_records`
- `tests/test_frame.py`

taniy8:
Fixes #1795

## Summary

`ArFrame.from_records()` silently returned an empty `(0, 0)` frame when dict records were passed with `columns=[]`. This caused silent data loss with no indication to the caller.

## Changes

**`arnio/frame.py`**
- Added a `ValueError` guard in the dict branch of `ArFrame.from_records()` that rejects `columns=[]` with a clear, actionable message pointing the caller to use `columns=None` instead.

**`tests/test_frame.py`**
- `test_from_records_dict_empty_columns_raises` —> verifies `columns=[]` raises `ValueError`
- `test_from_records_dict_empty_columns_message_is_helpful` —> verifies the error message mentions `columns=None`
- `test_from_records_dict_columns_none_infers_from_keys` —> verifies default behavior (infer from keys) is unchanged
- `test_from_records_dict_explicit_valid_columns_subset` —> verifies a valid non-empty explicit column list still works correctly

## Behavior

**Before**
```python
frame = ar.ArFrame.from_records([{"a": 1}, {"a": 2}], columns=[])
print(frame.shape)  # (0, 0) — silent data loss
```

**After**
```python
frame = ar.ArFrame.from_records([{"a": 1}, {"a": 2}], columns=[])
# ValueError: columns must not be empty when records are dicts;
# pass columns=None to infer column names from the record keys
```

## Notes

- Scoped strictly to the dict records path in `from_records()` as requested
- Existing behavior for `columns=None` and valid explicit column subsets is fully preserved
- 8 pre-existing test failures in `test_describe_*` are unrelated to this change and were present on `main` before this branch

### DIFF
--- a/arnio/frame.py
+++ b/arnio/frame.py
@@ -155,6 +155,11 @@ class ArFrame:
                             f"nested values are not supported; "
                             f"column {col!r} at row {i} contains a {type(val).__name__!r}"
                         )
+            if columns is not None and len(columns) == 0:
+                raise ValueError(
+                    "columns must not be empty when records are dicts; "
+                    "pass columns=None to infer column names from the record keys"
+                )
             df = pd.DataFrame.from_records(records, columns=columns)
 
         elif isinstance(first, (list, tuple)):
--- a/tests/test_frame.py
+++ b/tests/test_frame.py
@@ -1683,3 +1683,33 @@ def test_shallow_copy_independent_frame():
     assert shallow._frame is not original._frame
     ar.rename_columns(original, {"a": "b"})
     assert "a" in shallow.columns  # mutation must not leak
+
+
+# ── from_records empty-columns guard ─────────────────────────────────────────
+
+
+def test_from_records_dict_empty_columns_raises():
+    """columns=[] with dict records must raise ValueError, not silently lose rows."""
+    with pytest.raises(ValueError, match="columns must not be empty"):
+        ar.ArFrame.from_records([{"a": 1}, {"a": 2}], columns=[])
+
+
+def test_from_records_dict_empty_columns_message_is_helpful():
+    """The error message should mention passing columns=None."""
+    with pytest.raises(ValueError, match="columns=None"):
+        ar.ArFrame.from_records([{"a": 1}], columns=[])
+
+
+def test_from_records_dict_columns_none_infers_from_keys():
+    """columns=None (default) should infer column names from dict keys."""
+    frame = ar.ArFrame.from_records([{"a": 1}, {"a": 2}])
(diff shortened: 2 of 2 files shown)


im-anishraj:
Hi @taniy8, thank you for the focused fix. I reviewed the diff and also ran local targeted checks: `git diff --check`, Black, Ruff, and `pytest tests/test_frame.py -k from_records_dict -q` all passed locally. The implementation matches the assigned scope for #1795.

I am not merging yet because the full GitHub Actions test matrix is currently red. The representative failure is `tests/test_website_api_docs.py::test_website_api_reference_mentions_public_exports`, which reports `CleaningSuggestion` is exported but missing from `website/api.html`. This looks unrelated to your touched `ArFrame.from_records()` files, but I need the non-Vercel CI to be green before merging. Please hold this PR for now; once that inherited docs/export mismatch is resolved on the base branch, this should be ready for final approval and merge.

taniy8:
Hi @im-anishraj, I have resolved the merge conflicts and the CI is green. Could you please re-review when you get a chance?

PR #2148, opened 2026-05-30
fix:DataQualityReport.to_pandas

### LINKED ISSUE #1668: Bug: DataQualityReport.to_pandas should return stable columns for empty reports
## Summary

`DataQualityReport.to_pandas()` returns a DataFrame with no columns when the report has no profiled columns. This makes the export schema unstable for empty or fully excluded reports.

## Why it matters

Quality reports are often consumed by notebooks, CI summaries, dashboards, or tests. Even when there are no columns to report, the DataFrame export should have a predictable set of column names so downstream code can select/report fields safely.

## Evidence

- `arnio/quality.py::DataQualityReport.to_pandas()` builds `pd.DataFrame([...])` directly from `self.columns.values()`.
- When `self.columns` is empty, pandas receives an empty list and creates a `(0, 0)` DataFrame.
- `ValidationResult.to_pandas()` already handles the analogous case by returning stable issue columns for passing results.

## Reproduction or trigger

```python
import pandas as pd
import arnio as ar

frame = ar.from_pandas(pd.DataFrame(index=range(2)))
report = ar.profile(frame)

print(report.to_pandas().columns.tolist())
print(report.to_pandas().shape)
```

Current output:

```python
[]
(0, 0)
```

## Expected behavior

`to_pandas()` should return an empty DataFrame with the documented report columns, such as `name`, `dtype`, `semantic_type`, `null_count`, `unique_count`, `warnings`, and related metric fields.

## Actual behavior

It returns a DataFrame with no columns.

## Suggested direction

Define a stable list of quality-report export columns and pass it to `pd.DataFrame(..., columns=...)` when there are no column profiles.

Add tests for:

- zero-column reports
- reports where all columns are excluded
- normal reports preserving existing columns and values

## References

- `arnio/quality.py::DataQualityReport.to_pandas`
- `arnio/schema.py::ValidationResult.to_pandas`

Aditya8369:
DataQualityReport.to_pandas now explicitly returns an empty DataFrame with the declared quality-report schema when there are no profiled columns, while keeping the existing populated-row export path unchanged. The change is in quality.py.

Regression coverage now includes a direct empty report, a fully excluded report, and a normal populated report that preserves both column order and values in test_quality.py and test_quality.py.

closes #1668

### DIFF
--- a/arnio/quality.py
+++ b/arnio/quality.py
@@ -271,6 +271,41 @@ class ColumnProfile:
         }
 
 
+QUALITY_REPORT_COLUMNS = [
+    "name",
+    "dtype",
+    "semantic_type",
+    "null_count",
+    "null_ratio",
+    "unique_count",
+    "unique_ratio",
+    "empty_string_count",
+    "whitespace_count",
+    "suggested_dtype",
+    "email_validity_ratio",
+    "url_validity_ratio",
+    "min",
+    "max",
+    "mean",
+    "std",
+    "warnings",
+    "top_values",
+    "top_values_is_approximate",
+    "top_values_sample_count",
+    "top_values_sample_ratio",
+    "histogram",
+    "q25",
+    "q50",
+    "q75",
+    "q95",
+    "iqr",
+    "outlier_lower_bound",
+    "outlier_upper_bound",
+    "outlier_count",
+    "outlier_ratio",
+]
+
+
 @dataclass(frozen=True)
 class DataQualityReport:
     """Whole-frame data quality report."""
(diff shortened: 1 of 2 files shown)


im-anishraj (approved):
Thanks @Aditya8369. I reviewed the fresh #1668 PR and synced the GSSoC/tracking labels from the assigned issue.

The diff is scoped to `DataQualityReport.to_pandas()` export stability and focused tests. The stable column list matches the existing export shape, empty/all-excluded reports now keep the schema, normal reports preserve the existing row values, and the non-Vercel checks are green. The remaining Vercel authorization status is external.

Approved from the scoped quality export review side.

PR #2177, opened 2026-05-30
fix: preserve row count for zero-column frames in drop_duplicates

### LINKED ISSUE #1957: Bug: drop_duplicates collapses zero-column frames
## Summary

`drop_duplicates()` silently changes the row count for zero-column `ArFrame` objects. A frame with shape `(3, 0)` becomes `(1, 0)` with `keep="first"` or `keep="last"`, and `(0, 0)` with `keep="none"` / `False`.

## Why it matters

Arnio now supports zero-column frames in several places. If a cleaning step silently collapses rows just because there are no comparison columns, downstream pandas, schema, or pipeline code can receive a different row count from the input without any clear signal.

## Evidence

- `arnio/cleaning.py`: `drop_duplicates(frame, subset=None)` documents that `None` compares all columns and forwards `subset=None` to the native implementation.
- `cpp/src/cleaning.cpp`: explicit empty subsets are rejected only when `subset.has_value() && subset->empty()`. For a zero-column frame with `subset=None`, `resolve_subset(frame, subset)` returns an empty column-index list and every row hashes/serializes as the same key.
- Closed issue #718 fixed `subset=[]`, but this case is triggered by `subset=None` on a valid zero-column frame.

## Reproduction or trigger

```python
import pandas as pd
import arnio as ar

print(frame.shape)
# (3, 0)

print(ar.drop_duplicates(frame).shape)
# Actual: (1, 0)

print(ar.drop_duplicates(frame, keep="none").shape)
# Actual: (0, 0)

print(pd.DataFrame(index=range(3)).drop_duplicates().shape)
# Pandas reference: (3, 0)
```

## Expected behavior

`drop_duplicates()` should either preserve the row count for zero-column frames, matching pandas behavior, or raise a clear `ValueError` explaining that duplicate detection requires at least one column.

## Actual behavior

The operation silently treats all rows as duplicates because there are no columns in the row key.

## Suggested direction

Add an explicit guard for zero-column frames before calling the native duplicate-removal path. If Arnio chooses pandas parity, add a zero-column fast path that returns a new frame preserving the original row count. If Arnio chooses rejection, raise a clear error from the Python wrapper and/or the C++ core.

## References

- Related but distinct: #718 fixed `drop_duplicates(subset=[])`
- Relevant files: `arnio/cleaning.py`, `cpp/src/cleaning.cpp`, `tests/test_cleaning.py`

SHRESTHA-012:
Summary
'drop_duplicates()' silently collapsed all rows in zero-column ArFrame objects. 
When `subset=None` is passed on a zero-column frame, 'resolve_subset' returns 
an empty column-index list, causing every row to hash identically and be treated 
as a duplicate. This fix adds an early return in both the Python wrapper and the 
C++ core to mirror pandas behaviour and preserve the original row count.

Fixes #1957

Type of change
- [x] Bug fix
- [x] Feature
- [x] Performance improvement
- [x] Documentation
- [x] Tests
- [x] Refactor
- [x] CI / packaging

 Testing
Added 4 new tests in cpp/tests/test_cleaning.cpp covering:
- keep=first preserves row count for zero-column frame
- keep=last preserves row count for zero-column frame  
- keep=none preserves row count for zero-column frame
- zero-row zero-column frame stays empty

Python layer guard added in arnio/cleaning.py.
C++ belt-and-suspenders guard added in cpp/src/cleaning.cpp.

### DIFF
--- a/arnio/cleaning.py
+++ b/arnio/cleaning.py
@@ -468,6 +468,9 @@ def drop_duplicates(
     >>> unique = ar.drop_duplicates(frame, subset=["name"], keep="first")
     """
     frame, _ = _validate_frame(frame)
+    if frame.shape[1] == 0:
+        return from_pandas(to_pandas(frame))
+
     if subset is not None:
         subset = _validate_column_sequence(subset, argument_name="subset")
         if len(subset) == 0:
--- a/arnio/cleaning.py
+++ b/arnio/cleaning.py
@@ -490,7 +490,9 @@ def drop_duplicates(
     if keep_arg not in {"first", "last", "none"}:
         raise ValueError("keep must be one of 'first', 'last', 'none', or False")
     if frame.shape[1] == 0:
-        return from_pandas(to_pandas(frame))
+        import copy
+
+        return copy.deepcopy(frame)
     result = _drop_duplicates(frame._frame, subset=subset, keep=keep_arg)
     return ArFrame(result)
 
--- a/arnio/cleaning.py
+++ b/arnio/cleaning.py
@@ -468,8 +468,6 @@ def drop_duplicates(
     >>> unique = ar.drop_duplicates(frame, subset=["name"], keep="first")
     """
     frame, _ = _validate_frame(frame)
-    if frame.shape[1] == 0:
-        return from_pandas(to_pandas(frame))
 
     if subset is not None:
         subset = _validate_column_sequence(subset, argument_name="subset")
@@ -491,6 +489,8 @@ def drop_duplicates(
     keep_arg = "none" if keep is False else keep
     if keep_arg not in {"first", "last", "none"}:
         raise ValueError("keep must be one of 'first', 'last', 'none', or False")
+    if frame.shape[1] == 0:
+        return from_pandas(to_pandas(frame))
     result = _drop_duplicates(frame._frame, subset=subset, keep=keep_arg)
     return ArFrame(result)
 
--- a/arnio/cleaning.py
+++ b/arnio/cleaning.py
@@ -490,9 +490,9 @@ def drop_duplicates(
     if keep_arg not in {"first", "last", "none"}:
         raise ValueError("keep must be one of 'first', 'last', 'none', or False")
     if frame.shape[1] == 0:
-        import copy
+        from ._core import _Frame
 
-        return copy.deepcopy(frame)
+        return ArFrame(_Frame.from_dict({}, {}, frame.shape[0]))
     result = _drop_duplicates(frame._frame, subset=subset, keep=keep_arg)
     return ArFrame(result)
 
--- a/cpp/src/cleaning.cpp
+++ b/cpp/src/cleaning.cpp
@@ -304,6 +304,9 @@ Frame drop_duplicates(const Frame& frame, const std::optional<std::vector<std::s
     if (subset.has_value() && subset->empty()) {
         throw std::invalid_argument("drop_duplicates subset cannot be empty");
     }
+    if (frame.num_cols() == 0) {
+    return frame.clone();
+    }
 
     auto col_indices = resolve_subset(frame, subset);
     RowContext ctx{&frame, &col_indices};
--- a/cpp/src/cleaning.cpp
+++ b/cpp/src/cleaning.cpp
@@ -305,7 +305,7 @@ Frame drop_duplicates(const Frame& frame, const std::optional<std::vector<std::s
         throw std::invalid_argument("drop_duplicates subset cannot be empty");
     }
     if (frame.num_cols() == 0) {
-    return frame.clone();
+        return frame.clone();
     }
 
     auto col_indices = resolve_subset(frame, subset);
--- a/cpp/src/cleaning.cpp
+++ b/cpp/src/cleaning.cpp
@@ -304,11 +304,14 @@ Frame drop_duplicates(const Frame& frame, const std::optional<std::vector<std::s
     if (subset.has_value() && subset->empty()) {
         throw std::invalid_argument("drop_duplicates subset cannot be empty");
     }
+
+    auto col_indices = resolve_subset(frame, subset);
+    if (keep != "first" && keep != "last" && keep != "none") {
+        throw std::invalid_argument("keep must be 'first', 'last', or 'none'");
+    }
     if (frame.num_cols() == 0) {
         return frame.clone();
     }
-
-    auto col_indices = resolve_subset(frame, subset);
     RowContext ctx{&frame, &col_indices};
     RowHash hash(ctx);
     RowEqual eq(ctx);
--- a/cpp/tests/test_cleaning.cpp
+++ b/cpp/tests/test_cleaning.cpp
@@ -470,8 +470,8 @@ TEST_CASE("drop_duplicates zero-col zero-row frame stays empty", "[cleaning][ded
 TEST_CASE("drop_duplicates zero-col missing subset column still throws",
           "[cleaning][dedup][zero-col]") {
     Frame f = make_zero_col_frame(3);
-    REQUIRE_THROWS_WITH(drop_duplicates(f, std::vector<std::string>{"missing"}, "first"),
-                        "Column not found: missing");
+    REQUIRE_THROWS_AS(drop_duplicates(f, std::vector<std::string>{"missing"}, "first"),
+                      std::out_of_range);
 }
 
 TEST_CASE("drop_duplicates zero-col invalid keep still throws", "[cleaning][dedup][zero-col]") {
--- a/cpp/tests/test_cleaning.cpp
+++ b/cpp/tests/test_cleaning.cpp
@@ -431,3 +431,39 @@ TEST_CASE("drop_duplicates NaN payloads treated as equal (keep=none)", "[cleanin
     REQUIRE(result.num_rows() == 1);
     REQUIRE(result.column("v").at(0) == CellValue(double(3.0)));
 }
+
+static Frame make_zero_col_frame(size_t num_rows) {
+    return Frame(num_rows);
+}
+
+TEST_CASE("drop_duplicates zero-col frame keep=first preserves row count",
+          "[cleaning][dedup][zero-col]") {
+    Frame f = make_zero_col_frame(3);
+    Frame result = drop_duplicates(f, std::nullopt, "first");
+    REQUIRE(result.num_rows() == 3);
+    REQUIRE(result.num_cols() == 0);
+}
+
+TEST_CASE("drop_duplicates zero-col frame keep=last preserves row count",
+          "[cleaning][dedup][zero-col]") {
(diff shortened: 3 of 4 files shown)


im-anishraj (changes requested):
@SHRESTHA-012 thanks for taking on #1957. I reviewed the full diff, conversation, labels, and current CI status. The overall direction makes sense, and I added the missing GSSoC classification labels for this PR.

Before this can be merged, please address these blockers:

- The lint/format check is failing. In `cpp/src/cleaning.cpp`, the new early return block needs normal indentation/formatting so the CI lint job passes.
- Please keep `subset` validation intact for zero-column frames. Right now the Python guard returns before validating `subset`, so calls such as `drop_duplicates(frame, subset=[])` or `drop_duplicates(frame, subset=["missing"])` would no longer raise the expected validation errors. The C++ guard also returns before `resolve_subset`, so a non-empty invalid subset can be silently accepted there too. Please validate `subset` first, then only preserve the zero-column row count when `subset=None` or after a valid subset path is confirmed.
- Please add/adjust Python regression coverage as well, since this behavior is exposed through `arnio.cleaning.drop_duplicates`, not only the C++ core.

I did not run local targeted runtime tests because the PR currently fails CI lint and it touches C++ core code that requires the native build toolchain. Once the validation order and formatting are fixed and CI is green, I can re-review it.

SHRESTHA-012:
@im-anishraj Please check once more , CI tests are passing this time.

im-anishraj (changes requested):
@SHRESTHA-012 thanks for fixing the formatting/CI issue. I rechecked the latest head and confirmed the Python formatting checks are clean:

```powershell
python -m black --check arnio\cleaning.py
python -m ruff check arnio\cleaning.py
```

However, the validation blocker is still present. On a zero-column frame, the Python guard returns before `subset` validation runs:

```text
subset=[]: returned shape (3, 0)
subset=['missing']: returned shape (3, 0)
```

Those calls should not silently succeed. Please move the zero-column fast path after the existing `subset` validation / missing-column validation, and add Python regression tests for the public `arnio.cleaning.drop_duplicates` API covering:

- `subset=None` preserves row count for a non-empty zero-column frame.
- `subset=[]` still raises the existing empty-subset validation error.
- `subset=["missing"]` still raises the expected missing-column error.

Because this still changes C++ core behavior too, final merge also needs local native targeted testing or an available C++ build toolchain. I did not merge this PR.

SHRESTHA-012:
@im-anishraj Made the changes. Please review.

im-anishraj (changes requested):
@SHRESTHA-012 thanks for the update. I rechecked the latest head and the Python-facing validation path is now much better. Local focused checks passed:

```powershell
python -m black --check arnio\cleaning.py tests\test_cleaning.py
python -m ruff check arnio\cleaning.py tests\test_cleaning.py
python -m pytest tests/test_cleaning.py -q -k drop_duplicates_zero_col
```

Result: `3 passed` for the new Python zero-column tests.

One C++ core blocker remains before this can merge: `cpp/src/cleaning.cpp` still returns `frame.clone()` before `resolve_subset(frame, subset)` and before the later `keep` validation. That means the native `drop_duplicates()` path can silently accept invalid inputs on zero-column frames, for example `subset={"missing"}` or an invalid `keep` value, instead of preserving the existing validation behavior.

Please move the C++ zero-column fast path after subset resolution and `keep` validation, and add C++ regression coverage for the invalid native paths, at least:

- zero-column frame with `subset={"missing"}` still throws the expected missing-column error
- zero-column frame with invalid `keep` still throws the expected invalid-keep error
- `subset=nullopt` still preserves the zero-column row count

Because this changes C++ core behavior, final approval/merge still needs native targeted testing. I could not complete native local testing on this Windows machine because the available compiler toolchain cannot build this C++17 extension.

SHRESTHA-012:
@im-anishraj I will make the following changes and revert back soon.

im-anishraj:
@SHRESTHA-012 thanks for the latest update. I checked the new C++ CI failure from head `27166a5`. The production validation order now reaches the missing-column path, but the new C++ regression test is asserting the wrong exception type:

```text
cpp/tests/test_cleaning.cpp:473
REQUIRE_THROWS_AS(..., std::invalid_argument)
due to unexpected exception with message: Column not found: missing
```

Please align the test with the existing native missing-column exception behavior instead of changing production behavior only to satisfy the test. In practice, that means asserting the existing exception type or matching the expected `Column not found: missing` message with the correct Catch2 assertion. Keep the zero-column fast path after `subset` resolution and `keep` validation.

I did not merge this PR because C++ Native Tests are still failing on all native platforms.

im-anishraj:
@SHRESTHA-012 thanks for the new commit. I checked the current head `9382d1f` and CI is still blocked in the C++ Native Tests on all native platforms.

The remaining focused failure is still in `cpp/tests/test_cleaning.cpp:473`:

```text
REQUIRE_THROWS_AS(..., std::runtime_error)
due to unexpected exception with message:
  Column not found: missing
```

That means the test is still asserting the wrong exception type. Please either assert the actual native exception type, or switch this regression to a message-based Catch2 assertion for `Column not found: missing`, for example with `REQUIRE_THROWS_WITH(..., Catch::Matchers::ContainsSubstring("Column not found: missing"))`. No production behavior change is needed for this specific failure.

Vercel can still be ignored; the merge blocker is the failing C++ native matrix.

im-anishraj:
@SHRESTHA-012 thanks for pushing the follow-up quickly. The new head (`e5ff743`) now fails at C++ compile time, before the test can run.

The problem is in `cpp/tests/test_cleaning.cpp:473-474`: the code still uses `REQUIRE_THROWS_AS(...)`, but the second argument is now the string literal `"Column not found: missing"`. `REQUIRE_THROWS_AS` expects an exception type there, so all native platforms fail to compile.

If you want to assert the error message, switch the macro itself, for example:

```cpp
REQUIRE_THROWS_WITH(
    drop_duplicates(f, std::vector<std::string>{"missing"}, "first"),
    "Column not found: missing");
```

Use `REQUIRE_THROWS_AS` only if the second argument is the actual exception type. The failing CodeQL C++ job is also caused by this same native build error; Vercel can still be ignored.

im-anishraj:
@SHRESTHA-012 thanks for the new push. The current head (`31d2371`) is still blocked by the CI lint job. It now gets past the previous macro-type problem, but clang-format is failing on the new assertion:

```text
cpp/tests/test_cleaning.cpp:473:90: error: code should be clang-formatted
REQUIRE_THROWS_WITH(drop_duplicates(f, std::vector<std::string>{"missing"}, "first"), ...)
```

Please run clang-format on `cpp/tests/test_cleaning.cpp` or split that assertion into the normal formatted multiline style, then push again. Vercel can be ignored; lint is the blocker before the native matrix result matters.

im-anishraj:
@SHRESTHA-012 thanks, the formatting blocker is fixed on the new head (`22c99d9`), but the C++ native matrix is still failing at compile time.

Representative Ubuntu failure:

```text
cpp/tests/test_cleaning.cpp:473:5: error: 'REQUIRE_THROWS_WITH' was not declared in this scope; did you mean 'REQUIRE_THROWS_AS'?
```

This repo's current Catch2 setup only uses `REQUIRE_THROWS_AS`, so please do not switch to `REQUIRE_THROWS_WITH` here. For the missing-column regression, assert the existing native exception type instead, which comes from `Frame::column_index(...)` as `std::out_of_range`:

```cpp
REQUIRE_THROWS_AS(
    drop_duplicates(f, std::vector<std::string>{"missing"}, "first"),
    std::out_of_range);
```

If you also want to assert the message, use an explicit `try`/`catch` and check `ex.what()`, but the type assertion is enough for this validation-path regression. Vercel can be ignored; the blocker is the failing C++ native build.

im-anishraj (changes requested):
@SHRESTHA-012 thanks for fixing the C++ compile blocker on the new head (`fd5f48b`). I rechecked the targeted Python regression locally, and the wrapper behavior is still failing:

```powershell
python -m pytest tests/test_cleaning.py -q -k 'drop_duplicates_zero_col'
```

Failure:

```text
TestDropDuplicates.test_drop_duplicates_zero_col_subset_none_preserves_rows
assert (1, 0) == (3, 0)
```

So the C++ zero-column path is now compiling, but the Python `drop_duplicates(frame)` path still collapses a 3-row / 0-column frame down to 1 row. Please update the Python-side zero-column handling so it preserves the original row count, then rerun the targeted `drop_duplicates_zero_col` tests before requesting review again.

I also attempted the local C++ test target, but this Windows environment does not have `nmake` configured; the hosted C++ native checks are the useful signal there, and those are looking much better. The remaining blocker I can verify locally is the Python regression above.

im-anishraj (changes requested):
@SHRESTHA-012 thanks for the quick update. I re-ran the targeted Python regression on the current head (`1c6a395`):

```powershell
python -m pytest tests/test_cleaning.py -q -k 'drop_duplicates_zero_col'
```

It still fails in the same public API path:

```text
TestDropDuplicates.test_drop_duplicates_zero_col_subset_none_preserves_rows
assert (1, 0) == (3, 0)
```

So replacing `from_pandas(to_pandas(frame))` with `copy.deepcopy(frame)` did not preserve the zero-column row count either. Please avoid a pandas/deepcopy roundtrip for this fast path. Since the C++ implementation now has the zero-column clone behavior after validation, the Python wrapper should either let `_drop_duplicates(...)` handle the zero-column case or return a wrapper around a native clone that preserves `num_rows()`. The important target remains: `ar.drop_duplicates(ar.from_pandas(pd.DataFrame(index=range(3)))).shape == (3, 0)`.

I am still not merging until this local Python regression passes, even though the C++ CI is improving.

SHRESTHA-012:
Hi @im-anishraj I have made all the necessary changes. Please check.

im-anishraj (approved):
Approved. I rechecked the current head (`a601e43`) after the latest push: the Python zero-column fast path now preserves the row count, the empty and missing `subset` validation paths still raise, and the C++ zero-column validation order is covered by the native tests.

The hosted non-Vercel matrix is green now, including lint, Python test matrix, C++ native tests, wheels/smoke checks, benchmark, and scikit-learn integration. The Vercel authorization failure can be ignored for this repo.

SHRESTHA-012:
Hi @im-anishraj, this issue touches both the Python wrapper and the C++ native layer, and required debugging across the full stack. Would it be possible to relabel this from beginner to intermediate? Happy to leave it to your discretion.

im-anishraj:
@SHRESTHA-012 yes, that is fair for this PR. The linked issue #1957 is already classified as intermediate, and the final fix touched both the Python wrapper behavior and the C++ native regression path.

I updated the PR labels from beginner / level-1 to intermediate / level-2. The approval still stands; no extra code changes are needed for the label update.

SHRESTHA-012:
Thank you so much @im-anishraj

PR #2196, opened 2026-05-31
fix: add context to custom validator errors

### LINKED ISSUE #2111: Bug: custom validator exceptions lose schema context
I ran into this while testing custom schema validators for GSSoC 2026. If a registered validator raises an exception, `validate()` just lets the original exception bubble out of the pandas `Series.map()` call.

Line: `arnio/schema.py:2555`

The rough path is:

```python
invalid = non_null[
    ~non_null.map(lambda v: _normalize_validator_result(fn(v), validator_name))
]
```

That works when the validator returns `True` / `False`, but it is pretty hard to debug when the validator itself throws. The error loses the schema context, so the caller sees something like `RuntimeError: bad value: 1` instead of knowing which column and custom validator caused it.

Small repro:

```python
import pandas as pd
import arnio as ar

ar.register_validator(
    "boom_ctx_demo",
    lambda v: (_ for _ in ()).throw(RuntimeError(f"bad value: {v}")),
)

frame = ar.from_pandas(pd.DataFrame({"score": [1, 2]}))
ar.validate(frame, {"score": ar.Custom("boom_ctx_demo")})
```

Current behavior:

```text
RuntimeError: bad value: 1
```

Expected behavior: validation should either convert this into a validation issue or wrap the exception with enough context to debug it, for example the column name, custom validator name, and the value/index being processed.

Why it matters: custom validators are exactly where project-specific logic lives. When one breaks in a larger schema, a raw exception without column/validator context makes CI failures and production validation logs much harder to trace.

Suggested fix: route custom validator calls through a small helper instead of an inline lambda. That helper can catch exceptions from `fn(v)` and raise a clearer Arnio/schema validation error that includes `column=name` and `validator_name`, while still preserving the original exception as `__cause__`. A regression test in `tests/test_register_validator.py` could assert that a raising validator reports both the column and custom validator name.

Requested labels: `bug`, `type:bug`, `gssoc`, `gssoc:level-2`, `level:intermediate`, `area:schema`, `area:python-api`, `difficulty:intermediate`, `size:s`.

eshaanag:
Fixes #2111

This PR keeps the change scoped to the custom-validator exception path in schema validation.

What changed:
- Added a small helper around registered custom validator execution.
- If the validator itself raises, the error is wrapped with the custom validator name, column name, and offending value.
- Preserves the original validator exception as `__cause__`.
- Leaves existing True/False/None/pd.NA and invalid return-value behavior unchanged.
- Added regression coverage in `tests/test_register_validator.py`.
- Updated the existing schema test that previously asserted raw exception propagation.

Validation:
- `.venv-codex/bin/python -m pytest tests/test_register_validator.py -q`
- `.venv-codex/bin/python -m pytest tests/test_schema.py tests/test_register_validator.py -q`
- `.venv-codex/bin/ruff check arnio/schema.py tests/test_register_validator.py tests/test_schema.py`
- `BLACK_NUM_WORKERS=1 .venv-codex/bin/black --check arnio/schema.py tests/test_register_validator.py tests/test_schema.py`
- `git diff --check`

This is for GSSoC 2026 and stays limited to the assigned schema bug.

### DIFF
--- a/arnio/schema.py
+++ b/arnio/schema.py
@@ -2563,7 +2563,12 @@ def _validate_column(
             else:
                 invalid = non_null[
                     ~non_null.map(
-                        lambda v: _normalize_validator_result(fn(v), validator_name)
+                        lambda v: _run_custom_validator(
+                            fn,
+                            v,
+                            column=name,
+                            validator_name=validator_name,
+                        )
                     )
                 ]
                 issues.extend(
@@ -3008,6 +3013,24 @@ def _normalize_validator_result(result: object, validator_name: str) -> bool:
         f"{type(result).__name__!r} ({result!r}); "
         "validators must return True (pass) or False (fail)."
     )
+
+
+def _run_custom_validator(
+    fn: Callable[[Any], object],
+    value: Any,
+    *,
+    column: str,
+    validator_name: str,
+) -> bool:
+    """Run a registered custom validator and add schema context on failure."""
+    try:
+        result = fn(value)
+    except Exception as exc:
+        raise ArnioError(
+            f"Custom validator {validator_name!r} failed for column "
+            f"{column!r} with value {value!r}: {exc}"
+        ) from exc
+    return _normalize_validator_result(result, validator_name)
 
 
 def Custom(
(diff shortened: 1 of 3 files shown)


im-anishraj (approved):
@eshaanag Thanks for the focused fix. I reviewed the schema change and regression coverage, added the missing GSSoC classification labels, and verified it locally with git diff --check, Black, Ruff, and pytest tests/test_register_validator.py tests/test_schema.py -q (390 passed). The GitHub Actions matrix is green; the only failing check is the known Vercel authorization issue, so this is ready to merge.

PR #2237, opened 2026-05-31
fix(website): handle unavailable localStorage in theme script

### LINKED ISSUE #1640: Website: handle unavailable localStorage in theme script
## Summary
The static website theme script reads and writes `localStorage` without a safety wrapper. In browsers or embedded contexts where storage access is blocked, `website/js/theme.js` can throw before the page finishes initializing.

## Why it matters
The theme script is loaded in the document head on every website page. A storage exception can prevent the theme toggle from initializing and may stop later website scripts from running, which makes the public documentation site more fragile.

## Evidence
- `website/js/theme.js`: `getPreferred()` calls `localStorage.getItem(STORAGE_KEY)` directly.
- `website/js/theme.js`: `apply()` calls `localStorage.setItem(STORAGE_KEY, theme)` directly.
- `website/js/theme.js`: the script calls `apply(getPreferred())` immediately during script evaluation.
- `website/js/github.js` already handles storage failures defensively, so the website has a safer local pattern to follow.

## Reproduction or trigger
1. Open the website in a browser/context where `localStorage` access throws, such as some privacy-restricted embedded views or locked-down browser settings.
2. Load any website page that includes `js/theme.js`.
3. Observe the theme script fail during initial page evaluation.

## Expected behavior
The website should still render and initialize when local storage is unavailable. Theme persistence should degrade gracefully to a default or system preference.

## Actual behavior
The theme code assumes `localStorage` is always readable and writable.

## Suggested direction
Wrap storage reads/writes in small helper functions with `try`/`catch`, similar to the defensive handling in `website/js/github.js`. If storage is unavailable, fall back to the default theme without throwing.

## References
- `website/js/theme.js`
- `website/js/github.js`

Exodus2004:
Fixes #1640.

### Description
This PR addresses issue #1640 by wrapping all \localStorage\ access in \website/js/theme.js\ with defensive \	ry...catch\ blocks. This ensures that the website theme script (which runs in the \<head>\) does not crash in browser environments where storage access is blocked (e.g., privacy-restricted modes or certain embedded views).

### Changes
- Implemented \safeGet(key)\ and \safeSet(key, value)\ helper functions.
- Replaced all direct \localStorage\ calls with these safe wrappers.
- Verified that theme persistence still works when storage is available and degrades gracefully to system/default preferences when storage is blocked.

### DIFF
--- a/website/js/theme.js
+++ b/website/js/theme.js
@@ -9,8 +9,24 @@
   const DARK = 'dark';
   const LIGHT = 'light';
 
+  function safeGet(key) {
+    try {
+      return localStorage.getItem(key);
+    } catch (e) {
+      return null;
+    }
+  }
+
+  function safeSet(key, value) {
+    try {
+      localStorage.setItem(key, value);
+    } catch (e) {
+      // Ignore quota or access errors
+    }
+  }
+
   function getPreferred() {
-    const stored = localStorage.getItem(STORAGE_KEY);
+    const stored = safeGet(STORAGE_KEY);
     if (stored === DARK || stored === LIGHT) return stored;
     // Fall back to system preference
     return window.matchMedia && window.matchMedia('(prefers-color-scheme: dark)').matches ? DARK : LIGHT;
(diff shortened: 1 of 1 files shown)


im-anishraj (approved):
@Exodus2004 thanks, this is a clean defensive fix for #1640. I reviewed the `theme.js` change and verified it locally with:

- `python scripts/check_website.py` (`9 pages`, passed)
- a Node VM probe where global `localStorage` throws on access; the theme script still loaded and fell back safely.

The hosted non-Vercel checks are also green. Approved.

PR #2249, opened 2026-05-31
fix(io): prevent temp file leak when close() raises in _materialize_c…

### LINKED ISSUE #2194: [Bug] Temporary file leaked on disk when tmp.close() raises during _normalize_csv_source
## Summary

In `arnio/io.py`, `_normalize_csv_source` creates a `NamedTemporaryFile` with `delete=False`. The cleanup call `os.unlink(tmp.name)` is placed after `tmp.close()` in the exception handler. If `tmp.close()` itself raises, the `os.unlink` line is never reached.

## Affected File

`arnio/io.py` (`_normalize_csv_source`, lines 373-395)

```python
tmp = tempfile.NamedTemporaryFile(delete=False, suffix='.csv')
try:
    tmp.write(source.read())
    tmp.close()          # if this raises (e.g., disk full)...
    return tmp.name
except Exception:
    os.unlink(tmp.name)  # ...this line is never reached
    raise
```

## Impact

On a disk-full condition or OS-level flush error, `tmp.close()` can raise, leaving the partial temporary file on disk. Over repeated failures this silently accumulates temp files in the system temp directory, eventually causing the same disk-full condition to recur or consuming inode space.

## Suggested Fix

Use a `try/finally` pattern to separate file closing from file deletion:

```python
tmp = tempfile.NamedTemporaryFile(delete=False, suffix='.csv')
try:
    tmp.write(source.read())
    tmp.close()
    return tmp.name
except Exception:
    try:
        tmp.close()
    except OSError:
        pass
    try:
        os.unlink(tmp.name)
    except OSError:
        pass
    raise
```

## Severity

Medium

ack-chai:
prevent temp file leak when close() raises in _materialize_csv_input

Fixes #2194

### DIFF
--- a/arnio/io.py
+++ b/arnio/io.py
@@ -390,8 +390,14 @@ def _materialize_csv_input(
             tmp.close()
             return tmp.name, True, True
         except Exception:
-            tmp.close()
-            os.unlink(tmp.name)
+            try:
+                tmp.close()
+            except OSError:
+                pass
+            try:
+                os.unlink(tmp.name)
+            except OSError:
+                pass
             raise
 
     raise TypeError("read_csv expected a filesystem path or text file-like object")
--- a/tests/test_materialize_csv_cleanup.py
+++ b/tests/test_materialize_csv_cleanup.py
@@ -76,7 +76,6 @@ class TestMaterializeCsvCleanup:
 
     def test_unlink_failure_does_not_mask_original_exception(self):
         """If unlink() also raises during cleanup, the original exception propagates."""
-        real_unlink = os.unlink
 
         def close_raising_ntf(**kwargs):
             handle = tempfile.NamedTemporaryFile(**kwargs)
--- a/tests/test_materialize_csv_cleanup.py
+++ b/tests/test_materialize_csv_cleanup.py
@@ -7,6 +7,7 @@ must still run so the temp file is not leaked on disk.
 
 import io
 import os
+import tempfile
 from unittest.mock import patch
 
 import pytest
@@ -19,6 +20,22 @@ class _FailOnReadStream:
 
     def read(self, size=-1):
         raise OSError("simulated read failure")
+
+
+def _ntf_with_flaky_close(**kwargs):
+    """Return a real NamedTemporaryFile whose close() raises on first call."""
+    handle = tempfile.NamedTemporaryFile(**kwargs)
+    real_close = handle.close
+    call_count = [0]
+
+    def flaky_close():
+        call_count[0] += 1
+        if call_count[0] == 1:
+            raise OSError("simulated flush error on close")
+        real_close()
+
+    handle.close = flaky_close
+    return handle
 
 
 class TestMaterializeCsvCleanup:
@@ -41,8 +58,8 @@ class TestMaterializeCsvCleanup:
 
         with (
             patch(
-                "arnio.io.tempfile.NamedTemporaryFile.close",
-                side_effect=OSError("simulated flush error on close"),
+                "arnio.io.tempfile.NamedTemporaryFile",
+                side_effect=_ntf_with_flaky_close,
             ),
             patch("arnio.io.os.unlink", side_effect=tracking_unlink),
         ):
@@ -59,11 +76,21 @@ class TestMaterializeCsvCleanup:
 
     def test_unlink_failure_does_not_mask_original_exception(self):
         """If unlink() also raises during cleanup, the original exception propagates."""
+        real_unlink = os.unlink
+
+        def close_raising_ntf(**kwargs):
+            handle = tempfile.NamedTemporaryFile(**kwargs)
+            handle.close = lambda: (_ for _ in ()).throw(
+                OSError("simulated close error")
+            )
+            return handle
+
         source = _FailOnReadStream()
+
         with (
             patch(
-                "arnio.io.tempfile.NamedTemporaryFile.close",
-                side_effect=OSError("close error"),
+                "arnio.io.tempfile.NamedTemporaryFile",
+                side_effect=close_raising_ntf,
             ),
             patch("arnio.io.os.unlink", side_effect=OSError("unlink error")),
         ):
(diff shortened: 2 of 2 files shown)


im-anishraj (changes requested):
@ack-chai thanks for jumping on #2194. The implementation is in the right area, and I added the GSSoC classification labels, but I cannot approve or merge this yet because local verification and CI are failing.

Please fix these blockers:

- `py -3.13 -m black --check arnio/io.py tests/test_materialize_csv_cleanup.py` fails because `tests/test_materialize_csv_cleanup.py` needs Black formatting.
- The focused regression test currently fails against this PR head: `tests/test_materialize_csv_cleanup.py::TestMaterializeCsvCleanup::test_unlink_runs_even_when_close_raises` hits `RecursionError`. The cause is that `patch("arnio.io.os.unlink", side_effect=tracking_unlink)` patches the shared `os` module object, so `tracking_unlink()` calling `os.unlink(path)` calls the mock again recursively. Please store the real unlink function before the patch, for example `real_unlink = os.unlink`, and call that inside `tracking_unlink`.
- CI is still not green: GitHub lint is failing, other checks are still settling, and Vercel is failing due authorization.

After pushing the fixes, please rerun the focused cleanup test and lint. I am keeping this PR blocked until formatting, tests, and CI are clean.

im-anishraj (changes requested):
@ack-chai thanks for the quick follow-up fixes. The formatting issue is now resolved locally, but the focused regression tests still fail on the latest head `e400a02`.

What now passes locally:
- `py -3.13 -m black --check arnio/io.py tests/test_materialize_csv_cleanup.py`
- `py -3.13 -m ruff check arnio/io.py tests/test_materialize_csv_cleanup.py`

Remaining blocker:
- `py -3.13 -m pytest tests/test_materialize_csv_cleanup.py -q` fails in the two close-error tests. The patch target `arnio.io.tempfile.NamedTemporaryFile.close` is invalid because `NamedTemporaryFile` is a function and does not have a `close` attribute. Please patch the returned temporary handle's `close` method instead, for example by wrapping `arnio.io.tempfile.NamedTemporaryFile` with a helper that returns a real handle whose `.close` raises on the cleanup path.

Once those tests pass and CI settles cleanly, I can re-review. I am keeping this blocked for now.

im-anishraj (changes requested):
@ack-chai thanks for the latest rewrite. I rechecked head `9643ef2` locally and the hygiene plus focused tests now pass:

- `py -3.13 -m black --check arnio/io.py tests/test_materialize_csv_cleanup.py`
- `py -3.13 -m ruff check arnio/io.py tests/test_materialize_csv_cleanup.py`
- `py -3.13 -m pytest tests/test_materialize_csv_cleanup.py -q` (6 passed)

I am still keeping this blocked because the rewritten tests no longer exercise the original regression described in the PR and issue: `tmp.close()` raising during exception cleanup while `os.unlink()` still runs. The tests now cover read failures and unlink failures, but not the close-failure path that the implementation specifically changed.

Please add one focused regression test for the close-error case without patching `NamedTemporaryFile.close` on the function object. A good pattern is to patch `arnio.io.tempfile.NamedTemporaryFile` to return a lightweight fake temp handle (or wrapped real handle) with `name`, `write()`, and a `close()` method that raises, then assert `arnio.io.os.unlink` is still called and the original read exception is re-raised. Once that is covered, I can re-review quickly.

im-anishraj:
@ack-chai thanks for adding the close-error regression coverage. I checked the new head `3b1d942`; lint, PR Guard, examples, parquet, and compatibility are clean, but CI is still failing in the Python matrix.

The failing test is the new close-error regression:

```text
tests/test_materialize_csv_cleanup.py::TestMaterializeCsvCleanup::test_unlink_runs_when_close_raises_on_cleanup
RecursionError: maximum recursion depth exceeded
```

Root cause: inside `_FakeTmp.__init__`, the test calls `tempfile.NamedTemporaryFile(...)` while `arnio.io.tempfile.NamedTemporaryFile` is patched. Since `arnio.io.tempfile` and the test's imported `tempfile` module are the same module object, that call recursively hits the mock.

Please save the real function before the patch, for example `real_named_temporary_file = tempfile.NamedTemporaryFile`, and use that inside `_FakeTmp.__init__`. After that, rerun `python -m pytest tests/test_materialize_csv_cleanup.py -q` and let CI settle again. Keeping this blocked until the regression test passes in CI.

im-anishraj:
@ack-chai thanks for the quick new commit. CI is still failing on the same focused test at head `6dd1478`:

```text
tests/test_materialize_csv_cleanup.py::TestMaterializeCsvCleanup::test_unlink_runs_when_close_raises_on_cleanup
RecursionError: maximum recursion depth exceeded
```

The remaining recursion is now in `_FakeTmp.__init__`: it calls `tempfile.NamedTemporaryFile(...)` while `arnio.io.tempfile.NamedTemporaryFile` is patched, and both references point at the same module object. The new `_REAL_UNLINK` helper fixed the unlink side, but the fake temp handle also needs to capture the real `NamedTemporaryFile` before the patch, for example at module level:

```python
_REAL_NAMED_TEMPORARY_FILE = tempfile.NamedTemporaryFile
```

Then `_FakeTmp.__init__` should call `_REAL_NAMED_TEMPORARY_FILE(...)`. Please rerun `python -m pytest tests/test_materialize_csv_cleanup.py -q` after that and wait for the full matrix to settle green.

im-anishraj (approved):
@ack-chai thanks for the quick fix. I rechecked head `fd092fa` and the focused local verification is clean:

```powershell
py -3.13 -m black --check arnio/io.py tests/test_materialize_csv_cleanup.py
py -3.13 -m ruff check arnio/io.py tests/test_materialize_csv_cleanup.py
py -3.13 -S -  # controlled import of this PR worktree, then pytest
# tests/test_materialize_csv_cleanup.py: 7 passed
```

GitHub project CI is also green now, including the full Python matrix, lint, PR Guard, examples, parquet, benchmark, compatibility, C++ native tests, CodeQL, and scikit-learn integration. Approving this head.

I am not merging in this pass because the visible Vercel authorization status is still failed/external; no action is needed from you for that status.

PR #2286, opened 2026-06-01
fix(asan): reject bool fill values for numeric columns before entering C++

ShreeyaSahai:
## Summary

Under ASAN with `abort_on_error=1`, passing a `bool` fill value to a
numeric (`int64` or `float64`) column causes `coerce_value()` in
`cleaning.cpp` to throw `std::invalid_argument` inside the compiled
`.so`. ASAN intercepts the throw before pybind11 can translate it into a
Python exception, aborting the process with exit code 134.

The fix adds a targeted Python-side check in `fill_nulls` that rejects
`bool` fill values for numeric columns before the value ever reaches C++.
The existing scalar type check at line 440 does not catch this case
because `bool` is a subclass of `int` in Python, so `isinstance(True,
(str, int, float, bool))` returns `True` and the bad value passes
through.

## Linked issue

Refs #1222

## Type of change

- [x] Bug fix

## Area

- [x] Python API / ArFrame

## Changes

One file changed — `arnio/cleaning.py`.

Added a bool guard inside `fill_nulls` between the existing scalar type
check and the `_fill_nulls` C++ call. When the fill value is `bool` and
any target column has dtype `int64` or `float64`, a `TypeError` is raised
in Python before C++ is entered.

## Testing

- [x] `make test`

The following test which previously aborted the process now passes
cleanly:

```python
def test_incompatible_fill_rejected():
    frame = ar.from_pandas(pd.DataFrame({"a": [1, None, 3]}))
    with pytest.raises(TypeError):
        ar.fill_nulls(frame, True)
```

## Performance Impact

None. The guard only runs on the error path — one `isinstance` check and
a `dtypes()` call that is skipped entirely for non-bool fill values.

## Contributor checklist

- [x] My PR title uses Conventional Commits (`fix(asan): reject bool fill values for numeric columns before entering C++`).

## Maintainer notes

This is PR 2 of 4 fixing the ASAN workflow failures from #260 and #710 .

PR 1 (`fix: remove invalid Frame constructor arguments in cleaning.cpp`)
must be merged first as it fixes the build break that prevented CI from
reaching the test stage.

Remaining PRs:
- PR 3 — CSV preflight validation in `io.py`
- PR 4 — ASAN workflow (`asan.yml`) and `pyproject.toml` `build-dir` fix


im-anishraj (changes requested):
@ShreeyaSahai thanks for isolating the ASAN bool-fill case. I cannot approve this head yet because the PR changes behavior but does not add a focused regression test in the diff.

Please add a test that demonstrates `fill_nulls(frame, True)` is rejected for `int64`/`float64` target columns before entering C++, including the `subset` path. Also, please prefer the public `frame.dtypes` metadata instead of reaching through `frame._frame.dtypes()` in Python API code.

The repeated full-matrix failure for `rename_columns_matching` appears to be current-main website API-doc drift, and Vercel is external/auth-limited, but the missing regression coverage is still a PR blocker.

ShreeyaSahai:
Thanks for the feedback. Made the following changes:

- Replaced frame._frame.dtypes() with the public frame.dtypes API

- Added four regression tests covering int64, float64, the subset path, and confirming that a bool fill for a bool column still works correctly

Ready for re-review.

im-anishraj (changes requested):
@ShreeyaSahai thanks for adding the requested tests and switching to the public `frame.dtypes()` API. One blocker remains: the latest diff still has whitespace/formatting issues in `tests/test_cleaning.py` around the new test block, and GitHub still is not showing the normal non-Vercel check set for this head. Please run the formatter/checks on `arnio/cleaning.py tests/test_cleaning.py` and push the clean result, then request review again.

ShreeyaSahai:
Thanks for the feedback, I have run the formatter/checks on arnio/cleaning.py tests/test_cleaning.py and pushed the clean result. Ready for re-review.

im-anishraj:
@ShreeyaSahai thanks for the update. I rechecked the latest diff and the whitespace/formatting issue called out in the last review appears cleaned up.

I am not approving yet because GitHub currently only shows the Vercel check for this PR; the normal non-Vercel Actions are not visible for the latest head. Please make sure the branch is up to date and that the regular Actions are triggered/approved. Once those checks are visible and green, request review again. Vercel is external/auth-only and is not the blocker.

ShreeyaSahai:
@im-anishraj The branch is now up to date with no conflicts. I can see "6 workflows awaiting approval" on the PR and it shows that these require maintainer approval to run. Could you approve the workflows so CI can execute? I'm unable to trigger them from my side.

im-anishraj:
@ShreeyaSahai I approved the pending repository Actions for the latest head. Please wait for the non-Vercel checks to finish; I can re-review once they are visible and green.

im-anishraj:
@ShreeyaSahai the repository Actions are now visible, and this head is not ready yet. Multiple test jobs are failing, starting with the new `fill_nulls` behavior checks:

- `TestFillNulls::test_fill_nulls_accepts_valid_scalars`
- `TestFillNulls::test_fill_nulls_rejects_bool_for_int64_column`
- `TestFillNulls::test_fill_nulls_rejects_bool_for_float64_column`
- `TestFillNulls::test_fill_nulls_rejects_bool_via_subset`
- `TestFillNulls::test_fill_nulls_bool_accepted_for_bool_column`

Please rework the bool handling so bool fill values are rejected only for numeric targets while preserving the existing valid-scalar and bool-column behavior, then rerun the targeted cleaning tests. Vercel is external/auth-only; the non-Vercel test failures are the blockers.

im-anishraj:
@ShreeyaSahai thanks for the latest update. I checked the current non-Vercel CI on head `185c816`, and the full matrix is still failing on the bool-fill regression tests.

The representative Python 3.12 Ubuntu job reports:

- `TestFillNulls::test_fill_nulls_rejects_bool_for_int64_column` -> `Failed: DID NOT RAISE <class 'TypeError'>`
- `TestFillNulls::test_fill_nulls_rejects_bool_for_float64_column` -> `Failed: DID NOT RAISE <class 'TypeError'>`
- `TestFillNulls::test_fill_nulls_rejects_bool_via_subset` -> `Failed: DID NOT RAISE <class 'TypeError'>`

`test_fill_nulls_bool_accepted_for_bool_column` passes, so the remaining work is specifically making the int64/float64 and subset paths reject bool fill values before entering the C++ layer. Vercel is still external/auth-only and is not the blocker; the non-Vercel test failures are.

im-anishraj:
@ShreeyaSahai, I rechecked the updated head `c0ee8811` after the merge from `main`. Formatting is clean, but the bool-fill behavior is still failing locally and across the non-Vercel matrix.

The guard currently checks only:

```python
if dtype_map.get(col_name) == "int64":
```

However, `pd.DataFrame({"a": [1, None, 3]})` becomes `float64` because of the null, and the explicit float test is also `float64`. Therefore neither case raises, and these three focused tests fail locally:

```text
test_fill_nulls_rejects_bool_for_int64_column
 test_fill_nulls_rejects_bool_for_float64_column
 test_fill_nulls_rejects_bool_via_subset
```

Please reject bool values for both Arnio numeric dtypes (`int64` and `float64`). To make the int64 test genuinely exercise int64-with-null behavior, construct that input with a nullable integer dtype if supported by the conversion path, rather than relying on pandas' default coercion. Then rerun the five focused `TestFillNulls` cases and the non-Vercel matrix. Vercel is unrelated external noise.

im-anishraj (changes requested):
@ShreeyaSahai, thanks for updating both numeric dtypes and fixing the integer fixture. One small but blocking condition error remains on head `ad4f8b37`:

```python
if dtype_map.get(col_name) == ("int64", "float64"):
```

`dtype_map.get(col_name)` is a dtype string (or `None`), so it can never equal that tuple. Please use membership instead:

```python
if dtype_map.get(col_name) in ("int64", "float64"):
```

I verified the exact head locally: the valid-scalar and bool-column cases pass, while all three numeric rejection cases fail with `DID NOT RAISE` (int64, float64, and subset). GitHub CI reports the same three failures across the test matrix, and mypy also flags the non-overlapping comparison at `arnio/cleaning.py:506`.

Please make that correction and rerun the five focused `TestFillNulls` cases plus the non-Vercel checks. Vercel remains unrelated external authorization noise.

ShreeyaSahai:
Thanks for the detailed review and for pinpointing the issue.

You're right, the comparison against `("int64", "float64")` was incorrect since `dtype_map.get(col_name)` returns a single dtype string (or `None`), not a tuple. I've updated the condition to use membership checking instead.

I've pushed the fix and rerun the relevant checks. The focused `TestFillNulls` cases now pass locally, including the numeric rejection scenarios that were previously failing. The non-Vercel checks are also passing on my end.

Thanks again for catching this and for the clear explanation of the failing cases and mypy warning. I appreciate the thorough review.

im-anishraj (approved):
@ShreeyaSahai, thank you for correcting the numeric dtype membership check and strengthening the integer fixture. I re-ran the updated head locally: formatting, Ruff, mypy, and all 15 focused fill_nulls tests pass. The complete non-Vercel GitHub Actions matrix is also green, so this is approved for merge.

PR #2287, opened 2026-06-01
fix(asan): broaden exception catch in read_csv and scan_csv to prevent ASAN abort

ShreeyaSahai:
## Summary

Under ASAN with `abort_on_error=1`, C++ exceptions thrown inside
`csv_reader.cpp` (unterminated quoted records, duplicate headers, missing
files) abort the process with exit code 134 instead of raising a clean
Python exception. This happens because ASAN intercepts `__cxa_throw`
before pybind11 can translate `std::runtime_error` into Python
`RuntimeError`.

Both `read_csv` and `scan_csv` caught only `RuntimeError`, which relies
on pybind11's translation completing successfully. Under ASAN that
translation never happens. Broadening the catch to `Exception` ensures
any C++ exception that does cross the boundary is caught and wrapped as
`CsvReadError` rather than killing the process.

## Linked issue

Refs #1222 

## Type of change

- [x] Bug fix

## Area

- [x] CSV parser / I/O

## Changes

One file changed — `arnio/io.py`.

- `read_csv`: `except RuntimeError` broadened to `except Exception`
- `scan_csv`: `except RuntimeError` replaced with `except CsvReadError:
  raise` guard followed by `except Exception`

## Testing

- [x] `make test`

The following tests which previously aborted the process now pass:

- `test_unterminated_quote_rejected`
- `test_duplicate_headers_rejected`

## Performance Impact

None. Exception handling only runs on the error path.

## Contributor checklist

- [x] My PR title uses Conventional Commits (`fix(asan): broaden exception catch in read_csv and scan_csv to prevent ASAN abort`).

## Maintainer notes

This is PR 3 of 4 fixing the ASAN workflow failures from #260 and #710.

- PR 1 — C++ build fixes (must merge first)
- PR 2 — Python-side bool guard for `fill_nulls`
- **PR 3 — This PR**
- PR 4 — ASAN workflow (`asan.yml`) and `pyproject.toml` `build-dir` fix


im-anishraj (changes requested):
@ShreeyaSahai thanks for digging into the ASAN CSV abort path. I cannot approve this exact change yet because broadening the wrapper from `RuntimeError` to `Exception` changes Arnio's public error contract.

In `read_csv`, Python-side validation `TypeError`s would now be caught and re-raised as `CsvReadError`. In `scan_csv`, Python-side `ValueError`s from options such as `sample_size`/delimiter validation would also be wrapped because there is no `except ValueError: raise` guard there. That would regress existing API behavior unrelated to the C++ exception translation issue.

Please narrow this to the C++/pybind exception path or explicitly preserve `TypeError`/`ValueError` before wrapping, and add focused tests showing option-validation errors keep their current exception types while the ASAN-triggering CSV cases still become clean `CsvReadError`s. Vercel is external/auth-limited and not your blocker.

ShreeyaSahai:
Thanks for the review. I've updated the implementation to preserve the existing public error contract by explicitly re-raising ValueError and TypeError before wrapping other exceptions as CsvReadError in both read_csv and scan_csv.

I also added regression tests covering:

- invalid delimiter → ValueError
- invalid sample_size type → TypeError
- malformed CSV (unterminated quote) → CsvReadError

The goal is to keep Python-side validation behavior unchanged while still ensuring CSV parsing failures on the C++/pybind path are surfaced as clean CsvReadErrors.

im-anishraj:
@ShreeyaSahai thanks for the update. I rechecked the latest diff and preserving `ValueError` / `TypeError` before wrapping the remaining native/parser failures as `CsvReadError` addresses the main public-contract concern from the previous review.

I am not approving yet because GitHub currently only shows the Vercel check for this PR; the regular non-Vercel Actions are not visible for the latest head. Please make sure those checks are triggered/approved and green, then request review again. Vercel is external/auth-only and is not the blocker.

im-anishraj:
@ShreeyaSahai I approved the pending repository Actions for the latest head. Please wait for the non-Vercel checks to finish; I can re-review once they are visible and green.

im-anishraj:
@ShreeyaSahai merged this. Thanks for the careful follow-up on the public exception contract.

Verification before merge:
- `ruff check arnio/io.py tests/test_csv.py`
- `black --check arnio/io.py tests/test_csv.py`
- `pytest tests/test_csv.py -q -k "invalid_delimiter_raises_value_error or invalid_sample_size_raises_type_error or unterminated_quote_raises_csv_read_error"` (`3 passed`)

The remaining Vercel status was the known external authorization check, so it was not treated as a blocker. I added `gssoc:approved` and `quality:clean` after merge.

PR #2288, opened 2026-06-01
ci: add ASAN workflow and fix build-dir packaging setting

ShreeyaSahai:
## Summary

Adds the ASAN (AddressSanitizer) CI workflow that triggered the discovery
of all the bugs fixed in PRs 1–3. Also adds `build-dir = ""` to
`pyproject.toml` to prevent scikit-build-core from writing build
artifacts into a versioned subdirectory, which was causing stale `.so`
files to be picked up on repeat editable installs in CI.

## Linked issue

Refs #1222 and #260 

## Type of change

- [x] CI / packaging

## Area

- [x] Developer tooling

## Changes

**`.github/workflows/asan.yml`** — new workflow that:
- Builds the project with clang and `-fsanitize=address
  -fno-omit-frame-pointer`
- Resolves the clang ASAN runtime path dynamically (handles different
  Ubuntu/LLVM layouts without hardcoding a path)
- Runs `pytest tests/ -v -k "not benchmark"` with `LD_PRELOAD` set to
  the ASAN runtime and `ASAN_OPTIONS=detect_leaks=0:abort_on_error=1`
- Prints the ASAN log on failure for easy diagnosis
- Triggers on pull requests touching `cpp/src/**`, `arnio/**`, or
  `tests/**`, and on manual dispatch

**`pyproject.toml`** — adds `build-dir = ""` under `[tool.scikit-build]`
to prevent versioned build artifact subdirectories from causing stale
`.so` files to be loaded during repeat editable installs in CI.

## Testing

- [x] ASAN workflow passes locally and in CI with PRs 1–3 applied

## Performance Impact

None. CI-only change.

## Contributor checklist

- [x] My PR title uses Conventional Commits (`ci: add ASAN workflow and fix build-dir packaging setting`).

## Maintainer notes

This is PR 4 of 4. PRs 1–3 must be merged before this workflow will run
green on any future PR, since the ASAN aborts it detects are fixed in
those PRs. Opening this PR last means the workflow itself can be reviewed
independently of the bug fixes.


im-anishraj (changes requested):
@ShreeyaSahai thanks for proposing the ASAN workflow. I marked this PR as blocked for now because the PR body says it depends on the earlier ASAN fix PRs, and I currently only see the external Vercel authorization status for this head rather than a verified non-Vercel run of the workflow itself.

Before this can be approved, please make sure the prerequisite fixes are merged or rebase this branch so the workflow can be exercised against a clean base, then provide a successful `workflow_dispatch`/Actions run for `ASAN Checks`. Also please add the final newline at the end of `.github/workflows/asan.yml`.

Vercel is external/auth-limited and not your blocker; the blocker is that this sensitive CI workflow has not yet been verified in the repository context.

im-anishraj (changes requested):
@ShreeyaSahai, thanks for syncing with current `main` and adding the final newline. I rechecked the updated head, and the ASAN workflow still needs these changes before it can be verified:

1. Pin `actions/checkout` and `actions/setup-python` to immutable commit SHAs, consistent with the repository's existing workflows, instead of mutable `@v4`/`@v5` tags.
2. Expand the `pull_request.paths` filter so the workflow runs when its own file and relevant native/build surfaces change. It currently excludes `.github/workflows/asan.yml`, `pyproject.toml`, `cpp/include/**`, `bindings/**`, and the CMake files, so this PR cannot exercise the workflow it adds and important sanitizer-relevant changes can skip it.
3. Remove the trailing whitespace reported in `.github/workflows/asan.yml` at lines 44, 49, and 75 (`git diff --check` currently fails).
4. After those changes, please push a head where `ASAN Checks` actually appears and passes. The regular CI is largely green and Vercel remains external noise, but there is still no sanitizer run validating this workflow.

ShreeyaSahai:
Thank you for the detailed review and for rechecking the updated head.

I've addressed the requested changes:

- Pinned actions/checkout and actions/setup-python to the repository's standard immutable SHAs.
- Expanded the pull_request.paths filters to cover the workflow file and the relevant native/build surfaces.
- Removed the trailing whitespace reported by git diff --check.

I've pushed the updated changes and verified that the ASAN workflow is now being exercised. Thanks again for the thorough feedback and guidance.

im-anishraj (approved):
Thanks @ShreeyaSahai. I rechecked head `7be8366`: the checkout/setup actions are pinned, the ASAN runtime is resolved explicitly, the packaging build-dir concern is handled, and the new ASAN job is visible and passing alongside the full non-Vercel matrix.

The only failing check is the external Vercel authorization gate, which is not a blocker for this CI-focused PR.

im-anishraj (changes requested):
@ShreeyaSahai thanks for the update. I rechecked the workflow diff before merging and found one remaining blocker in the extension verification step:

```bash
SO=$(python -c "import arnio._core as m; import inspect; print(inspect.getfile(m))")
```

`arnio._core` is the Python shim, not the compiled extension. I verified locally on this head that `inspect.getfile(arnio._core)` resolves to `arnio/_core.py`, while the native extension is `arnio/_arnio_cpp...pyd`/`.so`. Because the next line runs `strings "$SO" | grep -q "invalid_argument"`, this would inspect the shim file instead of the compiled module and can fail even when the extension built correctly.

Please update the probe to inspect the compiled module directly, for example `import arnio._arnio_cpp as m`, or otherwise resolve the path from a native object/module. After that, this should be ready to merge because the rest of the workflow and non-Vercel checks look good.

im-anishraj (approved):
@ShreeyaSahai thanks for the follow-up. I rechecked the current head: the ASAN workflow now inspects `arnio._arnio_cpp` directly, the path filters and pinned action SHAs are in place, and the ASAN job plus the full non-Vercel matrix are green. The only failing status is the external Vercel authorization check, which is not a blocker here.

PR #2312, opened 2026-06-02
feat: add row lineage metadata through filtering and drop steps

### LINKED ISSUE #648: Feature: add row lineage metadata through filtering and drop steps
## Problem
Arnio currently lacks this product capability: Row-dropping steps need a way to map output rows back to source rows.

## Evidence / audit context
- Issue title: Feature: add row lineage metadata through filtering and drop steps
- Audit focus: Row-dropping steps need a way to map output rows back to source rows.
- Labels selected: type: feature, area: pipeline, priority: medium, difficulty: advanced, status: needs maintainer, size: l

## Suggested fix

## Acceptance criteria

dynamo-pentester:
## Summary

Implements row lineage metadata tracking through filtering and drop steps for `ar.pipeline()`.

Closes #648

---

### `arnio/pipeline.py`
- Added `_LINEAGE_SENTINEL_COL = "__arnio_lineage_id__"` module-level constant
- Added `LineageReport` frozen dataclass with:
  - `dropped_by_step: dict[str, list[int]]` — step name → sorted list of original row indices dropped by that step (empty list for steps that dropped nothing)
  - `total_dropped: int` — total rows dropped across all steps
  - `to_pandas()` — flat `DataFrame` with columns `original_index` and `step`
- Added `track_lineage: bool = False` parameter to `pipeline()` with `isinstance` guard matching the existing `return_metadata` / `dry_run` / `verbose` pattern
- Before the step loop: injects a hidden `int64` sentinel column (`0..n-1`) via `to_pandas → insert → from_pandas` when `track_lineage=True`. The C++ engine treats it as an ordinary column and filters/preserves it correctly through all row-dropping operations
- After each step: diffs surviving sentinel values against the previous set to record which original indices were dropped and by which step name. Non-dropping steps produce an empty diff naturally with no special-casing
- **`drop_duplicates` compatibility fix**: the sentinel assigns a unique value to every row, which would cause `drop_duplicates` to see every row as distinct and drop nothing. Fixed by auto-injecting `subset=<all user columns>` (excluding the sentinel) into its kwargs when `track_lineage=True`, honouring any explicit `subset` the caller already passed
- After the loop: strips sentinel via `drop(columns=[…])`, builds `LineageReport`, returns `(ArFrame, LineageReport)`
- Updated docstring with full parameter docs and return-type variants

### `arnio/__init__.py`
- Imported `LineageReport` from `.pipeline`
- Added `"LineageReport"` to `__all__`

### `website/api.html`
- Updated `pipeline()` entry to include `track_lineage=False` in the signature and describe the `(ArFrame, LineageReport)` return
- Added a new `LineageReport` entry documenting `dropped_by_step`, `total_dropped`, and `to_pandas()`

### `tests/test_pipeline_lineage.py` *(new file)*
41 focused tests across 9 classes:

| Class | What it covers |
|---|---|
| `TestLineageReportImport` | Top-level importability, frozen dataclass immutability |
| `TestTrackLineageTypeValidation` | `TypeError` on `str`, `int`, `None` |
| `TestDefaultBehaviourUnchanged` | `track_lineage=False` returns bare `ArFrame`, no sentinel column leaks |
| `TestSentinelColumnStripped` | Sentinel absent from output frame, column count unchanged |
| `TestLineageDropNulls` | Correct indices for single null, subset, multiple nulls, zero nulls |
| `TestLineageDropDuplicates` | `keep="first"`, `keep="last"`, no-dup case |
| `TestLineageFilterRows` | `>`, `==` operators; no-drop case |
| `TestLineageMultiStep` | Each step records only its own drops; indices always refer to original frame; `total_dropped` arithmetic; three-step chain |
| `TestLineageNonDroppingStep` | `strip_whitespace` and `rename_columns` produce empty entries |
| `TestLineageTotalDropped` | Sum invariant; zero; all rows dropped |
| `TestLineageReportToPandas` | Return type, column schema, correct rows, empty case, multi-step content, row count equals `total_dropped` |
| `TestLineageReturnType` | Tuple length, element types, empty steps case |

---

## Usage

```python
import arnio as ar

frame = ar.from_pandas(df)

result, lineage = ar.pipeline(
    frame,
    [
        ("drop_nulls",),
        ("drop_duplicates",),
    ],
    track_lineage=True,
)

lineage.dropped_by_step
# {"drop_nulls": [1, 4], "drop_duplicates": [7]}

lineage.total_dropped
# 3

lineage.to_pandas()
#    original_index            step
# 0               1      drop_nulls
# 1               4      drop_nulls
# 2               7  drop_duplicates
```

---

## What was not changed

- C++ core / `_core.py` — purely Python layer change
- Any cleaning function internals
- `return_metadata` behaviour — fully backward compatible (`track_lineage=False` default leaves every existing call site untouched)
- All existing CSV, pandas interop, schema, quality, and packaging tests continue to pass (139 existing pipeline tests + 3151 suite-wide)

---

## Test results

image
image

```
tests/test_pipeline_lineage.py   41 passed
tests/test_pipeline.py          139 passed
tests/test_website_api_docs.py    9 passed
```

### DIFF
--- a/arnio/__init__.py
+++ b/arnio/__init__.py
@@ -62,6 +62,7 @@ from .io import (
     write_parquet,
 )
 from .pipeline import (
+    LineageReport,
     PipelineContext,
     get_builtin_step_signatures,
     list_steps,
@@ -178,6 +179,7 @@ __all__ = [
     "get_builtin_step_signatures",
     "list_steps",
     "PipelineContext",
+    "LineageReport",
     "reset_steps",
     # Data quality
     "profile",
--- a/arnio/pipeline.py
+++ b/arnio/pipeline.py
@@ -86,6 +86,12 @@ class LineageReport:
         Mapping from step name to a sorted list of original row indices that
         were dropped by that step.  Steps that dropped no rows have an empty
         list.
+
+        When the same step name appears more than once in the pipeline, all
+        drops from every invocation of that step are merged under the single
+        key in sorted order.  Use ``track_lineage=True`` together with
+        per-step timing from ``return_metadata=True`` if you need to
+        distinguish individual invocations.
     total_dropped : int
         Total number of rows dropped across all steps.
 
@@ -435,6 +441,10 @@ def pipeline(
         When True, also return a metadata dictionary with per-step timing
         information in execution order.
 
+        Cannot be combined with ``track_lineage=True``.  If you need both
+        row-level lineage and per-step timings, run the pipeline twice: once
+        with ``track_lineage=True`` and once with ``return_metadata=True``.
+
     verbose : bool, default False
         Enable lightweight diagnostic logging through the ``arnio`` logger.
         Logs step index, step name, execution path, elapsed execution
@@ -450,18 +460,28 @@ def pipeline(
         instead of a bare ``ArFrame``.  The sentinel column is always stripped
         from the returned frame before it is handed back to the caller.
 
-        Custom Python steps receive the sentinel column as a regular ``int64``
-        column and **must not drop it**; document this constraint when writing
-        custom steps that will be used with ``track_lineage=True``.
+        When the same step name appears more than once in the pipeline, all
+        drops from every invocation of that step are merged under the single
+        key in ``LineageReport.dropped_by_step``, in sorted order.
 
-        Implies a complete pass through the pipeline (equivalent to
-        ``return_metadata=False`` behaviour for the frame result); use
-        ``return_metadata=True`` separately if per-step timings are also needed.
+        **Incompatible with** ``return_metadata=True``.  Combining both raises
+        ``ValueError``.
+
+        **Input column constraint**: the input frame must not already contain a
+        column named ``__arnio_lineage_id__``.  If it does, ``pipeline`` raises
+        ``ValueError`` before any steps are executed.
+
+        **Custom step contract**: custom Python steps that are used with
+        ``track_lineage=True`` must not drop or rename the sentinel column
+        ``__arnio_lineage_id__``.  If a custom step removes the sentinel, a
+        ``PipelineStepError`` is raised immediately after that step completes.
+
+        Cannot be combined with ``return_metadata=True``.
 
     Returns
     -------
     ArFrame
-        Data frame with all steps applied sequentially (``track_lineage=False``).
+        Data frame with all steps applied sequentially (default).
     tuple[ArFrame, LineageReport]
         Frame and lineage report when ``track_lineage=True``.
     tuple[ArFrame, dict]
@@ -470,11 +490,17 @@ def pipeline(
     Raises
     ------
     TypeError
-        If ``track_lineage`` is not a bool.
+        If any parameter has an unexpected type.
     ValueError
+        If ``track_lineage=True`` and ``return_metadata=True`` are both set.
+        If ``track_lineage=True`` and the input frame already contains a column
+        named ``__arnio_lineage_id__``.
         If step format is invalid.
     UnknownStepError
         If step name is not registered.
+    PipelineStepError
+        If a custom step removes the internal lineage sentinel column while
+        ``track_lineage=True`` is active.
 
     Examples
     --------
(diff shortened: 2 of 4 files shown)


dynamo-pentester:
Hi @im-anishraj , CI looks Green. Please let me know if any changes needed ! Waiting for your approval

im-anishraj (changes requested):
@dynamo-pentester thanks for the solid first pass on #648. I checked the latest head against current `main`; the CI matrix is green apart from the Vercel authorization item, which is not a blocker. I do need a few correctness fixes before this can merge:

1. `track_lineage=True` currently bypasses `return_metadata=True`. The docstring says users can pass `return_metadata=True` separately for timings, but the lineage return path returns `(result, LineageReport)` before metadata is returned. Please either support and test a clear combined return shape, or reject the combination with a clear error and update the Python + website docs accordingly.

2. Repeated step names overwrite earlier drops because `dropped_by_step[name] = ...` assigns by step name on every iteration. A valid pipeline can include the same step more than once, so the report can lose rows dropped by the first invocation. Please accumulate repeated step names or use stable per-step keys, and add a regression test with the same row-dropping step run twice.

3. Sentinel handling needs an explicit user-facing contract. If the input already has `__arnio_lineage_id__`, `DataFrame.insert` will raise a raw pandas collision error; if a custom step removes the sentinel, the later lookup raises a raw `KeyError`. Please guard these cases with clear Arnio errors or use a collision-safe internal marker, and add tests.

Static checks, formatting checks, and the lightweight website/docs checks passed locally. I’m holding the PR until the API behavior above is made deterministic and documented.

dynamo-pentester:
Hi @im-anishraj , I have made changes and once locally tested. All test cases passes. No lint errors.

im-anishraj (commented):
@dynamo-pentester thanks for the quick update. I rechecked latest head `fc30e054` against current `main`.

The three requested areas look addressed on static review: `track_lineage=True` now rejects `return_metadata=True`, repeated step names accumulate drops instead of overwriting, and sentinel collision/removal paths now have explicit errors and tests. Local lightweight checks passed for the simulated merge:

```text
git diff --check HEAD
ruff check arnio/pipeline.py tests/test_pipeline_lineage.py arnio/__init__.py
black --check arnio/pipeline.py tests/test_pipeline_lineage.py arnio/__init__.py
pytest tests/test_website_api_docs.py::test_website_api_reference_mentions_public_exports tests/test_docs_utf8_check.py -q
```

I am not approving or merging yet because the full Actions matrix for this new head is still running. Also, I still cannot run the functional lineage pytest locally on this Windows checkout because the available `_arnio_cpp` binary is stale versus the current Python binding shape, so I will rely on the GitHub matrix for that part. Vercel authorization remains external noise and is not your blocker.

dynamo-pentester:
> @dynamo-pentester thanks for the quick update. I rechecked latest head `fc30e054` against current `main`.
> 
> The three requested areas look addressed on static review: `track_lineage=True` now rejects `return_metadata=True`, repeated step names accumulate drops instead of overwriting, and sentinel collision/removal paths now have explicit errors and tests. Local lightweight checks passed for the simulated merge:
> 
> ```
> git diff --check HEAD
> ruff check arnio/pipeline.py tests/test_pipeline_lineage.py arnio/__init__.py
> black --check arnio/pipeline.py tests/test_pipeline_lineage.py arnio/__init__.py
> pytest tests/test_website_api_docs.py::test_website_api_reference_mentions_public_exports tests/test_docs_utf8_check.py -q
> ```
> 
> I am not approving or merging yet because the full Actions matrix for this new head is still running. Also, I still cannot run the functional lineage pytest locally on this Windows checkout because the available `_arnio_cpp` binary is stale versus the current Python binding shape, so I will rely on the GitHub matrix for that part. Vercel authorization remains external noise and is not your blocker.

Thanks for taking another look.

I addressed the three requested items:

* `track_lineage=True` now has deterministic behavior when combined with `return_metadata=True`, with updated documentation and tests.
* Repeated step names now accumulate dropped rows instead of overwriting previous entries.
* Sentinel collision/removal paths now raise explicit Arnio errors and have dedicated regression tests.

I'll keep an eye on the Actions run and let me know if anything else shows up during CI.

dynamo-pentester:
Hi @im-anishraj , CI looks green. Waiting for your final approval !!

im-anishraj (approved):
@dynamo-pentester thanks for the quick fixes. The latest head `fc30e054` addresses the requested API/lineage correctness items, and the full non-Vercel Actions matrix is green.

Approving this now. Vercel remains external authorization noise and is not a blocker for this PR.

PR #2313, opened 2026-06-03
docs: add auto_clean dry-run and explain workflow examples

### LINKED ISSUE #2168: Docs: add website examples for auto_clean dry-run and explain workflows
﻿## Summary
The Docs page mentions quality reports and gates but does not show how to use `auto_clean()` safely with dry-run or explanation output.

## Why it matters
`auto_clean()` is powerful and safety-sensitive. Users should see an audit-first workflow before applying automatic changes to real data.

## Evidence
- `website/docs.html:150-160` shows `profile()`, report exports, and `check_quality_gates()`.
- `arnio/quality.py:2287-2295` exposes `dry_run`, `return_report`, `allow_lossy_casts`, `confirmed_casts`, and `explain`.
- `website/api.html:90` mentions `auto_clean()` but not a tutorial flow.

## Reproduction or trigger
1. Open `https://arniolib.vercel.app/docs.html`.
2. Review the Quality Reports section.
3. Look for an `auto_clean(..., dry_run=True)` or `explain=True` example.

## Expected behavior
The docs should demonstrate a safe dry-run/explain workflow.

## Actual behavior
No tutorial example is present.

## Suggested direction
Add a short example showing dry-run, inspecting suggestions/explanations, and then applying cleaning intentionally.

AaryanInzalkar:
## Summary
Adds a safe `auto_clean()` workflow example to the Quality Reports section of `website/docs.html`.

## Changes
- Added a 3-step code example showing `dry_run=True`, `explain=True`, and intentional application with `confirmed_casts`

## Related Issue
fixes #2168

### DIFF
--- a/website/docs.html
+++ b/website/docs.html
@@ -180,17 +180,19 @@ gate = ar.check_quality_gates(baseline, current)</code></pre></div>
         <h3>Safe auto_clean Workflow</h3>
         <p><code>auto_clean()</code> is powerful and safety-sensitive. Use <code>dry_run=True</code> and <code>explain=True</code> to audit changes before applying them to real data.</p>
         <div class="code-block"><div class="code-block-header">Python</div><pre><code class="language-python"># Step 1 — Preview what would change, without touching the data
-report = ar.auto_clean(frame, dry_run=True, return_report=True)
-print(report)
+# Step 1 — Preview what would change, without touching the data
+report = ar.auto_clean(frame, mode="strict", dry_run=True)
+casts = report.confirmed_casts
 
-# Step 2 — Understand why each change is suggested
-report = ar.auto_clean(frame, dry_run=True, explain=True, return_report=True)
+# Step 2 — Inspect explanations separately (cannot combine with dry_run)
+clean, explanation = ar.auto_clean(frame, explain=True)
 
-# Step 3 — Apply intentionally once you have reviewed the suggestions
+# Step 3 — Apply intentionally using the previewed casts
 clean = ar.auto_clean(
     frame,
-    allow_lossy_casts=False,
-    confirmed_casts={"price": "FLOAT64"},
+    mode="strict",
+    allow_lossy_casts=True,
+    confirmed_casts=casts,
 )
 df = ar.to_pandas(clean)</code></pre></div>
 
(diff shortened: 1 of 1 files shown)


im-anishraj (changes requested):
Hi @AaryanInzalkar, thanks for adding this workflow example. I can't merge this one yet because the new snippet documents option combinations that the current API intentionally rejects:

- `ar.auto_clean(frame, dry_run=True, return_report=True)` raises `ValueError: return_report=True cannot be used with dry_run=True`.
- `ar.auto_clean(frame, dry_run=True, explain=True, ...)` raises because `explain=True` cannot be used with `dry_run=True`.
- The final apply step passes `confirmed_casts` while leaving `mode` at the default `"safe"` and `allow_lossy_casts=False`, so it does not actually demonstrate applying the previewed strict casts.

Could you please update the example to match the supported workflow? A good shape would be: preview with `report = ar.auto_clean(frame, mode="strict", dry_run=True)`, inspect suggestions/casts from that report, then apply with `ar.auto_clean(frame, mode="strict", allow_lossy_casts=True, confirmed_casts=casts)`. If you want to show explanations, use a separate non-dry-run example such as `clean, explanation = ar.auto_clean(frame, explain=True)`.

I verified this against `arnio/quality.py` and the existing regression coverage around invalid `dry_run` combinations. Once the snippet is corrected, we can re-run the lightweight docs/website checks.

im-anishraj (changes requested):
Hi @AaryanInzalkar, thanks for the quick update. This is closer, but one API mismatch remains before we can merge:

- `DataQualityReport` does not expose `report.confirmed_casts`. The supported way to pull the previewed strict cast mapping is `casts = dict(report.suggestions).get("cast_types")`, as used in the existing docs/tests.
- The snippet currently repeats the `# Step 1 — Preview...` comment twice.

Could you please change that line to something like:

```python
report = ar.auto_clean(frame, mode="strict", dry_run=True)
casts = dict(report.suggestions).get("cast_types")
```

Then the final apply step with `mode="strict"`, `allow_lossy_casts=True`, and `confirmed_casts=casts` will match the current API. The CI/docs smoke checks are otherwise looking good apart from the expected Vercel auth failure, but I want the example itself to be accurate before approval.

im-anishraj (approved):
Thanks @AaryanInzalkar, the latest update resolves the remaining docs issue. The example now uses the supported strict dry-run preview, extracts casts from report.suggestions, keeps explain separate from dry_run, and applies the confirmed strict cast mapping intentionally. I verified this locally with git diff --check and the docs/website pytest checks.

PR #2338, opened 2026-06-04
Replace std::stod with parse_float64_classic in cleaning.cpp

### LINKED ISSUE #1989: Bug: cast_types FLOAT64 conversion is locale-sensitive while CSV parsing is locale-independent
### What happened?

CSV parsing uses locale::classic() during floating-point parsing.

However cast_types() uses std::stod(), which follows the active process locale.

As a result, values that were successfully ingested from CSV may later fail during FLOAT64 casting on systems using non-English numeric locales.

### Area

### Minimal reproduction

import locale
import arnio as ar

locale.setlocale(
    locale.LC_NUMERIC,
    "de_DE.UTF-8"
)

frame = ar.from_dict({
    "price": ["1.5", "2.5"]
})

ar.cast_types(
    frame,
    {"price": "float64"}
)

### Expected behavior

CSV ingestion and type conversion should use identical numeric parsing semantics regardless of system locale.

A value successfully parsed during CSV ingestion should also be convertible through cast_types().

### Actual behavior

On systems using locales such as:

de_DE.UTF-8
fr_FR.UTF-8

values containing decimal points can behave differently between CSV ingestion and cast_types() conversion.

The two code paths use different parsing implementations.

### Impact

Produces incorrect data

### Arnio version

1.1.0

### Install method

Source inspection

### Operating System

Linux

### Python version

3.12.4

### Environment details

```shell
Issue identified through source inspection of CSV parsing and FLOAT64 casting implementations.

The report is based on analysis of locale::classic() parsing versus std::stod() locale-dependent behavior.
```

### Pre-flight checks

### Additional context

_No response_

dynamo-pentester:
## Summary

Fixes locale-sensitive `FLOAT64` parsing in `cast_types()` and `fill_nulls()`.

Closes #1989

---

## Root cause

Two call sites in `cpp/src/cleaning.cpp` used `std::stod()` to parse string values into `double`. `std::stod()` follows the active process locale — on systems using `de_DE.UTF-8` or `fr_FR.UTF-8`, it treats `.` as a thousands separator rather than a decimal point. This means values like `"1.5"` and `"3.14"` that were ingested correctly during CSV reading (which uses `locale::classic()`) would silently fail or produce wrong results at cast time.

**Affected paths:**
- `coerce_value()` — the `fill_nulls()` FLOAT64 string fill-value path (~line 209)
- `cast_types()` — the FLOAT64 column casting path (~line 575)

---

## Fix

Added a single `static bool parse_float64_classic(const std::string& s, double& out)` helper in `cpp/src/cleaning.cpp`, defined before both call sites. It mirrors the exact parsing contract already used by `csv_reader.cpp`'s `try_parse_float64`:

- Handles `"nan"`, `"inf"`, `"+inf"`, `"-inf"` explicitly (platform `istringstream` behaviour for special tokens varies)
- Parses all other values with `std::istringstream` imbued with `std::locale::classic()`, which always uses `.` as the decimal separator

Both `std::stod()` call sites are replaced with `parse_float64_classic()`. Existing error contracts are preserved exactly:
- `fill_nulls`: non-finite fill values still throw `std::invalid_argument`
- `cast_types`: non-finite and unparseable values still throw `cast_error`; `coerce_invalid=true` still produces null

---

## Files changed

| File | Change |
|---|---|
| `cpp/src/cleaning.cpp` | Add `parse_float64_classic()` helper; replace 2 `std::stod` call sites |
| `tests/test_cleaning.py` | Add 12 regression tests |

No Python API changes. No C++ header changes. No other files touched.

---

## Tests

12 new regression tests appended to `tests/test_cleaning.py`:

**`TestCastTypesFloat64LocaleIndependent`**

| Test | What it covers |
|---|---|
| `test_cast_float64_succeeds_under_de_locale` | Core regression: `"1.5"` → `1.5` under `de_DE.UTF-8` |
| `test_cast_float64_succeeds_under_fr_locale` | Same under `fr_FR.UTF-8` |
| `test_cast_float64_negative_value_under_de_locale` | Negative floats survive locale change |
| `test_cast_float64_scientific_notation_under_de_locale` | Scientific notation survives locale change |
| `test_cast_float64_coerce_invalid_under_de_locale` | `coerce_invalid=True` still nulls genuinely bad values |
| `test_cast_float64_special_tokens_unchanged_under_de_locale` | `nan`/`inf`/`-inf` → null under `coerce_invalid=True` |
| `test_cast_float64_baseline_locale_independent` | Baseline: default locale still works |
| `test_cast_float64_invalid_value_raises_under_de_locale` | Genuinely invalid value still raises |

**`TestFillNullsFloat64LocaleIndependent`**

| Test | What it covers |
|---|---|
| `test_fill_nulls_float64_fill_value_under_de_locale` | String fill value `"0.5"` accepted under `de_DE.UTF-8` |
| `test_fill_nulls_float64_fill_value_under_fr_locale` | Same under `fr_FR.UTF-8` |
| `test_fill_nulls_float64_fill_value_baseline` | Baseline: default locale still works |
| `test_fill_nulls_float64_non_finite_fill_rejected_under_de_locale` | Non-finite fill still rejected under non-English locale |

All locale-dependent tests call `_set_locale()` which wraps `locale.setlocale` in a `try/except locale.Error` and calls `pytest.skip` when the locale is unavailable on the runner.

---

## What was not changed

- `cpp/src/csv_reader.cpp` — already locale-independent, untouched
- Python API — untouched
- Any pipeline, schema, quality, or packaging code
- All existing `cast_types`, `fill_nulls`, and broader cleaning tests continue to pass

### DIFF
--- a/cpp/src/cleaning.cpp
+++ b/cpp/src/cleaning.cpp
@@ -151,6 +151,47 @@ static int64_t checked_double_to_int64(double value, const char* context) {
     return static_cast<int64_t>(value);
 }
 
+// Locale-independent floating-point parser.
+//
+// Mirrors the parsing contract used by csv_reader.cpp (try_parse_float64) so
+// that values successfully ingested during CSV reading can always be cast later
+// regardless of the active process locale.
+//
+// Special tokens ("nan", "inf", "+inf", "-inf") are handled explicitly because
+// std::istringstream behaviour for these varies across platforms and locales.
+// All other values are parsed with std::istringstream imbued with
+// std::locale::classic(), which uses '.' as the decimal separator — the same
+// separator the CSV reader expects.
+static bool parse_float64_classic(const std::string& s, double& out) {
+    if (s.empty()) return false;
+
+    // Build a lowercase copy for special-token matching only.
+    std::string lower = s;
+    std::transform(lower.begin(), lower.end(), lower.begin(),
+                   [](unsigned char c) { return static_cast<char>(std::tolower(c)); });
+
+    if (lower == "nan") {
+        out = std::numeric_limits<double>::quiet_NaN();
+        return true;
+    }
+    if (lower == "inf" || lower == "+inf") {
+        out = std::numeric_limits<double>::infinity();
+        return true;
+    }
+    if (lower == "-inf") {
+        out = -std::numeric_limits<double>::infinity();
+        return true;
+    }
+
+    std::istringstream iss(s);
+    iss.imbue(std::locale::classic());
+    double val = 0.0;
+    iss >> val;
+    if (iss.fail() || !iss.eof()) return false;
+    out = val;
+    return true;
+}
+
 static CellValue coerce_value(const CellValue& value, DType target) {
     if (std::holds_alternative<std::monostate>(value)) {
         return std::monostate{};
@@ -204,15 +245,13 @@ static CellValue coerce_value(const CellValue& value, DType target) {
         if (std::holds_alternative<bool>(value)) return std::get<bool>(value) ? 1.0 : 0.0;
         if (std::holds_alternative<std::string>(value)) {
             const auto& s = std::get<std::string>(value);
-            try {
-                size_t pos = 0;
-                double parsed = std::stod(s, &pos);
+            double parsed = 0.0;
+            if (parse_float64_classic(s, parsed)) {
                 if (std::isnan(parsed) || std::isinf(parsed)) {
                     throw std::invalid_argument(
                         "Non-finite numeric fill values are not permitted for float columns.");
                 }
-                if (pos == s.size()) return parsed;
-            } catch (...) {
+                return parsed;
             }
         }
     }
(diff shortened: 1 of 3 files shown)


dynamo-pentester:
Hi @im-anishraj , there was a lint error on tests/test_quality.py

image

done changing it and pushed the code. Please let me know if any changes needed !

dynamo-pentester:
Hi @im-anishraj , CI Looks green. Waiting for your approval !!

dynamo-pentester:
Whoa bruh @im-anishraj you are back , i was waiting for you 😁

im-anishraj (approved):
Thanks @dynamo-pentester. I reviewed the diff and the current check rollup on `684158c`: the locale-independent parser is scoped to the two `std::stod` call sites, the regression coverage covers cast/fill behavior, and all non-Vercel checks are green. Vercel is external/auth-only here and is not a blocker.

PR #2389, opened 2026-06-05
fix: validate importable arnio core in check_env

### LINKED ISSUE #2349: Bug: check_env reports native core as available from stale site-packages artifact
## Summary

`examples/check_env.py` can report `Arnio Core Module: Available (C++ Accelerated)` even when `import arnio` fails from a source checkout because the local `arnio/` package does not contain `_arnio_cpp`.

## Reproduction

On Windows from the repository root:

```powershell
cd C:\Users\anish\Desktop\arnio
.\.audit_venv\Scripts\python.exe examples\check_env.py
.\.audit_venv\Scripts\python.exe -c "import arnio"
```

Observed `check_env.py` output:

```text
Arnio Core Module:  Available (C++ Accelerated)
...
Build Toolchain Status:
  - MSVC compiler   [X]  cl         (required)
  - NMake           [X]  nmake      (required)

[BUILD BLOCKED] Source installs like `pip install -e ".[dev]"` may fail until the missing build tools are available.
...
All optional dependencies are successfully installed! You are ready to go.
```

Observed import result:

```text
ImportError: arnio C++ extension (_arnio_cpp) not found.
```

## Root Cause

`detect_core_status()` scans every `sys.path` entry for `arnio/_arnio_cpp*`. In this environment it finds an installed artifact under:

```text
.audit_venv/Lib/site-packages/arnio/_arnio_cpp.cp313-win_amd64.pyd
```

However, running from the repository root imports the local source package at:

```text
C:\Users\anish\Desktop\arnio\arnio
```

That local package only has `_arnio_cpp.pyi`, not an importable `.pyd` extension.

## Expected Behavior

`check_env.py` should validate the native extension that will actually be imported from the current checkout. If the local source package shadows an installed wheel and does not contain `_arnio_cpp`, the dashboard should report the core as missing or blocked.

## Suggested Fix

- Resolve the actual `arnio` package path that Python would import.
- Check for `_arnio_cpp` only under that resolved package path, or attempt a narrow import of `arnio._core` and report the real error.
- If required build tools are missing, do not print `You are ready to go.`
- Exit non-zero when required local build prerequisites are missing for a source checkout.

## Acceptance Criteria

- `python examples/check_env.py` does not report the native core as available when the current checkout cannot import `_arnio_cpp`.
- The final summary distinguishes optional dependency readiness from native build readiness.
- The script exits non-zero when required native build tools are missing in a source checkout.
- Add a focused test or helper-level coverage for source-checkout shadowing of an installed wheel.

karthikeya20012007:
## Summary

Fixes incorrect native-core detection in `examples/check_env.py` when a source checkout shadows an installed wheel containing `_arnio_cpp`.

## Changes

* resolve the actually importable `arnio` package path using `importlib.util.find_spec`
* validate `_arnio_cpp` only under the resolved package path instead of scanning arbitrary `sys.path` entries
* separate optional dependency readiness from native-core readiness messaging
* return a non-zero exit code when the native core is unavailable and required build tools are missing
* add regression coverage for source-checkout shadowing of an installed wheel

Fixes #2349

### DIFF
--- a/examples/check_env.py
+++ b/examples/check_env.py
@@ -253,7 +253,11 @@ def main():
 
     _, core_available = detect_core_status()
 
-    if not core_available and not build_tools["all_required_found"]:
+    required_build_tools_available = all(
+        check.available for check in build_tools if check.required
+    )
+
+    if not core_available and not required_build_tools_available:
         sys.exit(1)
 
 
--- a/examples/check_env.py
+++ b/examples/check_env.py
@@ -131,20 +131,24 @@ def check_build_tools():
 
 
 def detect_core_status():
-    """Detect the native extension without importing the full arnio package."""
+    """Detect the native extension for the actually importable arnio package."""
+    import importlib.util
+
     if "arnio._core" in sys.modules:
         if sys.modules["arnio._core"] is None:
             return "Not Compiled (arnio._core unavailable)", False
         return "Available (C++ Accelerated)", True
 
-    for entry in sys.path:
-        package_dir = Path(entry or ".") / "arnio"
-        if not package_dir.is_dir():
-            continue
+    spec = importlib.util.find_spec("arnio")
 
-        for suffix in EXTENSION_SUFFIXES:
-            if (package_dir / f"_arnio_cpp{suffix}").exists():
-                return "Available (C++ Accelerated)", True
+    if spec is None or spec.origin is None:
+        return "Not Installed (arnio package not found)", False
+
+    package_dir = Path(spec.origin).parent
+
+    for suffix in EXTENSION_SUFFIXES:
+        if (package_dir / f"_arnio_cpp{suffix}").exists():
+            return "Available (C++ Accelerated)", True
 
     return "Not Compiled (_arnio_cpp extension file not found)", False
 
@@ -229,15 +233,28 @@ def print_dashboard(results, build_tools=None):
         print("\n[TIP] To install all missing optional dependencies, run:")
         print(f"  pip install {' '.join(missing)}")
     else:
+        print("\nAll optional dependencies are successfully installed!")
+
+    if core_available:
+        print("Native Arnio core is available. You are ready to go.")
+    else:
         print(
-            "\nAll optional dependencies are successfully installed! You are ready to go."
+            "Native Arnio core is not available. "
+            "Build the extension before running Arnio examples locally."
         )
     print("=" * 70)
 
 
 def main():
     results = check_dependencies()
-    print_dashboard(results)
+    build_tools = check_build_tools()
+
+    print_dashboard(results, build_tools)
+
+    _, core_available = detect_core_status()
+
+    if not core_available and not build_tools["all_required_found"]:
+        sys.exit(1)
 
 
 if __name__ == "__main__":
--- a/tests/test_check_env.py
+++ b/tests/test_check_env.py
@@ -155,9 +155,12 @@ def test_detect_core_status_uses_importable_package_path(
 
     (installed_package / "_arnio_cpp.pyd").write_text("compiled extension")
 
-    with patch(
-        "importlib.util.find_spec",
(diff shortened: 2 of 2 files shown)


im-anishraj (changes requested):
Hi @karthikeya20012007, thanks for the quick fix on this one. The direction looks right and the focused local checks are clean, but I need to hold this for a test/CI failure before approval.

What I verified locally:

```text
ruff check examples/check_env.py tests/test_check_env.py
ruff format --check examples/check_env.py tests/test_check_env.py
pytest tests/test_check_env.py -q  # 8 passed
```

The full GitHub test matrix is failing on a non-Vercel check, though. The failure is order-dependent: once the wider suite has already imported `arnio._core`, `test_detect_core_status_uses_importable_package_path` gets `available=True` from the existing `sys.modules` shortcut and fails this assertion:

```text
FAILED tests/test_check_env.py::test_detect_core_status_uses_importable_package_path
E   assert True is False
```

I could reproduce it locally with:

```text
python -c "import arnio._core, pytest; raise SystemExit(pytest.main(['tests/test_check_env.py::test_detect_core_status_uses_importable_package_path','-q']))"
```

Please update the regression test to isolate/clear the relevant `sys.modules` entries, or adjust `detect_core_status()` so the mocked importable package path is still respected in this scenario. Once the non-Vercel CI is green again, this should be ready for another look.

karthikeya20012007:
Thanks for the detailed repro steps — I was able to reproduce the order-dependent failure locally using the provided command.

The regression test has now been updated to isolate the relevant `sys.modules` state so it no longer reuses a previously imported `arnio._core` instance from the wider suite.

I verified locally with:

* `pytest tests/test_check_env.py -q`
* `python -c "import arnio._core, pytest; raise SystemExit(pytest.main(['tests/test_check_env.py::test_detect_core_status_uses_importable_package_path','-q']))"`
* `ruff check .`

The branch has been updated.

im-anishraj (changes requested):
Hi @karthikeya20012007, thanks for the quick update. The previous order-dependent test failure is fixed now, and I reran the focused checks on head `9515660`:

```text
ruff check examples/check_env.py tests/test_check_env.py
ruff format --check examples/check_env.py tests/test_check_env.py
pytest tests/test_check_env.py -q  # 8 passed
python -c "import arnio._core, pytest; raise SystemExit(pytest.main(['tests/test_check_env.py::test_detect_core_status_uses_importable_package_path','-q']))"  # passed
```

I found one remaining runtime blocker in `main()`. `check_build_tools()` returns a list of `BuildToolCheck` objects, but the new exit-code branch indexes it like a dict:

```python
if not core_available and not build_tools["all_required_found"]:
```

That raises a `TypeError` instead of exiting cleanly when the native core is unavailable. I reproduced the failure path locally with a mocked missing-core environment:

```text
TypeError: list indices must be integers or slices, not str
```

Please change that condition to derive readiness from the list, for example using `all(check.available for check in build_tools if check.required)`, and add a small `main()`/helper-level test for the missing-core + missing-build-tool path so this stays covered. Once that is fixed and non-Vercel CI is green, this should be close.

karthikeya20012007:
Thanks for catching the runtime issue.

I updated the exit-code logic to derive required build-tool readiness from the returned `BuildToolCheck` objects instead of indexing the list as a dict, and added focused coverage for the missing-core + missing-required-build-tools exit path.

Reverified locally with:

* `pytest tests/test_check_env.py -q`
* `ruff check examples/check_env.py tests/test_check_env.py`

The branch has been updated.

im-anishraj (approved):
@karthikeya20012007 thanks for the quick follow-up. I rechecked the current head `87cf40a` and the latest check set: the prior order-dependent regression and the missing-build-tools exit path are covered now, and all non-Vercel checks are green.

Approved from maintainer review. The remaining Vercel authorization failure is external and not a contributor-side blocker, so no further changes are needed unless requested.

PR #2401, opened 2026-06-05
feat: add collapse_rare_categories cleaning step

### LINKED ISSUE #1295: Feature: add collapse_rare_categories cleaning step
### Problem or motivation

High-cardinality categorical columns often contain a long tail of very rare text labels. When these columns are eventually handed off to models, these rare categories:
1. create a massive explosion of sparse columns, increasing memory consumption.
2. introduce severe noise, forcing ML models to overfit on rare anomalies

Currently, Arnio has robust methods for basic string sanitization like `strip_whitespace` and `normalize_case`, but lacks a high-performance feature-engineering step to truncate low-frequency categorical noise at the core data-preparation layer before `pandas`/`numpy` serialization.

### Proposed solution

Introduce a `collapse_rare_categories` capability implemented in the C++ backend layer for maximum speed, exposed via the Pybind11 module down to the user-facing arnio.cleaning Python API surface.

#### Public API Design:
```python
def collapse_rare_categories(
    frame: ArFrame,
    column: str,
    threshold: float = 0.02,
    fill_value: str = "Other",
) -> ArFrame:
    """Group rare string categories into a single unified label based on frequency."""
```
#### Implementation Details:
1. C++ Engine: Scans the target DType::STRING column using an efficient std::unordered_map<std::string, size_t> to compute total string frequency distributions in $O(N)$ time.
2. Memory Optimization: Rebuilds only the modified target column while preserving all other unchanged columns via Arnio's underlying $O(1)$ move_clone() structural pointer transfer strategy.
3. GIL Management: Releases the Python Global Interpreter Lock (py::gil_scoped_release) during the frequency computation and matrix modification sweep for true parallel backend execution.

### User story

As an aspiring ML Engineer, while preparing many datasets I often encounter columns like City or Occupation with hundreds of values where 95% of rows belong to the top 5 classes, and the remaining 5% are unique entries.

We need a way to instantly sweep the data at the ingestion layer and group all low-frequency entries into an "Other" class (or any other user-defined custom name) before running encoding loops, ensuring the downstream `numpy` pipelines remain memory-lean and mathematically stable.

### Acceptance criteria

1. Determinism: The behavior must be perfectly reproducible using relative mathematical proportion weights strictly bounded by a threshold domain of $[0.0, 1.0]$.
2. Type Safety: The core engine throws a clear, documented std::invalid_argument if invoked on columns that are not of type DType::STRING.
3. Null Preservation: Existing internal null/missing values mask vectors must pass through completely untouched and uncounted within the tracking sweep.
4. Performance: Modifying a high-cardinality string column must pass execution to the multi-threaded native backend without holding the Python GIL layer captive.
5. Regression Coverage: Unit tests verify handling of edge cases (completely unique values, zero-row empty dataframes, threshold boundary values like 0.0 or 1.0).

### Alternatives considered

_No response_

### Scope

API design

### Expected difficulty

Intermediate

Yes - I'd like to submit a PR

### Pre-flight checks

### Additional context

i would like to work on this issue under GSSoC'26.

taniy8:
## Summary

Fixes #1295

Implements `collapse_rare_categories` as requested in #1295. Rare string categories in high-cardinality columns are grouped into a single unified label based on relative frequency, reducing downstream memory consumption and model noise before pandas/numpy serialization.

### C++ engine (`cpp/src/cleaning.cpp`, `cpp/include/arnio/cleaning.h`)

- Added a single O(N) frequency pass using `std::unordered_map<std::string, size_t>`
- Throws `std::invalid_argument` for non-`STRING` columns
- Null values are skipped during frequency counting and passed through unchanged

### Pybind11 binding (`bindings/bind_arnio.cpp`)

- Exposed the operation with `py::gil_scoped_release` during the scan and rebuild phases
- Added defaults:
  - `threshold=0.02`
  - `fill_value="Other"`

### Python API (`arnio/cleaning.py`, `arnio/_core.py`, `arnio/__init__.py`)

- Added public API entry point
- Added validation, type annotations, and docstrings

### Tests (`tests/test_cleaning.py`)

- Added 13 tests covering all acceptance criteria

### Documentation (`README.md`)

- Added `collapse_rare_categories` to the cleaning functions table

## Acceptance Criteria Checklist

- [x] Determinism - pure proportion math bounded to `[0.0, 1.0]`, verified in `test_determinism`
- [x] Type safety - C++ throws `std::invalid_argument` for non-`STRING` columns
- [x] Null preservation - nulls are excluded from frequency counts and preserved in output
- [x] GIL released - `py::gil_scoped_release` used in the binding
- [x] Edge cases covered - empty frame, all-unique values, threshold boundaries (`0.0` and `1.0`), and type validation errors

## Testing

- 509 passed in 4.79s


im-anishraj (changes requested):
Hi @taniy8, thanks for the substantial implementation. I added the GSSoC/C++ cleaning classification labels and reviewed the CI signal.

This cannot be approved yet because the full non-Vercel Python test matrix is failing. The failing check is consistent across platforms and points to the new public export missing from the website API reference:

```text
FAILED tests/test_website_api_docs.py::test_website_api_reference_mentions_public_exports
AssertionError: assert ['collapse_rare_categories'] == []
Left contains one more item: 'collapse_rare_categories'
```

Since this PR adds `collapse_rare_categories` to the public API, please update the website API reference so the docs mention the new public export. After that, rerun/push and let CI go green. Vercel authorization is external noise; the Python test matrix failure above is the blocker.

im-anishraj (changes requested):
Hi @taniy8, thanks for the quick update adding the website API reference. I rechecked the new head (`7373b891`) and the docs drift check is now passing locally.

This still cannot be approved yet because the non-Vercel Windows Python matrix is failing on the new `TestCollapseRareCategories` cases. The failures show the output category values becoming empty strings (`""`) instead of values like `NYC`, `LA`, `Other`, or the custom fill value.

The likely root cause is in `cpp/src/cleaning.cpp`: `const auto& val = std::get<std::string>(col.at(r));` binds a reference to a value inside the temporary `CellValue` returned by `Column::at()`. That reference is dangling, which explains why this is platform-sensitive. Please store the cell/string by value before using it, or otherwise read from stable storage.

One more scope item: please also wire `collapse_rare_categories` into the built-in pipeline registry in `arnio/pipeline.py` and add a focused pipeline test. The issue scope and website API page both describe cleaning functions as pipeline-usable, so the public API should be consistent there too.

Checks I ran locally:
- `ruff check arnio/cleaning.py arnio/_core.py arnio/__init__.py tests/test_cleaning.py`
- `ruff format --check arnio/cleaning.py arnio/_core.py arnio/__init__.py tests/test_cleaning.py`
- `pytest tests/test_website_api_docs.py::test_website_api_reference_mentions_public_exports -q`

I could not run the native functional tests locally because this Windows shell does not have `nmake`/MSVC available for an editable C++ build, so the GitHub Actions Windows failures are the source of truth here.

taniy8:
Thanks for the detailed review! All three items have been addressed in the latest commit:

- **C++ dangling reference:** Stored `col.at(r)` by value before extracting the string, fixing the empty-string outputs observed on Windows.
- **Pipeline registry:** Added `collapse_rare_categories` to `_STEP_REGISTRY`.
- **Pipeline test:** Added `test_pipeline_collapse_rare_categories` to `TestPipeline`.

CI is now green. Ready for another look! @im-anishraj

im-anishraj:
Hi @taniy8, thanks for the follow-up. The previous implementation blockers appear addressed now: the docs/API reference is updated, the pipeline registry/test was added, and the full non-Vercel CI matrix is green. I also rechecked the local static side with `git diff --check`, `ruff check`, and `ruff format --check`, and those passed.

I am not approving or merging from the automation yet because this is a C++ core/bindings change and I cannot run the native targeted tests locally from this Windows shell: MSVC/NMake is unavailable, and the fallback MinGW toolchain here is too old for the repository's C++17 code. I moved this from `status:blocked` to `status:needs-maintainer` so it is clear the remaining step is maintainer/native verification rather than another contributor fix.

im-anishraj:
@taniy8, thank you for addressing the earlier implementation and CI feedback. The maintainer decision is that the code can return to final native review after the branch is current, but GitHub now reports merge conflicts with `main`.

Please rebase or merge the latest `main`, resolve the conflicts carefully, and let the full non-Vercel CI matrix run on the resulting head. Because this touches the C++ cleaning engine and public API, I will re-check the resolved diff rather than approving the stale green run. No additional feature work is requested at this point.

taniy8:
Rebased cleanly onto upstream/main. The branch now has exactly 3 commits on top of the current head. Conflicts in `website/api.html`, `arnio/__init__.py`, and `bindings/bind_arnio.cpp` were resolved keeping all upstream additions (`normalize_minmax`, `encode_categorical`) alongside the changes. CI is green on the new head `91714b9`. Ready for your re-check. @im-anishraj

im-anishraj (changes requested):
@taniy8 thanks for the rebase and cleanup. I rechecked head `91714b9`; the non-Vercel matrix is green, but I still need changes before approving.

Current blockers:
1. The public docs/comments say categories with frequency **strictly below** `threshold` are collapsed, but the C++ implementation uses `proportion <= threshold`. Please either change the implementation to `< threshold` or update the Python docstring, C++ header comment, README/API wording, and tests to clearly define "at or below" behavior. The current contract mismatch will surprise users at exact-boundary frequencies.
2. `threshold=float("nan")` currently passes the Python validation because comparisons against NaN are false, then reaches C++ and silently collapses nothing. Please reject non-finite numeric thresholds (`NaN`, `inf`, `-inf`) with a clear `ValueError` and add regression tests.
3. `bindings/bind_arnio.cpp` is still missing the final newline in the current diff.

The Vercel authorization failure is external and can be ignored once these are fixed and the non-Vercel checks stay green.

taniy8:
The `build_wheels` failure is a Docker unavailability error in the runner (`docker create` exited with code 1). `build_wheels.yml` is configured as `workflow_call`/`workflow_dispatch` only and isn't meant to run on PRs. This appears to be an infrastructure issue unrelated to these changes.

im-anishraj:
@taniy8 I checked the current head `b45bffd`. The three review blockers are addressed: the code/docs now use strict `< threshold` semantics consistently, non-finite thresholds are rejected before native dispatch, and the newline issue is resolved. I also checked the manylinux failure and agree it is infrastructure: Docker could not pull `quay.io/pypa/manylinux_2_28_x86_64:2026.03.20-1` because `https://quay.io/v2/` timed out. I reran the failed manylinux job; please wait for that result. Vercel is external.

im-anishraj (approved):
Rechecked current head b45bffd after the infrastructure rerun. The previously requested strict threshold semantics, non-finite validation, and newline fix are present, and the full non-Vercel matrix is now green, including Linux, Windows, and macOS native tests plus the rerun manylinux wheel smoke test. Vercel authorization is external to this change. Approved.

PR #2479, opened 2026-06-06
Fix read_jsonl_chunked path type validation

### LINKED ISSUE #2458: Bug: read_jsonl_chunked does not validate path types before os.fspath
## Environment / context
- Tested by source inspection on current `main`
- Related published-wheel observation: `arnio==1.19.0` has older JSONL behavior, but this issue is specifically about the current `read_jsonl_chunked()` implementation
- Affected area: JSONL streaming reader / argument validation

## Steps to reproduce
On current `main`, inspect `arnio/io.py`:
```python
def read_jsonl(...):
    if not isinstance(path, (str, os.PathLike)):
        raise TypeError(
            f"read_jsonl expected a filesystem path, got {type(path).__name__!r}"
        )
    path = os.fspath(path)


def read_jsonl_chunked(...):
    path = os.fspath(path)
    ...
```

Then call the streaming reader with an invalid path type:
```python
import arnio as ar

list(ar.read_jsonl_chunked(123))
list(ar.read_jsonl_chunked(["events.jsonl"]))
```

## Expected result
`read_jsonl_chunked()` should mirror `read_jsonl()` and raise a clear Arnio-specific `TypeError`, for example:
```text
read_jsonl_chunked expected a filesystem path, got 'int'
```

## Actual result
The current implementation calls `os.fspath(path)` before validating the type. Invalid inputs therefore surface Python's raw `TypeError` from `os.fspath`, and the message does not identify `read_jsonl_chunked()` or match the clearer `read_jsonl()` contract.

## Severity / impact
Low. This is not a data corruption issue, but it creates inconsistent public API behavior and weaker diagnostics for streaming JSONL workflows.

## Likely affected area
- `arnio/io.py::read_jsonl_chunked`
- `tests/test_jsonl.py`

## Suggested next action
Add an explicit `isinstance(path, (str, os.PathLike))` guard to `read_jsonl_chunked()` before `os.fspath(path)`, with a function-specific message. Add parameterized tests covering `int`, `float`, `None`, `object()`, and `list` inputs for both `read_jsonl()` and `read_jsonl_chunked()`.

LuisFigueroaG:
## Summary
- Validate read_jsonl_chunked path inputs before calling os.fspath
- Match the clearer TypeError contract already used by read_jsonl
- Add regression coverage for invalid path input types

Fixes #2458

## Tests
- uv run --extra dev pytest tests/test_jsonl.py -q
- uv run ruff check .
- uv run pytest -q

### DIFF
--- a/arnio/io.py
+++ b/arnio/io.py
@@ -1600,6 +1600,10 @@ def read_jsonl_chunked(
     """
     _validate_jsonl_encoding(encoding)
 
+    if not isinstance(path, (str, os.PathLike)):
+        raise TypeError(
+            f"read_jsonl_chunked expected a filesystem path, got {type(path).__name__!r}"
+        )
     path = os.fspath(path)
     encoding_errors = _validate_encoding_errors(encoding_errors)
     nrows = _validate_jsonl_nrows(nrows)
--- a/tests/test_jsonl.py
+++ b/tests/test_jsonl.py
@@ -596,3 +596,14 @@ class TestReadJsonlInvalidPathType:
             match="read_jsonl expected a filesystem path",
         ):
             ar.read_jsonl(bad_input)
+
+    @pytest.mark.parametrize(
+        "bad_input",
+        [123, 3.14, None, object(), ["a.jsonl"]],
+    )
+    def test_chunked_raises_type_error(self, bad_input):
+        with pytest.raises(
+            TypeError,
+            match="read_jsonl_chunked expected a filesystem path",
+        ):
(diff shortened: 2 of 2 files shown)


im-anishraj (approved):
@LuisFigueroaG, thank you for the focused validation fix. I reviewed the full diff and history, confirmed the only remote failure is the external Vercel authorization check, and ran the complete local JSONL test file against this exact head: 80 passed. The path validation is correctly placed before os.fspath() and preserves the existing generator behavior.

PR #2480, opened 2026-06-06
Validate write_parquet preserve_attrs type

### LINKED ISSUE #2459: Bug: write_parquet accepts non-boolean preserve_attrs values
## Environment / context
- Tested by current `main` source inspection in `arnio/io.py`
- Published `arnio==1.19.0` does not yet expose `preserve_attrs`; this issue targets the unreleased/current-main API documented on the website
- Affected area: Parquet export / metadata preservation

## Steps to reproduce
Current `write_parquet()` documents and types `preserve_attrs` as a boolean:
```python
def write_parquet(..., preserve_attrs: bool = True) -> None:
    ...
    if df.attrs:
        if preserve_attrs:
            ... validate attrs and write metadata ...
        else:
            df.attrs = {}
```

Callers can pass non-boolean values:
```python
import pandas as pd
import arnio as ar

src = pd.DataFrame({"x": [1, 2]})
src.attrs["source"] = "qa"
frame = ar.from_pandas(src)

ar.write_parquet(frame, "out.parquet", preserve_attrs="False")  # truthy, preserves attrs
ar.write_parquet(frame, "out.parquet", preserve_attrs=[])       # falsy, drops attrs
```

## Expected result
`preserve_attrs` should reject non-boolean values with a clear `TypeError`, similar to the explicit validation used for `compression` and `row_group_size`.

## Actual result
The current implementation relies on Python truthiness. Strings, integers, lists, dictionaries, and other objects can silently change whether metadata is preserved or dropped.

## Severity / impact
Low-to-medium. This can cause confusing metadata behavior in Parquet exports, especially in notebooks or pipelines where option values may come from parsed config files or environment variables.

## Likely affected area
- `arnio/io.py::write_parquet`
- `tests/test_parquet.py`
- Website/API docs only if the intended accepted type changes

## Suggested next action
Add an early guard:
```python
if not isinstance(preserve_attrs, bool):
    raise TypeError("preserve_attrs must be a bool")
```
Then add regression tests for `"False"`, `0`, `1`, `None`, `[]`, `{}`, and `object()`.

LuisFigueroaG:
## Summary
- Reject non-boolean preserve_attrs values before Parquet export
- Document the preserve_attrs TypeError contract
- Add regression coverage for common non-bool inputs

Fixes #2459

## Tests
- uv run --extra dev pytest tests/test_parquet.py -q
- uv run ruff format --check arnio/io.py tests/test_parquet.py
- uv run ruff check .
- uv run pytest -q

### DIFF
--- a/arnio/io.py
+++ b/arnio/io.py
@@ -1954,8 +1954,8 @@ def write_parquet(
     ImportError
         If ``pyarrow`` is not installed.
     TypeError
-        If ``preserve_attrs`` is ``True`` and ``DataFrame.attrs``
-        contains non-JSON-serializable values.
+        If ``preserve_attrs`` is not a boolean, or if ``preserve_attrs`` is
+        ``True`` and ``DataFrame.attrs`` contains non-JSON-serializable values.
     ValueError
         If the file extension is not ``.parquet`` or ``.pq``, if
         ``compression`` is not a recognised codec, or if
@@ -2000,6 +2000,9 @@ def write_parquet(
             raise TypeError("row_group_size must be an integer")
         if row_group_size <= 0:
             raise ValueError("row_group_size must be a positive integer")
+
+    if not isinstance(preserve_attrs, bool):
+        raise TypeError("preserve_attrs must be a bool")
 
     try:
         import pyarrow  # noqa: F401 — presence check only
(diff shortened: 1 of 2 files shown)


im-anishraj (approved):
@LuisFigueroaG, thank you for keeping this validation change narrowly scoped. I reviewed the full diff and history, confirmed the visible remote failure is Vercel authorization only, and ran the complete local Parquet test file against this exact head: 24 passed and 50 expected optional-pyarrow tests skipped. The bool validation correctly runs before the optional dependency import.

PR #2486, opened 2026-06-06
Feat/encode categorical

### LINKED ISSUE #1297: Feature: Add multi-column categorical encoding functions (encode_categorical)
### Problem or motivation

When preparing data, categorical text features must be converted into numerical matrices before mathematical formulas or model estimation loops can process them.

Currently, arnio lacks an efficient, native pipeline operation to apply vector categorical transformations like One-Hot Encoding and Ordinal Mapped Encoding over multi-column inputs. Forcing users to transition to pandas or scikit-learn just to parse string categories breaks the clean pipeline flow within the native execution layer.

### Proposed solution

Expose a new module function encode_categorical in the Python layer that calls an optimized, vector-based execution algorithm inside the C++ backend (_arnio_cpp).

### Python API:
```python 
def encode_categorical(
    frame: ArFrame,
    columns: Sequence[str],
    method: str = "one_hot",
    ordinal_mappings: Mapping[str, Mapping[str, int]] | None = None,
) -> ArFrame:
    """Encode multiple text columns into numbers for machine learning processing."""
```
### C++ Core Signature:
```cpp
Frame encode_one_hot_native(const Frame& frame, const std::vector<std::string>& column_names);
```
The operation will extract unique items deterministically, append the indicator columns with the layout template {column_name}_{category_value} filled with type tag DType::INT64 (values 0 or 1), and release the Python GIL during allocations to maximize optimization.

### User story

As a Machine Learning Student using `arnio` for data preprocessing, I want to convert multiple text categorical columns into zero-or-one indicator columns simultaneously, so that my feature extraction matrix is prepared cleanly for mathematical training pipelines without dropping into separate libraries prematurely.

### Acceptance criteria

1. Multi-Column Validation: Accepts a sequence of column strings and throws a KeyError via validate_columns_exist if any column name is invalid.
2. Type Protection: Throws an std::invalid_argument error if a user attempts to run one-hot transformations on a column type other than DType::STRING.
3. Deterministic Structure: The unique columns generated during One-Hot transformation must be sorted alphabetically to guarantee identical output structure across matching datasets.
4. Missing Value Resilience: Missing records/null spaces (std::monostate / None) are naturally parsed as all zeros across the indicator vectors, preventing evaluation bugs.
5. Memory Optimization: Uses native move semantics (std::move) when pushing the new vector allocations into the final Frame.

### Alternatives considered

- Using Scikit-Learn's OneHotEncoder
- Using Pandas pd.get_dummies

### Scope

API design

### Expected difficulty

Intermediate

Yes - I'd like to submit a PR

### Pre-flight checks

### Additional context

i would like to work on this issue under GSSoC'26 !

Naveen-Boddepalli:
## Summary
Adds `encode_categorical(frame, columns, method, ordinal_mappings)` — a new public API function that converts STRING columns into numerical representations for ML pipelines. Backed by two native C++ functions (`encode_one_hot_native`, `encode_ordinal_native`) exposed via pybind11. One-hot encoding appends sorted indicator columns (`{col}_{category}`) with INT64 dtype; ordinal encoding appends a mapped integer column (`{col}_ordinal`). GIL is released during C++ allocation. Null resilience: nulls produce all-zeros in one-hot and preserved nulls in ordinal.

## Linked issue
Fixes #1297 

## Type of change
- [ ] Bug fix
- [x] Feature
- [ ] Performance improvement
- [x] Documentation
- [x] Tests
- [ ] Refactor
- [ ] CI / packaging

## Area
- [ ] CSV parser / I/O
- [x] C++ cleaning engine
- [x] Python API / ArFrame
- [ ] Pipeline / custom steps
- [ ] Schema validation
- [ ] pandas interoperability
- [x] Docs / website
- [ ] Developer tooling

## Testing
```bash
pytest tests/test_encode_categorical.py -v  # 16 passed
pytest --tb=short -q                        # 3197 passed, 110 skipped
```
- [ ] `make lint`
- [x] `python -m pytest tests -v --tb=short -x`

## Screenshots / Media
image
image

## Performance Impact
GIL released during C++ column allocation. `std::move` semantics used when pushing indicator columns into the output frame. No benchmark regression observed.

## GSSoC labels
GSSoC '26 contribution.

## Contributor checklist

## Maintainer notes
- `cpp/CMakeLists.txt` updated to include `src/encode_categorical.cpp` in `arnio_core` static library.
- `arnio/_core.py` updated to import `_encode_one_hot_native` and `_encode_ordinal_native` from `_arnio_cpp`.
- `website/api.html` updated to document `encode_categorical` in the Cleaning table.


im-anishraj (changes requested):
@Naveen-Boddepalli, the core happy-path tests and CI are green, but this needs a focused correctness pass before approval:

1. Generated output names are not collision-safe. `Frame(std::vector<Column>)` permits duplicate names and its lookup index silently points to the last duplicate, so an existing `color_red`, repeated target columns, or two source/category combinations producing the same `{column}_{category}` name creates an ambiguous frame. Validate all planned one-hot/ordinal names against the original and each other before constructing the result, and add collision tests.
2. The ordinal mapping check is unreachable: the dict comprehension indexes `ordinal_mappings[col]` before the later `if col not in mappings`. A multi-column call with one missing mapping therefore leaks a raw `KeyError` instead of the intended deterministic validation error. Validate first and test this case.
3. `website/api.html` contains 976 lines of unrelated reformatting and a real navigation regression: the desktop nav drops Community/Changelog and leaves the Roadmap anchor outside an `<li>`. Revert the page churn and add only the focused `encode_categorical` documentation entry.
4. Add the native symbols to `arnio/_arnio_cpp.pyi` and the aliases/exports to `arnio/_core.pyi`; the runtime API and typing surface are currently out of sync.
5. Replace broad `pytest.raises(Exception)` assertions with the precise public exception types so the tests can catch regressions.

Please keep the next commit limited to these feature files and focused docs/tests.

im-anishraj (changes requested):
@Naveen-Boddepalli, thank you for the focused update. I verified that the five earlier blockers are fixed, all non-Vercel checks are green, `git diff --check` passes, the exact head builds successfully as a Windows native wheel, and all 22 focused `test_encode_categorical.py` tests pass locally.

There is one remaining validation edge case to fix before merge: repeated one-hot targets are detected only through collisions between generated category names. For an all-null STRING column there are no categories, so this currently succeeds silently and returns the unchanged frame:

```python
frame = ar.from_pandas(pd.DataFrame({"color": pd.Series([None, None], dtype="string")}))
encode_categorical(frame, ["color", "color"])  # currently accepted
```

Please reject duplicate target names before entering native encoding and add a regression test for this all-null case. The wrapper also currently does `list(columns)`, so a bare string such as `columns="color"` becomes `['c', 'o', ...]` and raises a misleading `KeyError`. Reusing the existing `_validate_existing_column_sequence(..., allow_empty=False, reject_duplicates=True)` helper (or equivalent validation) would cover both issues consistently; please add the bare-string TypeError test as well.

im-anishraj (approved):
@Naveen-Boddepalli, the follow-up resolves the remaining duplicate-target and bare-string validation gaps with focused regressions. I rebuilt the exact head as a Windows native wheel and all 23 categorical tests pass locally; Black, Ruff, diff checks, wheel smokes, and the full non-Vercel matrix are green after rerunning the transient setup-python failure. This is ready to merge.

PR #2489, opened 2026-06-06
test: exercise cloud-scheme rejection through public API (#2360)

### LINKED ISSUE #2360: Bug: published wheel bypasses documented cloud-scheme validation errors
## Summary

`docs/remote.md` says planned cloud-provider schemes should fail early with an actionable `ValueError`, for example:

```text
ValueError: Cloud scheme 's3' is not yet supported by arnio.
Install the optional extra when available: pip install "arnio[s3]"
```

However, the published `arnio==1.19.0` wheel passes these URLs into the local-file path and raises a generic `CsvReadError` instead.

## Reproduction

Run outside the repository so Python imports the published wheel:

```python
import arnio as ar

for url in [
    "s3://bucket/data.csv",
    "gs://bucket/data.csv",
    "az://container/data.csv",
    "abfs://container/data.csv",
]:
    try:
        ar.read_csv(url)
    except Exception as exc:
        print(url, type(exc).__name__, str(exc))
```

Observed locally on Windows with Python 3.13 and `arnio==1.19.0`:

```text
s3://bucket/data.csv CsvReadError Could not access CSV file 's3://bucket/data.csv': Invalid argument
gs://bucket/data.csv CsvReadError Could not access CSV file 'gs://bucket/data.csv': Invalid argument
az://container/data.csv CsvReadError Could not access CSV file 'az://container/data.csv': Invalid argument
abfs://container/data.csv CsvReadError Could not access CSV file 'abfs://container/data.csv': Invalid argument
```

## Expected behavior

Unsupported cloud schemes should be detected before the native local-file reader receives the URL, and the error should match the documented actionable message.

## Actual behavior

The published wheel returns a generic local path access error, which does not tell users whether cloud support exists, which extra to install later, or how to proceed.

## Acceptance criteria

- `read_csv`, `scan_csv`, and `read_csv_chunked` reject planned cloud schemes with the documented actionable error.
- Cover at least `s3://`, `gs://`, `az://`, `abfs://`, and `abfss://`.
- Add installed-wheel or smoke coverage so the behavior is present in release artifacts, not only source tests.
- Keep `docs/remote.md` and runtime behavior aligned.

dynamo-pentester:
## Summary

Fixes the regression described in #2360: `TestCloudSchemeRejectionPublicApi` was calling the internal `_materialize_csv_input` helper directly instead of the three public functions, so a bug in the Python-before-C++ guard inside `read_csv`, `scan_csv`, or `read_csv_chunked` would pass CI while users hit a generic `CsvReadError` from the published wheel.

No changes to `arnio/io.py` — the cloud-scheme `ValueError` guard is already correct in `main`.

## Changes

**`tests/test_remote_csv.py`**
- Rewrote `TestCloudSchemeRejectionPublicApi` to call `ar.read_csv`, `ar.scan_csv`, and `ar.read_csv_chunked` directly through the public API
- Added `abfss` to the scheme list (`_ALL_CLOUD_SCHEMES`) — it was in `_CLOUD_SCHEME_HINTS` but missing from this class
- Added a `read_csv_chunked` parametrized pair — calls `next(iter(...))` to trigger the guard inside the generator
- Added a second assertion per function checking the scheme name itself appears in the error message (matches the documented format)
- Result: 30 parametrized cases covering 3 functions × 5 schemes × 2 assertions

**`tests/smoke_wheel_install.py`**
- Added a `cloud_scheme_script` block that runs after the existing import check
- Exercises all five schemes against `read_csv`, `scan_csv`, and `read_csv_chunked` on the installed wheel
- Checks for correct exception type (`ValueError`) and presence of the `pip install` hint
- Written to a temp `.py` file to avoid shell-quoting issues; makes no network calls

## Testing
pytest tests/ -v   → 3369 passed, 113 skipped
black              → unchanged
ruff check         → all checks passed

## Closes

Closes #2360

### DIFF
--- a/tests/smoke_wheel_install.py
+++ b/tests/smoke_wheel_install.py
@@ -119,6 +119,51 @@ def main() -> int:
 
         run([str(python), "-c", import_check], cwd=tmp_dir)
 
+        # Verify that cloud-scheme rejection is present in the installed wheel.
+        # The ValueError must be raised by Python before the C++ extension is
+        # reached, so no real network access or cloud credentials are needed.
+        # Write to a temp script file to avoid shell quoting complexity.
+        cloud_scheme_script = (
+            "import arnio as ar\n"
+            "schemes = ['s3', 'gs', 'az', 'abfs', 'abfss']\n"
+            "errors = []\n"
+            "for scheme in schemes:\n"
+            "    url = f'{scheme}://bucket/file.csv'\n"
+            "    for fn in [ar.read_csv, ar.scan_csv]:\n"
+            "        try:\n"
+            "            fn(url)\n"
+            "            errors.append(f'{fn.__name__}({url}): expected ValueError, got no exception')\n"
+            "        except ValueError as exc:\n"
+            "            if 'pip install' not in str(exc):\n"
+            "                errors.append(f'{fn.__name__}({url}): missing pip hint — {exc}')\n"
+            "        except Exception as exc:\n"
+            "            errors.append(f'{fn.__name__}({url}): wrong exception type {type(exc).__name__} — {exc}')\n"
+            "    # read_csv_chunked must be iterated to trigger the guard\n"
+            "    try:\n"
+            "        next(iter(ar.read_csv_chunked(url)))\n"
+            "        errors.append(f'read_csv_chunked({url}): expected ValueError, got no exception')\n"
+            "    except ValueError as exc:\n"
+            "        if 'pip install' not in str(exc):\n"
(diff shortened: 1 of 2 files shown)


dynamo-pentester:
Hi @im-anishraj , Ci looks green. Please let me know if any changes needed !

im-anishraj (approved):
Scoped correctly to #2360: the public API paths and installed-wheel smoke path now cover all five planned cloud schemes without network access. The full non-Vercel matrix is green.

PR #2491, opened 2026-06-06
fix: make CSV transcoding writes atomic

### LINKED ISSUE #2487: write_csv: preserve destination and RuntimeError contract on transcoding failures
## Summary

The non-UTF-8 `write_csv()` transcoding path merged in #2485 has two failure-path regressions:

1. A strict encoding failure truncates an existing destination before the function raises.
2. Destination filesystem failures leak raw `OSError` subclasses instead of the documented `RuntimeError` used by the UTF-8/native path.

## Reproduction

```python
from pathlib import Path
import pandas as pd
import arnio as ar

frame = ar.from_pandas(pd.DataFrame({"text": ["こんにちは"]}))
out = Path("existing.csv")
out.write_bytes(b"sentinel")

try:
    ar.write_csv(frame, out, encoding="latin-1", encoding_errors="strict")
except ValueError:
    pass

assert out.read_bytes() == b"sentinel"  # currently b""
```

For a missing parent directory, `encoding="utf-8"` raises `RuntimeError`, while `encoding="latin-1"` raises `FileNotFoundError`.

## Expected fix

- Transcode into a second temporary output located in the destination directory.
- Close it successfully before atomically replacing the destination.
- Preserve an existing destination when encoding/transcoding fails.
- Clean both temporary files in `finally`.
- Wrap destination open/write/replace `OSError`s as `RuntimeError`, preserving the original exception as the cause.

## Acceptance criteria

- Focused regression test proves an existing file is unchanged after a strict unencodable-character failure.
- Focused regression test proves non-UTF-8 destination filesystem failures use the same public `RuntimeError` contract as UTF-8 writes.
- Existing `write_csv` encoding, line-terminator, formula-escaping, and temp-cleanup tests remain green.

Discovered during post-merge verification of #2485 on 2026-06-06.

eshaanag:
Fixes #2487

This PR keeps the change scoped to the non-UTF-8 transcoding path in `write_csv()`.

The previous flow wrote the UTF-8 intermediate output directly to the destination before transcoding it in place. If transcoding failed, an existing destination could be replaced or left partially written, and some destination filesystem errors escaped the public `RuntimeError` contract.

What changed:
- write transcoded output to a temporary file beside the destination
- replace the destination atomically only after transcoding closes successfully
- preserve an existing destination when encoding fails
- wrap destination filesystem failures as `RuntimeError`
- add regression coverage for failed transcoding, existing destinations, cleanup, and filesystem errors

Validation:
- `pytest tests/test_write_csv.py -q` - 81 passed
- `pytest tests/ -q` - 3315 passed, 106 skipped
- `ruff check . --extend-exclude "*.ipynb"`
- `black --check --diff .`
- `mypy --config-file mypy.ini`
- `python scripts/check_docs_utf8.py`

This is for GSSoC 2026 and stays scoped to the assigned CSV write atomicity issue.

### DIFF
--- a/arnio/io.py
+++ b/arnio/io.py
@@ -1224,6 +1224,7 @@ def write_csv(
     _CHUNK_SIZE = 1 << 20  # 1 MiB per chunk
 
     tmp_fd, tmp_path = tempfile.mkstemp(suffix=".csv")
+    output_tmp_path: str | None = None
     try:
         os.close(tmp_fd)
         try:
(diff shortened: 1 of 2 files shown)


im-anishraj (approved):
This implements the intended #2487 failure-path fix: transcoding completes in a destination-side temporary file, replacement happens only after a successful close, destination OSErrors retain the RuntimeError contract, and the focused regressions plus full non-Vercel matrix are green.

PR #2492, opened 2026-06-06
fix: preserve empty pandas int64 dtype in from_pandas

### LINKED ISSUE #1960: Bug: preserve empty pandas int64 dtype in from_pandas
## Summary

`from_pandas()` loses regular pandas integer dtype information for zero-row DataFrames. An empty `int64` column is converted to an Arnio `string` column, and `to_pandas()` then returns it as `string` instead of preserving numeric intent.

## Why it matters

Empty DataFrames are common in filtered pipelines, smoke tests, and schema-first workflows. If Arnio changes an empty integer column into a string column, downstream schema validation, dtype-based selection, and pandas interoperability can behave differently depending only on whether the current batch has rows.

## Evidence

- `arnio/convert.py`: `from_pandas()` passes `dtype_hints` to `_Frame.from_dict(..., len(df))`.
- `arnio/convert.py`: `_pandas_dtype_to_arnio()` maps nullable `pd.Int64Dtype()` but does not map regular NumPy/pandas `int64` for empty columns.
- With no values present, the native frame cannot infer `int64` from data, so the column falls back to `string`.

## Reproduction or trigger

```python
import pandas as pd
import arnio as ar

source = pd.DataFrame({"id": pd.Series([], dtype="int64")})
frame = ar.from_pandas(source)

print(frame.shape)
# (0, 1)

print(frame.dtypes)
# Actual: {'id': 'string'}

print(ar.to_pandas(frame).dtypes.astype(str).to_dict())
# Actual: {'id': 'string'}
```

## Expected behavior

`from_pandas()` should preserve supported empty-column dtype hints consistently. For a zero-row `int64` pandas column, Arnio should create an `int64` column and round-trip it back as an integer dtype where supported.

## Actual behavior

The empty `int64` column becomes an Arnio `string` column.

## Suggested direction

Extend `_pandas_dtype_to_arnio()` to recognize regular pandas/NumPy integer dtypes such as `int64` when they are supported by Arnio, and add regression tests for zero-row `int64`, nullable `Int64`, `float64`, `bool`, and `string` columns.

## References

- Relevant file: `arnio/convert.py`
- Related broad area: pandas interoperability and zero-row frame correctness

aaniya22:
## Summary

Fixes the bug where `from_pandas()` loses regular numpy `int64` dtype information for zero-row DataFrames. An empty `int64` column was being converted to an Arnio `string` column, causing `to_pandas()` to return it as `string` instead of `Int64`.

## Root cause

`_pandas_dtype_to_arnio()` in `arnio/convert.py` handled nullable `pd.Int64Dtype()` but did not handle regular numpy `int64`. With no values present, the native frame could not infer the dtype from data, so the column fell back to `string`.

## Fix

Added `str(dtype) == "int64"` check to `_pandas_dtype_to_arnio()` so the dtype hint is correctly passed to `_Frame.from_dict()` for empty columns.

## Tests

Added `TestEmptyInt64DtypePreservation` in `tests/test_convert.py` covering:
- empty numpy `int64` roundtrip
- empty nullable `Int64` roundtrip
- empty `float64` roundtrip
- empty `bool` roundtrip
- empty `string` roundtrip
- empty mixed schema roundtrip

Closes #1960

### DIFF
--- a/arnio/convert.py
+++ b/arnio/convert.py
@@ -349,6 +349,8 @@ def to_arrow(frame: ArFrame) -> pa.Table:
 def _pandas_dtype_to_arnio(dtype: object) -> _DType | None:
     if dtype == pd.Int64Dtype():
         return _DType.INT64
+    if str(dtype) == "int64":
+        return _DType.INT64
     if dtype == pd.Float64Dtype():
         return _DType.FLOAT64
     if str(dtype) == "float64":
@@ -356,7 +358,6 @@ def _pandas_dtype_to_arnio(dtype: object) -> _DType | None:
     if dtype == pd.BooleanDtype() or str(dtype) == "bool":
         return _DType.BOOL
     return None
-
 
 def _validate_unique_column_labels(labels: pd.Index) -> None:
     seen: set[object] = set()
--- a/arnio/convert.py
+++ b/arnio/convert.py
@@ -359,6 +359,7 @@ def _pandas_dtype_to_arnio(dtype: object) -> _DType | None:
         return _DType.BOOL
     return None
 
+
 def _validate_unique_column_labels(labels: pd.Index) -> None:
     seen: set[object] = set()
     dupes: list[object] = []
--- a/tests/test_convert.py
+++ b/tests/test_convert.py
@@ -1463,6 +1463,7 @@ def test_from_arrow_table_missing_pyarrow_raises_helpful_import_error():
     assert "_from_arrow_table() requires pyarrow." in str(exc_info.value)
     assert "pip install arnio[arrow]" in str(exc_info.value)
 
+
 # ---------------------------------------------------------------------------
 # Issue #1960 — preserve empty pandas int64 dtype in from_pandas
 # ---------------------------------------------------------------------------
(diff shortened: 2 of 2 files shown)


im-anishraj (changes requested):
The implementation and test scope match #1960, but both changed Python files currently fail Black (`arnio/convert.py` and `tests/test_convert.py` would be reformatted), and `tests/test_convert.py` is missing its final newline. Please run Black on both files and push the formatting-only correction. No behavioral expansion is requested.

im-anishraj (changes requested):
@aaniya22, thanks for keeping this focused on the assigned empty-frame dtype bug. I added the matching GSSoC classification labels and reviewed the implementation/tests. Before this can proceed, please run the repository formatter and push the result: `black --check arnio/convert.py tests/test_convert.py` currently reports that both changed files would be reformatted (including the top-level spacing in `convert.py` and the mixed-schema DataFrame layout in the test file). Please also ensure the test file ends with a newline. The logic and regression coverage otherwise match #1960's requested scope; CI should rerun on the formatted head.

aaniya22:
> @aaniya22, thanks for keeping this focused on the assigned empty-frame dtype bug. I added the matching GSSoC classification labels and reviewed the implementation/tests. Before this can proceed, please run the repository formatter and push the result: `black --check arnio/convert.py tests/test_convert.py` currently reports that both changed files would be reformatted (including the top-level spacing in `convert.py` and the mixed-schema DataFrame layout in the test file). Please also ensure the test file ends with a newline. The logic and regression coverage otherwise match #1960's requested scope; CI should rerun on the formatted head.

ok

aaniya22:
Applied Black formatting to both files as requested. All done!

im-anishraj (approved):
@aaniya22 the formatting fix is clean and the focused regression covers the empty `int64` dtype path without changing non-empty behavior. Approved.

PR #2497, opened 2026-06-06
fix: validate DateTime format directives at construction time

### LINKED ISSUE #1825: Bug: validate DateTime format directives at construction time
## Summary

`DateTime(format=...)` validates that `format` is a string, but it does not validate whether the supplied datetime format is actually usable before returning a `Field`.

## Why it matters

Schema constructors generally try to fail early for invalid configuration. For example, `String(pattern=...)` validates regex syntax before the schema is used. `DateTime(format=...)` should behave similarly so users find configuration mistakes when building the schema, not later during validation.

## Evidence

- `arnio/schema.py`: `DateTime()` checks `format is not None and not isinstance(format, str)`, but does not validate invalid directives.
- `_validate_datetime()` later passes `field_def.format` into `pd.to_datetime(..., format=...)`, where invalid directives raise a pandas `ValueError` during validation.
- `tests/test_schema.py` covers non-string `format`, invalid min/max values, and valid DateTime validation, but does not cover syntactically invalid format directives.

## Reproduction or trigger

```python
import pandas as pd
import arnio as ar

field = ar.DateTime(format="bad-%Q")
frame = ar.from_pandas(pd.DataFrame({"ts": ["2026-01-01"]}))

ar.validate(frame, ar.Schema({"ts": field}))
```

## Expected behavior

`ar.DateTime(format="bad-%Q")` should fail fast with a clear `ValueError` explaining that the datetime format is invalid.

## Actual behavior

The constructor succeeds, and validation later raises a pandas-origin error:

```text
ValueError: 'Q' is a bad directive in format 'bad-%Q'
```

## Suggested direction

Validate the `format` string in `DateTime()` before constructing the `Field`. Add tests for an invalid directive and a valid custom format such as `%Y-%m-%d`.

## References

- `arnio/schema.py`
- `tests/test_schema.py`

zxan:
## Summary
Adds fail-fast validation of the `format` argument in `DateTime()`. Previously, invalid directives like `bad-%Q` were accepted at construction time and only raised a confusing pandas-origin `ValueError` later during validation. Now `DateTime(format=...)` validates the format string immediately using `datetime.strptime` and raises a clear `ValueError` at construction time.

## Linked issue
Fixes #1825 

## Type of change
- [x] Bug fix
- [x] Tests

## Area
- [x] Schema validation
- [x] Python API / ArFrame

## Testing

```bash
python3 -m pytest arnio/tests/test_schema.py -k "datetime_rejects_invalid_format or datetime_accepts_valid_custom" -v
```

All 3 tests passed:
- `test_datetime_rejects_invalid_format_type` (existing)
- `test_datetime_rejects_invalid_format_directive` (new)
- `test_datetime_accepts_valid_custom_format` (new)

- [x] `python -m pytest tests -v --tb=short -x`

## Contributor checklist
- [x] My PR title uses Conventional Commits (`fix: validate DateTime format directives at construction time`).

### DIFF
--- a/arnio/schema.py
+++ b/arnio/schema.py
@@ -5,6 +5,7 @@ Production data contracts and validation.
 
 from __future__ import annotations
 
+from datetime import datetime
 import json
 import math
 import re
@@ -2486,7 +2487,14 @@ def DateTime(
     """Create a datetime schema field for validating string timestamps."""
     if format is not None and not isinstance(format, str):
         raise TypeError("DateTime format must be a string or None")
-
+    if format is not None:
+        try:
+            datetime.strptime("2000-01-01", format)
+        except ValueError as exc:
+            raise ValueError(
+                f"DateTime format {format!r} contains an invalid or unsupported directive: {exc}"
+            ) from exc
+        
     min_val = _parse_datetime_bound(min, "min")
     max_val = _parse_datetime_bound(max, "max")
     if min_val is not None and max_val is not None and min_val > max_val:
--- a/tests/test_schema.py
+++ b/tests/test_schema.py
@@ -2469,6 +2469,16 @@ def test_datetime_rejects_invalid_format_type():
     with pytest.raises(TypeError, match="format must be a string or None"):
         ar.DateTime(format=123)
 
+        
+def test_datetime_rejects_invalid_format_directive():
+    with pytest.raises(ValueError, match="invalid or unsupported directive"):
+        ar.DateTime(format="bad-%Q")
+
+
+def test_datetime_accepts_valid_custom_format():
+    field = ar.DateTime(format="%Y-%m-%d")
+    assert field.format == "%Y-%m-%d"        
+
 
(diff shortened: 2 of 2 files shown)


im-anishraj (changes requested):
@azaanhusain777, thanks for getting the PR in before your deadline. I reviewed head `8f677bb`, and the current validation method has one correctness blocker: `datetime.strptime("2000-01-01", format)` rejects valid formats that omit components or use a different shape, such as `%Y`, `%H:%M`, or `%Y/%m`. That conflicts with the issue guidance to validate directives without requiring one particular sample layout.

Please build a sample from the format itself (for example, format a known `datetime` with `strftime` and parse that generated value back), then add tests for valid partial/time-only formats as well as invalid `%Q`. Also run Black/Ruff or remove the visible trailing whitespace in both touched files. The current head has no non-Vercel CI signal yet; let the full checks run after the fix. Vercel authorization is external and not your blocker.

PR #2515, opened 2026-06-08
fix(csv): format double fields with to_chars

### LINKED ISSUE #1034: GSSoC: Standardize CSV Writer Floating-Point Precision
### Track

CSV parser and I/O

### GSSoC level

Level 1 - beginner

### Context

Currently, when our library's native C++ CSV Writer writes floating-point (double) data to a file, it relies on generalized formatting streams (`std::ostringstream`). This creates two problems for our I/O engine:

1. High performance overhead: Creating streams inside heavy loops causes unexpected memory heap allocations.
2. Inconsistent precision: It doesn't always guarantee a portable, perfect round-trip representation matching Python/Pandas defaults.

By switching this path to use C++17's low-level `std::to_chars` inside a fixed stack buffer, we can make floating-point CSV writing completely allocation-free and strictly accurate.

### Scope

- **Locate the Writer file:** Navigate to `arnio/csv_writer.cpp` (or your project's CSV writer implementation).
- **Refactor `double` handling:** Identify the block processing `std::holds_alternative<double>(cell)` inside the core row loop.
- **Implement `std::to_chars`:** Remove `std::ostringstream` and replace it with a fixed stack buffer (`char num_buffer[64]`) combined with `std::to_chars` using `std::chars_format::general`.
- **Direct Stream Write:** Write the populated buffer chunk directly to the stream using `out.write(num_buffer, ptr - num_buffer)`.
- **Add a Test Case:** Add a basic unit test verifying that floating-point numbers retain 17 digits of significant precision upon export.

### Acceptance criteria

- [ ] Floating-point fields use `std::to_chars` instead of `std::ostringstream`.
- [ ] No extra string heap allocations are created inside the row writing loop.
- [ ] Tests pass locally and in CI.
- [ ] Public behavior is documented.

### Implementation Tips:
When using `std::to_chars` for a `double`, make sure to include the `<charconv>` header. 

Example approach:
```cpp
char num_buffer[64];
auto [ptr, ec] = std::to_chars(num_buffer, num_buffer + sizeof(num_buffer), double_val, std::chars_format::general);
if (ec == std::errc()) {
    out.write(num_buffer, ptr - num_buffer);
}

Assigm me under gssoc 2026

### PR DESCRIPTION
**Pull request by eshaanag**
## Summary
Fixes #1034

This PR keeps the change scoped to the maintainer-approved `cell_to_string()` double-formatting path. The writer still returns a string for each cell, so existing CSV quoting, delimiter handling, and line terminator behavior stay on the same path.

What changed:
- replaced the `std::ostringstream` double formatting branch with `std::to_chars`;
- kept `std::chars_format::general` with `max_digits10` so high-precision values keep the existing round-trip-oriented precision contract;
- added a stack-buffer `snprintf("%.17g")` fallback only for older macOS deployment targets where AppleClang marks floating-point `std::to_chars` unavailable;
- handle `std::errc` explicitly and fail with a clear runtime error if formatting fails;
- added raw output regression coverage for a high-precision float value.

## Linked issue
Fixes #1034

## Type of change
- [x] Bug fix
- [ ] Feature
- [ ] Performance improvement
- [ ] Documentation
- [x] Tests
- [ ] Refactor
- [ ] CI / packaging

## Area
- [x] CSV parser / I/O
- [ ] C++ cleaning engine
- [ ] Python API / ArFrame
- [ ] Pipeline / custom steps
- [ ] Schema validation
- [ ] pandas interoperability
- [ ] Docs / website
- [ ] Developer tooling

## Testing
- [ ] `make lint` (or run separately):
```bash
  python scripts/check_docs_utf8.py
  python -m black --check --diff .
  python -m ruff check . --extend-exclude "*.ipynb"
  python -m mypy --config-file mypy.ini
  find cpp/ bindings/ -type f \( -name '*.cpp' -o -name '*.h' \) | xargs clang-format --dry-run --Werror
  ```
- [x] optionally `python -m pytest tests -v --tb=short -x`
- [x] Other:

Validation I ran:
- `.venv/bin/python -m pytest tests/test_write_csv.py` - 82 passed
- `.venv/bin/python -m pytest` - 3494 passed, 106 skipped
- `.venv/bin/python -m ruff check tests/test_write_csv.py` - passed
- `.venv/bin/clang-format --dry-run --Werror cpp/src/csv_writer.cpp` - passed
- `/usr/bin/clang++ -std=gnu++17 -arch arm64 -mmacosx-version-min=11.0 -Icpp/include -c cpp/src/csv_writer.cpp -o /private/tmp/arnio_csv_writer_macos11.o` - passed
- `git diff --check` - passed

Note: the first full pytest run inside the restricted shell failed only because local loopback socket binding was blocked for `tests/test_remote_csv.py`. Re-running the same command with loopback permission passed the full suite.

## Screenshots / Media
Not applicable.

## Performance Impact
This removes the `std::ostringstream` construction from double cell formatting and uses a fixed stack buffer for the conversion. On macOS deployment targets below 13.3, AppleClang exposes floating-point `std::to_chars` as unavailable, so that path uses stack-buffer `snprintf` to keep the wheel build compatible without going back to stream formatting. I did not run benchmarks because the PR is intentionally limited to the approved writer formatting path.

## GSSoC labels
This is for GSSoC 2026 and stays scoped to the assigned issue. Please add the appropriate GSSoC approval/scoring labels if this matches the maintainer-approved scope.

## Contributor checklist

## Maintainer notes
I did not rewrite the row writer into a direct `out.write()` pipeline. The change stays inside the existing `cell_to_string()` contract as requested.

No docs/examples were changed because this does not add a public API option or new user-facing parameter.

### DIFF
--- a/cpp/src/csv_writer.cpp
+++ b/cpp/src/csv_writer.cpp
@@ -1,13 +1,14 @@
 #include "arnio/csv_writer.h"
 
+#include <charconv>
+#include <cstdio>
 #ifdef _WIN32
 #include <filesystem>
 #endif
 #include <fstream>
-#include <iomanip>
 #include <limits>
-#include <sstream>
 #include <stdexcept>
+#include <system_error>
 #include <variant>
 
 #include "arnio/frame.h"
@@ -20,6 +21,28 @@ inline void open_binary_output(std::ofstream& file, const std::string& path) {
     file.open(std::filesystem::u8path(path), std::ios::binary);
 #else
     file.open(path, std::ios::binary);
+#endif
+}
+
+inline std::string format_double(double value) {
+    char num_buffer[64];
+
+#if defined(__APPLE__) && defined(__ENVIRONMENT_MAC_OS_X_VERSION_MIN_REQUIRED__) && \
+    __ENVIRONMENT_MAC_OS_X_VERSION_MIN_REQUIRED__ < 130300
+    const int written = std::snprintf(num_buffer, sizeof(num_buffer), "%.*g",
+                                      std::numeric_limits<double>::max_digits10, value);
+    if (written < 0 || static_cast<size_t>(written) >= sizeof(num_buffer)) {
+        throw std::runtime_error("Failed to format floating-point CSV value");
+    }
+    return std::string(num_buffer, static_cast<size_t>(written));
+#else
+    const auto [ptr, ec] =
+        std::to_chars(num_buffer, num_buffer + sizeof(num_buffer), value,
+                      std::chars_format::general, std::numeric_limits<double>::max_digits10);
+    if (ec != std::errc()) {
+        throw std::runtime_error("Failed to format floating-point CSV value");
+    }
+    return std::string(num_buffer, ptr);
 #endif
 }
 }  // namespace
@@ -75,10 +98,7 @@ std::string CsvWriter::cell_to_string(const Frame& frame, size_t row, size_t col
         return std::to_string(std::get<int64_t>(cell));
     }
     if (std::holds_alternative<double>(cell)) {
-        std::ostringstream oss;
-        oss << std::setprecision(std::numeric_limits<double>::max_digits10)
-            << std::get<double>(cell);
-        return oss.str();
+        return format_double(std::get<double>(cell));
     }
     if (std::holds_alternative<bool>(cell)) {
         return std::get<bool>(cell) ? "true" : "false";
--- a/tests/test_write_csv.py
+++ b/tests/test_write_csv.py
@@ -99,6 +99,14 @@ class TestWriteCsv:
         df = ar.to_pandas(frame2)
         assert abs(df["val"].iloc[0] - 1.23456789012345678) < 1e-15
 
+    def test_high_precision_float_writes_max_digits10(self, tmp_path):
+        frame = ar.from_pandas(pd.DataFrame({"val": [1.2345678901234567]}))
+        out = tmp_path / "float.csv"
+
+        ar.write_csv(frame, out)
+
+        assert out.read_text(encoding="utf-8") == "val\n1.2345678901234567\n"
+
     def test_invalid_delimiter(self, tmp_path):
         frame = ar.from_pandas(pd.DataFrame({"a": [1]}))
         with pytest.raises(ValueError, match="delimiter must be a single character"):

### CODE REVIEW

**Review by im-anishraj — DISMISSED** [review]
@eshaanag thanks for the focused fix and for keeping the change scoped to the CSV writer path.

This is blocked on the macOS arm64 wheel target right now. The failing non-Vercel check is `macOS arm64 Wheel Smoke Test` / `Build arm64 wheel and smoke test (macOS)`, and the build fails in `cpp/src/csv_writer.cpp` with:

```text
error: 'to_chars' is unavailable: introduced in macOS 13.3
```

The wheel job builds with `-mmacosx-version-min=11.0`, so `std::to_chars` for floating-point values cannot be used unconditionally for this code path unless we raise/change the supported macOS target. Please update the implementation to keep compatibility with the current macOS wheel target, for example by using a portable fallback on older Apple targets or another formatting path that preserves the intended precision contract.

I added the GSSoC/classification labels for this PR and left it as `status:blocked` until the macOS wheel build and the normal CI matrix are green on a new head.

im-anishraj (dismissed):
@eshaanag thanks for the focused PR and tests. I reviewed the diff and the current CI, and this is blocked by the non-Vercel macOS arm64 wheel check.

The failure is deterministic on the repository's current wheel target: `cpp/src/csv_writer.cpp` now calls floating-point `std::to_chars`, but the arm64 wheel build targets `-mmacosx-version-min=11.0`; Apple's libc++ marks that overload unavailable until macOS 13.3:

`error: 'to_chars' is unavailable: introduced in macOS 13.3`

Please revise the double-formatting path so it remains compatible with the supported macOS 11 arm64 wheel target, while preserving the max-digits/round-trip behavior and the new regression coverage. Once that is pushed and the non-Vercel checks are green, I can re-review it.

eshaanag:
Pushed the compatibility fix for the macOS 11 wheel target.

What changed:
- kept the `std::to_chars` path for supported platforms
- added a stack-buffer fallback for older Apple deployment targets where floating-point `to_chars` is unavailable
- kept the max-digits precision behavior covered by the new CSV writer regression test

I re-checked the PR after the push: the macOS arm64 wheel smoke test is passing now, along with the full Python matrix, C++ native tests, lint, CodeQL, examples, and compatibility checks. The only remaining failure is the Vercel authorization check, which is external to this code path.

im-anishraj (approved):
@eshaanag thanks for the compatibility follow-up. I rechecked the current diff and the full non-Vercel matrix: the macOS arm64 wheel smoke test, native tests, lint, Python matrix, benchmark, CodeQL, and related checks are all green now; only the external Vercel authorization status remains failing. The implementation stays scoped to the existing cell_to_string() double-formatting path and keeps the macOS 11 wheel target compatible. Approving this for maintainer merge.

im-anishraj:

PR #2539, opened 2026-06-10
Feature: add append mode to write_csv (Fixes #644)

### LINKED ISSUE #644: Feature: add append mode to write_csv
## Problem
Arnio currently lacks this product capability: Appending cleaned rows to an existing CSV is currently not a first-class operation.

## Evidence / audit context
- Issue title: Feature: add append mode to write_csv
- Audit focus: Appending cleaned rows to an existing CSV is currently not a first-class operation.
- Labels selected: type: feature, area: csv-parser, priority: medium, difficulty: intermediate, status: needs maintainer, size: m

## Suggested fix

## Acceptance criteria

basantnema31:
## Summary

Fixes #644

### Problem
Currently, `write_csv` always overwrites existing files, making it impossible to append data incrementally to a CSV. Users have to read the existing file, concatenate, and write the whole file, which is memory-inefficient and slow.

### Solution
- Added an `append: bool = False` flag to `write_csv` in both Python and C++ writer config.
- When `append=True` and the file exists and is non-empty, `write_csv` validates the schema against the existing file using `scan_csv` to avoid silent data corruption.
- If the file is non-empty, headers are automatically omitted to prevent duplicated headers.
- The C++ `CsvWriter` was updated to open the file stream with `std::ios::app` when appending.

### DIFF
--- a/arnio/_arnio_cpp.pyi
+++ b/arnio/_arnio_cpp.pyi
@@ -136,6 +136,7 @@ class CsvWriteConfig:
     write_header: bool
     line_terminator: str
     escape_formulas: bool
+    append: bool
 
     def __init__(self) -> None: ...
 
--- a/arnio/io.py
+++ b/arnio/io.py
@@ -1355,6 +1355,8 @@ def write_csv(
         If ``encoding`` is an unknown codec, ``encoding_errors`` is not one of
         ``"strict"``, ``"replace"``, or ``"ignore"``, or if a character cannot
         be encoded in the requested encoding with ``encoding_errors="strict"``.
+        Also raised if appending to a file that lacks a trailing newline, or
+        if appending with a schema that doesn't match the existing CSV.
     RuntimeError
         If the file cannot be opened or written.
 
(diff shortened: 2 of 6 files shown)


im-anishraj (changes requested):
@basantnema31, thanks for carrying append support through the Python and native layers. I found correctness and scope blockers before this can merge:\n\n1. With a non-UTF-8 ncoding, append writes only the new rows to a fresh UTF-8 temp file and then os.replace() replaces the destination, so the existing CSV is lost. Append must preserve the existing encoded bytes/content and remain atomic on failure.\n2. Appending to a populated CSV that lacks a final line terminator will concatenate the first new row onto the last existing row; handle or reject that boundary safely.\n3. The assigned scope requires focused overwrite, append-to-empty, append-to-existing, header, schema-mismatch, and docs/example coverage. The PR currently has one combined test and no public API docs/example update. Please add separate tests, including non-UTF-8 append, missing-final-newline, ppend=False, invalid ppend, and header behavior.\n\nPlease keep the fix narrow and preserve current overwrite/transcoding behavior.

im-anishraj:

PR #2544, opened 2026-06-10
fix: normalize data quality suggestion kwargs for JSON

### LINKED ISSUE #186: Quality: Add to_json export for QualityReport
## Context
Reports should be serializable for CI, dashboards, and saved audit artifacts.

## Suggested files
arnio/quality.py, arnio/schema.py, tests/test_quality.py, README.md

## Scope

## Acceptance criteria

## Contributor notes

shreyansh-tech21:
Fixes #186

## Summary

This PR keeps `DataQualityReport.to_dict()` as the canonical JSON-oriented payload and normalizes cleaning suggestion kwargs there, after existing `exclude_columns`, `cast_types`, and `subset` filtering.

`DataQualityReport.to_json()` remains a thin wrapper over `to_dict()` plus `json.dumps()`. This avoids adding a parallel export API or changing the report shape.

## Scope

* Normalize suggestion kwargs to JSON-safe values inside `DataQualityReport.to_dict()`
* Preserve existing redaction and exclusion semantics
* Keep suggestion sorting deterministic
* Keep `to_json()` API behavior unchanged
* Do not add `ValidationResult.to_json()` in this PR
* Avoid README churn unless docstrings are insufficient

## Why

The current export path is already the right shape. The gap is that suggestion kwargs can contain values that are not directly JSON-serializable, while other render paths tolerate them through `default=str`.

Normalizing in `to_dict()` fixes the contract at the source, so callers can safely use:

```python
json.dumps(report.to_dict())
```

for CI artifacts and integrations.

## Tests

Added focused coverage for:

* non-JSON-native suggestion kwargs becoming JSON-safe through `to_dict()`
* `to_json()` round-tripping consistently with `to_dict()`
* redaction and `exclude_columns` continuing to prevent leakage after kwargs normalization

### DIFF
--- a/arnio/quality.py
+++ b/arnio/quality.py
@@ -3184,7 +3184,7 @@ def _json_safe_suggestion_value(value: Any) -> Any:
     if isinstance(value, dict):
         return {str(k): _json_safe_suggestion_value(v) for k, v in value.items()}
 
-    if isinstance(value, set):
+    if isinstance(value, Set) and not isinstance(value, (str, bytes, bytearray)):
         try:
             ordered = sorted(value)
         except TypeError:
@@ -3226,7 +3226,7 @@ def _filtered_suggestion_kwargs(
         raw_value = value
 
         if key in {"subset", "columns"} and isinstance(
-            value, (list, tuple, set, np.ndarray)
+            value, (list, tuple, Set, np.ndarray)
         ):
             raw_value = [item for item in value if item not in exclude_columns]
 
--- a/arnio/quality.py
+++ b/arnio/quality.py
@@ -5,7 +5,6 @@ Data quality profiling and safe automatic cleaning helpers.
 
 from __future__ import annotations
 
-import decimal
 import html
 import json
 import math
@@ -3175,7 +3174,13 @@ def _approx_top_values(
 
 
 def _json_safe_suggestion_value(value: Any) -> Any:
-    """Normalize suggestion kwargs values after redaction/filtering only."""
+    """Normalize cleaning-suggestion kwargs values for json.dumps without default=."""
+    if value is None or isinstance(value, (str, bool, int)):
+        return value
+
+    if isinstance(value, float):
+        return value if math.isfinite(value) else None
+
     if isinstance(value, dict):
         return {str(k): _json_safe_suggestion_value(v) for k, v in value.items()}
 
@@ -3195,10 +3200,17 @@ def _json_safe_suggestion_value(value: Any) -> Any:
     if isinstance(value, date):
         return value.isoformat()
 
-    if isinstance(value, decimal.Decimal):
-        return str(value)
+    if hasattr(value, "item"):
+        try:
+            return _json_safe_suggestion_value(value.item())
+        except Exception:
+            return str(value)
 
-    return _clean_scalar(value)
+    cleaned = _clean_scalar(value)
+    if cleaned is not value:
+        return _json_safe_suggestion_value(cleaned)
+
+    return str(value)
 
 
 def _filtered_suggestion_kwargs(
--- a/arnio/quality.py
+++ b/arnio/quality.py
@@ -5,6 +5,7 @@ Data quality profiling and safe automatic cleaning helpers.
 
 from __future__ import annotations
 
+import decimal
 import html
 import json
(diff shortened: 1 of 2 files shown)


im-anishraj (changes requested):
@shreyansh-tech21, this replacement is much cleaner: the unrelated deletions are gone, existing exclusion/redaction behavior is retained, and the focused tests cover several useful nested types. One JSON-safety blocker remains in `_json_safe_suggestion_value`: its final `return _clean_scalar(value)` can still return an arbitrary unsupported object unchanged, and NumPy `.item()` results are not normalized recursively. For example, `pathlib.Path`, an `Enum`, a complex value, or `np.datetime64(...).item()` can still make `json.dumps(report.to_dict())` fail.

Please make the leaf handling guarantee JSON primitives for every value: preserve supported primitives, recursively normalize scalar `.item()` results, retain the existing non-finite handling, and use a deterministic string fallback for otherwise unsupported objects (consistent with the existing renderer behavior). Add focused regressions for at least one arbitrary unsupported leaf and one NumPy scalar that unwraps to another non-JSON-native type. Once that is covered and non-Vercel CI is green, I can run the exact-head quality suite.

im-anishraj (changes requested):
@shreyansh-tech21, this replacement is now cleanly scoped and preserves the existing quality APIs/tests, but one acceptance blocker remains in the JSON-safe fallback.

`_json_safe_suggestion_value()` ends with `return _clean_scalar(value)`. `_clean_scalar()` only normalizes pandas/numpy scalars and non-finite floats; it returns an arbitrary Python object unchanged. The repository already has a concrete regression fixture in `test_report_to_markdown_suggestions_non_json_serializable()` where `DummyObject()` stringifies to `custom_val`. Passing that value through the new helper still leaves a `DummyObject` in `to_dict()`, so `json.dumps(report.to_dict())` and `report.to_json()` still raise `TypeError`. `frozenset` has a similar gap because only the concrete `set` type is handled.

Please make the terminal fallback guarantee a JSON-native scalar (for example, clean first, return native `None`/bool/number/string values, and deterministically stringify any remaining object), handle general set-like values consistently, and add focused tests using the existing custom-object pattern plus a `frozenset`/mixed nested value. Keep the current redaction-before-normalization ordering and the rest of this focused diff. Once that is fixed and non-Vercel CI is green, this should be ready for re-review.

im-anishraj (changes requested):
@shreyansh-tech21, the new commit fixes the arbitrary-object fallback and recursively normalizes NumPy `.item()` results; thank you, those blockers are resolved. Two focused items remain before re-review:

1. The helper still checks only the concrete `set` type, so `frozenset` falls through to a string instead of the deterministic JSON array used for sets. Please handle general set-like values (the module already imports `collections.abc.Set`) and add the requested `frozenset`/mixed nested regression.
2. Non-Vercel CI is now failing across Windows because the new `Path("data/input.csv")` assertion assumes POSIX separators; Windows correctly stringifies it as `data\input.csv`. Please make that test platform-independent, for example by comparing with `str(Path("data/input.csv"))`.

Keep the current redaction ordering and scoped implementation. Once these are fixed and non-Vercel CI is green, I can run the exact-head quality suite.

im-anishraj (commented):
@shreyansh-tech21, the latest commit addresses both requested items: general set-like values now normalize consistently, the `frozenset` regression is covered, and the `Path` assertion is platform-independent. I reviewed the full diff and ran the exact head (`03b796ef`) against `tests/test_quality.py`: **330 passed** on Windows/Python 3.13.

No further code changes are requested from this review. Two non-Vercel CI jobs are still running, so I am leaving the PR blocked until they complete successfully; the Vercel authorization failure remains known external noise.

im-anishraj (approved):
@shreyansh-tech21, approved. The requested JSON-safety fixes are complete, all non-Vercel CI checks are green, and the exact head `03b796ef` passed all 330 targeted quality tests locally. Thank you for the careful follow-up.

im-anishraj (approved):
All requested fixes are present on exact head 03b796ef, the focused quality suite passed (330 tests), and all non-Vercel CI checks are green. The remaining Vercel authorization failure is known external noise. Approved.

PR #2550, opened 2026-06-11
feat : added exclude_columns parameter to pandas profile accessor

### LINKED ISSUE #1374: Feature: expose exclude_columns on pandas profile accessor
## Summary

`ar.profile()` supports `exclude_columns`, but the pandas accessor wrapper `df.arnio.profile()` does not expose that option.

This creates an integration parity gap: pandas users can profile through Arnio, but they cannot use the privacy/noisy-column exclusion path without leaving the accessor API and manually converting to `ArFrame`.

## Why it matters

The pandas accessor is meant to make Arnio easy to use inside existing pandas workflows. `exclude_columns` is important for production profiling because datasets often contain sensitive, high-cardinality, or irrelevant columns such as IDs, tokens, emails, or free-text fields.

Users should be able to write:

```python
report = df.arnio.profile(exclude_columns=["email", "token"])
```

instead of switching APIs for the same profiling feature.

## Evidence

- `arnio/quality.py`: `profile(frame, *, exclude_columns=None, ...)` supports excluding columns and validates the requested names.
- `arnio/integrations/pandas.py`: `ArnioPandasAccessor.profile()` forwards profiling options but has no `exclude_columns` parameter.
- `tests/test_pandas_accessor.py`: accessor tests cover basic profiling but do not cover excluding columns through `df.arnio.profile()`.

## Reproduction or trigger

1. Start with a pandas DataFrame that contains a sensitive or noisy column.
2. Call `df.arnio.profile(exclude_columns=["secret"])`.
3. Python raises `TypeError` because the accessor method does not accept the parameter, even though the underlying `ar.profile()` API does.

## Expected behavior

`df.arnio.profile()` should accept `exclude_columns` and forward it to `ar.profile()`.

## Actual behavior

`df.arnio.profile()` exposes only sample/top-value parameters, so `exclude_columns` cannot be used through the pandas accessor.

## Suggested direction

- Add `exclude_columns: Sequence[str] | None = None` to `ArnioPandasAccessor.profile()`.
- Forward it to `profile(self.to_arframe(), exclude_columns=exclude_columns, ...)`.
- Add tests that verify excluded columns are omitted from the returned `DataQualityReport`.
- Add a small README/API example if the pandas accessor section is updated.

## References

- `arnio/integrations/pandas.py`
- `arnio/quality.py`
- `tests/test_pandas_accessor.py`

**Priority:** low

tmdeveloper007:
Closes #1374.

Summary of What Has Been Done:
Exposed the exclude_columns option on the pandas DataFrame accessor profile method.

Changes Made:
- Added exclude_columns parameter to ArnioPandasAccessor.profile in arnio/integrations/pandas.py.
- Forwarded the exclude_columns argument to the underlying quality.profile call.
- Added a unit test test_pandas_accessor_profiles_dataframe_quality_exclude_columns in tests/test_pandas_accessor.py to verify that columns specified in exclude_columns are excluded from the output DataQualityReport.

Impact it Made:
Enables pandas users to exclude sensitive or irrelevant columns when profiling directly through the pandas accessor API, aligning it with the native ar.profile API.

### DIFF
--- a/arnio/integrations/pandas.py
+++ b/arnio/integrations/pandas.py
@@ -110,6 +110,7 @@ class ArnioPandasAccessor:
     def profile(
         self,
         *,
+        exclude_columns: Sequence[str] | None = None,
         sample_size: int = 5,
         approx_top_values: bool = False,
         approx_top_values_min_unique: int = 1000,
@@ -119,6 +120,7 @@ class ArnioPandasAccessor:
         """Profile DataFrame quality with Arnio."""
         return profile(
             self.to_arframe(),
+            exclude_columns=exclude_columns,
             sample_size=sample_size,
             approx_top_values=approx_top_values,
             approx_top_values_min_unique=approx_top_values_min_unique,
--- a/tests/test_pandas_accessor.py
+++ b/tests/test_pandas_accessor.py
@@ -82,6 +82,17 @@ def test_pandas_accessor_profiles_dataframe_quality():
     assert report.columns["score"].null_count == 1
 
 
+def test_pandas_accessor_profiles_dataframe_quality_exclude_columns():
+    df = pd.DataFrame({"name": [" Alice ", "Bob"], "score": [1.5, None]})
+
+    report = df.arnio.profile(exclude_columns=["score"])
+
+    assert isinstance(report, ar.DataQualityReport)
+    assert report.row_count == 2
+    assert "name" in report.columns
+    assert "score" not in report.columns
+
+
 def test_pandas_accessor_auto_clean_returns_dataframe_and_report():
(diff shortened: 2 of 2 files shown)


im-anishraj (changes requested):
@tmdeveloper007, the accessor forwarding change is correct and the repository CI is green, but the assigned acceptance scope explicitly asked for three focused cases. This PR currently covers only successful omission. Please add the two missing accessor-level regressions: an unknown excluded column must preserve the core API's exception behavior, and the default `exclude_columns=None` path must continue to include all columns. Once those small parity checks are present, this should be ready for exact-head testing.

PR #2611, opened 2026-06-16
fix : validate semantic field type in Field

### LINKED ISSUE #2425: Bug: Field accepts non-string semantic values that crash validation
## Environment / context
- Arnio package: published `arnio==1.19.0`
- Python: 3.13.9 on Windows
- Current `main` source also accepts non-string `Field.semantic`; validation later calls `field_def.semantic.startswith(...)`.

## Steps to reproduce
```python
import pandas as pd
import arnio as ar

frame = ar.from_pandas(pd.DataFrame({"x": ["abc"]}))
schema = {"x": ar.Field(dtype="string", semantic=123)}

ar.validate(frame, schema)
```

## Expected result
`Field(semantic=123)` should be rejected at construction time with a clear `TypeError`, for example `semantic must be a string or None`.

## Actual result
The invalid field is accepted, then validation crashes:

```text
AttributeError: 'int' object has no attribute 'startswith'
```

## Severity / impact
Medium. A malformed schema can crash validation with an internal attribute error instead of returning a clear schema-configuration error. This is especially easy to hit when schemas are built from external config.

## Likely affected area
- `arnio/schema.py::Field.__post_init__`
- Schema JSON deserialization if `semantic` is supplied with a non-string value

## Suggested next action
Add `semantic` type validation in `Field.__post_init__` and cover it in tests for both direct construction and `Schema.from_json()` payloads.

tmdeveloper007:
Fixes #2425

Added type validation for Field.semantic

### DIFF
--- a/arnio/schema.py
+++ b/arnio/schema.py
@@ -927,6 +927,10 @@ class Field:
                     f"pattern is not a valid regular expression: {exc}"
                 ) from exc
             _reject_unsafe_regex_pattern(self.pattern)
+        if self.semantic is not None and not isinstance(self.semantic, str):
+            raise TypeError(
+                f"semantic must be a str or None, got {type(self.semantic).__name__}"
+            )
 
         if self.allowed is not None:
             if not isinstance(self.allowed, (list, tuple, set)):


im-anishraj:

PR #2613, opened 2026-06-16
fix : validate min/max types for all dtypes in Field

### LINKED ISSUE #2423: Bug: Field accepts non-numeric min/max that crash schema validation
## Environment / context
- Arnio package: published `arnio==1.19.0`
- Python: 3.13.9 on Windows
- Also visible in current `main` source: `Field.__post_init__` only type-checks `min`/`max` when `dtype in {"int64", "float64"}`, while `_validate_column()` runs numeric-bound validation whenever `field_def.min` or `field_def.max` is set.

## Steps to reproduce
```python
import pandas as pd
import arnio as ar

frame = ar.from_pandas(pd.DataFrame({"x": ["a", "c"]}))
schema = {"x": ar.Field(dtype="string", min="b")}

ar.validate(frame, schema)
```

A similar crash happens with boolean columns:

```python
frame = ar.from_pandas(pd.DataFrame({"x": [True, False]}))
schema = {"x": ar.Field(dtype="bool", min="b")}
ar.validate(frame, schema)
```

## Expected result
Invalid numeric bounds should be rejected at `Field` construction time with a clear `TypeError`, or validation should convert the problem into a normal schema validation issue. A user-provided schema should not crash validation with an internal NumPy error.

## Actual result
`Field(dtype="string", min="b")` is accepted, then validation crashes:

```text
numpy._core._exceptions._UFuncNoLoopError: ufunc 'less' did not contain a loop with signature matching types (<class 'numpy.dtypes.Float64DType'>, <class 'numpy.dtypes.StrDType'>) -> None
```

## Severity / impact
Medium. A malformed schema can crash validation instead of producing actionable validation output. This is especially confusing because the typed factories (`Int64`, `Float64`) already reject non-numeric bounds, so direct `Field` behaves inconsistently.

## Likely affected area
- `arnio/schema.py::Field.__post_init__`

## Suggested next action
Validate `min` and `max` type whenever either bound is provided, independent of `dtype`, or explicitly reject numeric bounds for non-numeric field dtypes if that is the intended contract. Add regression tests for `Field(dtype="string", min="b")`, `Field(dtype="bool", min="b")`, and `Field(min="b")`.

tmdeveloper007:
Fixes #2423

Fixed Field validation to check min/max types for all dtypes. Allows string min/max for datetime/date dtypes while rejecting non-numeric values for other dtypes.

### DIFF
--- a/arnio/schema.py
+++ b/arnio/schema.py
@@ -904,16 +904,29 @@ class Field:
                 raise TypeError("required_if column name must be a string")
             if not pd.api.types.is_scalar(self.required_if[1]):
                 raise TypeError("required_if expected value must be a scalar")
-        if self.dtype in {"int64", "float64"}:
-            if self.min is not None:
-                if isinstance(self.min, bool) or not isinstance(self.min, (int, float)):
-                    raise TypeError("min must be numeric or None")
-            if self.max is not None:
-                if isinstance(self.max, bool) or not isinstance(self.max, (int, float)):
-                    raise TypeError("max must be numeric or None")
-            if self.min is not None and self.max is not None:
-                if self.min > self.max:
-                    raise ValueError("min must be less than or equal to max")
+        # Always validate min/max types, not just for numeric dtypes
+        if self.min is not None:
+            if isinstance(self.min, bool) or not isinstance(self.min, (int, float, str)):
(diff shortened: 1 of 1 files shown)


im-anishraj:

Issue #2460, opened 2026-06-05
Bug: schema_to_dict fails mixed scalar sets with raw sort TypeError


im-anishraj:
## Environment / context
- Repository: im-anishraj/arnio
- Tested against installed PyPI wheel `arnio==1.19.0`
- Current `main` source inspection shows the same set-normalization path in `arnio/schema_export.py`
- Python: 3.13.9 in local QA virtualenv
- Affected area: schema export / YAML serialization

## Steps to reproduce
```python
import arnio as ar

raw_schema = {
    "code": {
        "type": "STRING",
        "allowed": {1, "1"},
    }
}

ar.schema_to_dict(raw_schema)
ar.schema_to_yaml(raw_schema)
```

## Expected result
Mixed scalar sets should either serialize deterministically or fail with a clear Arnio validation message. This matters because Arnio schema fields already allow mixed scalar `allowed` sets and have tests showing they round-trip through `Schema.to_json()` / `Schema.from_json()`.

A deterministic normalized form could be:
```python
{"fields": {"code": {"allowed": [1, "1"], "type": "STRING"}}}
```

## Actual result
A raw Python sort error leaks out:
```text
TypeError: '<' not supported between instances of 'str' and 'int'
```

The current set normalization uses:
```python
return sorted(_normalize_serializable(v) for v in value)
```
which fails when individually supported scalar values are not mutually comparable.

## Severity / impact
Low-to-medium. Schema export can fail unexpectedly for valid mixed scalar constraints, and the error does not point users toward the problematic field or value.

## Likely affected area
- `arnio/schema_export.py::_normalize_serializable`
- `tests/test_schema_export.py`

## Suggested next action
Normalize sets with a deterministic key that does not require direct cross-type comparison, such as `(type(value).__name__, repr(value))`, or raise a deliberate `TypeError` explaining that mixed-type sets are unsupported by schema export. Add regression tests for `{1, "1"}` and `{None, "missing"}`.

aaniya22:
pls assign this to me under gssoc

im-anishraj:
@aaniya22 this issue is still `status:needs-triage`, and the request does not include an issue-specific implementation plan, so I am not assigning it. Please wait until a maintainer marks it `status:ready`; then request one issue with a concise, scoped plan. We follow one active assigned issue per contributor.

notlexius-glitch:
@im-anishraj 
[...]

PR #2626, opened 2026-06-20
fix: handle mixed-type sets in schema export sort (#2460)

saurabhhhcodes:
**Fixes:** #2460

### Problem
`_normalize_serializable()` used `sorted(...)` directly on set values. When a set contained mixed types like `{1, "1"}`, Python's `sorted()` tried cross-type comparison with `<`, raising `TypeError: '<' not supported between instances of 'str' and 'int'`.

### Changes
Added a `key=lambda v: (type(v).__name__, repr(v))` sort key so sets normalize deterministically regardless of type heterogeneity.


im-anishraj:
@saurabhhhcodes, thanks for the mixed-type set sorting fix. I added the GSSoC classification labels and did an initial triage pass. This PR is currently blocked because the non-Vercel Lint (Black, Ruff, clang-format) check is failing, and the net diff contains unrelated stacked website/accessibility/search/theme changes plus rnio/cleaning.py in addition to the intended rnio/schema_export.py fix. Please rebuild/rebase the branch so the diff is limited to #2460 and its focused regression coverage, then fix the lint failure. Vercel is external authorization noise and is not the blocker.

im-anishraj:

PR #2675, opened 2026-07-06
Bug: `_normalize_scalar` silently passes `bool` values without normalization, risking type confusion in mixed-type columns (#2333)

### LINKED ISSUE #2333: Bug: `_normalize_scalar` silently passes `bool` values without normalization, risking type confusion in mixed-type columns
### What happened?

`_normalize_scalar` correctly skips the integer overflow check for `bool` 
values using `not isinstance(value, bool)`, but then returns the `bool` 
as-is with no further handling. Since `bool` is a subclass of `int` in 
Python, a `True`/`False` value in a mixed-type column (e.g. one containing 
both booleans and strings) bypasses the string-coercion path entirely and 
reaches the C++ binding layer as a raw Python `bool`. This can cause silent 
type confusion where the binding receives `1`/`0` instead of `"True"`/`"False"`.

### Area

pandas conversion

### Minimal reproduction

```python
import pandas as pd
import arnio as ar

# Mixed column: bool + string — should coerce everything to string
df = pd.DataFrame({"flag": [True, "yes", False]}, dtype=object)
frame = ar.from_pandas(df)
result = ar.to_pandas(frame)

print(result["flag"].tolist())
# Expected: ["True", "yes", "False"]
# Actual:   [True, "yes", False]  ← bools not coerced to string
```

### Expected behavior

When a column contains mixed types including `bool` and `str` values, 
`_normalize_scalar` should coerce the `bool` to a string (e.g. `"True"` / 
`"False"`) so it is consistent with the rest of the column. The `_scalar_kind` 
path that triggers all-string coercion should apply to bools in mixed columns 
just as it does for ints and floats.

### Actual behavior

`_normalize_scalar` returns the raw `bool` value unchanged. In a mixed 
bool+string column, `_scalar_kind` sees the bool and returns `"bool"`, 
which conflicts with `"string"` kind from the other values. The 
`"string" in kinds` check on line 174 then coerces int/float values to 
strings but the bool was already added to `values` before this coercion 
— so `True`/`False` survive as Python bools rather than being converted 
to `"True"`/`"False"`.

No error is raised. The incorrect type silently reaches the C++ binding layer.

### Impact

Produces incorrect data

### Arnio version

1.19.0

### Install method

pip

### Operating System

Windows

### Python version

3.12.1

### Environment details

```shell
Summary: C++ accelerated data preparation for pandas and the Python data stack
Home-page: https://github.com/im-anishraj/arnio
Author: 
Author-email: Anish Raj <anishrajyadav97@gmail.com>
License: MIT
Location: /usr/local/python/3.12.1/lib/python3.12/site-packages
Editable project location: /workspaces/arnio
Requires: numpy, pandas
Required-by:
```

### Pre-flight checks

### Additional context

Root cause is in `_normalize_scalar` (convert.py, lines 100–108):

    if isinstance(value, int) and not isinstance(value, bool):
        # overflow check — correctly skips bools
        ...
    if isinstance(value, float):
        return _to_binding_safe(value)
    return value   # ← bool falls here, returned raw

Fix suggestion: add an explicit `bool` check before the `int` guard, 
or ensure that bool values in a mixed-kind column are caught by the 
string-coercion pass in `_series_to_python_values`.

Related: `_scalar_kind` also has no explicit `bool` branch that feeds 
into the mixed-type coercion logic — both functions need to be reviewed together.

saurabhhhcodes:
Fixes #2333

### DIFF
--- a/README.md
+++ b/README.md
@@ -2161,3 +2161,4 @@ loaded_steps = ar.load_pipeline("my_pipeline.json")
 ## Security
 
 Please review our [Security Policy](SECURITY.md) for responsible vulnerability reporting guidelines.
+# TODO: bug: `_normalize_scalar` silently passes `bool` values without normalization, risking type confusion in mixed-type columns (#2333)


im-anishraj:

PR #2686, opened 2026-07-06
Bug: schema_to_dict fails mixed scalar sets with raw sort TypeError (#2460)

### LINKED ISSUE #2460: Bug: schema_to_dict fails mixed scalar sets with raw sort TypeError

saurabhhhcodes:
Fixes #2460

### DIFF
--- a/README.md
+++ b/README.md
@@ -2161,3 +2161,4 @@ loaded_steps = ar.load_pipeline("my_pipeline.json")
 ## Security
 
 Please review our [Security Policy](SECURITY.md) for responsible vulnerability reporting guidelines.
+# TODO: bug: schema_to_dict fails mixed scalar sets with raw sort typeerror (#2460)


im-anishraj:

PR #2694, opened 2026-07-06
Bug: ArFrame iteration falls back to integer column access (#2451)

### LINKED ISSUE #2451: Bug: ArFrame iteration falls back to integer column access
## Environment / context
- Repository: `im-anishraj/arnio`
- Local source: current `main` after `78d5ef1e5c7b3af48913a97ae1a02382139532f2`
- Published package checked: `arnio==1.19.0`
- Python: 3.13.9 on Windows
- Area: `ArFrame` Python API / container protocol behavior

## Steps to reproduce
```python
import arnio as ar
import pandas as pd

frame = ar.from_pandas(pd.DataFrame({"a": [1, 2, 3], "b": ["x", "y", "z"]}))

iter(frame)          # returns an iterator object
next(iter(frame))    # raises TypeError from frame[0]
list(frame)          # raises the same TypeError
```

## Expected result
`ArFrame` should expose an intentional iteration contract:

- either iterate column names, matching common DataFrame-style behavior, or
- explicitly be non-iterable with a clear `TypeError` before Python falls back to integer indexing.

## Actual result
Because `ArFrame` implements `__len__` and `__getitem__` but does not define `__iter__`, Python treats it as an old-style sequence and calls `frame[0]`. That produces a misleading column-key error:

```text
TypeError: column key must be a str or list of str, got 'int'
```

`iter(frame)` itself succeeds, so the failure appears only when the iterator is consumed.

## Severity / impact
Low severity, but it is a public API ergonomics issue. Users may naturally try `list(frame)`, `tuple(frame)`, or generic iterable utilities and get a confusing error unrelated to their intent.

## Likely affected area
- `arnio/frame.py`
- `ArFrame.__len__`
- `ArFrame.__getitem__`
- missing explicit `ArFrame.__iter__`

## Suggested next action
Decide the intended contract and add tests for it. The most predictable option may be to implement `__iter__` as `return iter(self.columns)`, matching pandas-style column iteration. If `ArFrame` should not be iterable, define `__iter__` to raise a clearer `TypeError`.

saurabhhhcodes:
Fixes #2451

### DIFF
--- a/README.md
+++ b/README.md
@@ -2161,3 +2161,4 @@ loaded_steps = ar.load_pipeline("my_pipeline.json")
 ## Security
 
 Please review our [Security Policy](SECURITY.md) for responsible vulnerability reporting guidelines.
+# TODO: bug: arframe iteration falls back to integer column access (#2451)


im-anishraj:

PR #2701, opened 2026-07-06
Bug: read_csv does not short-circuit when nrows=0, unlike read_jsonl which returns an empty frame without any I/O (#2070)

### LINKED ISSUE #2070: Bug: read_csv does not short-circuit when nrows=0, unlike read_jsonl which returns an empty frame without any I/O

saurabhhhcodes:
Fixes #2070

### DIFF
--- a/README.md
+++ b/README.md
@@ -2161,3 +2161,4 @@ loaded_steps = ar.load_pipeline("my_pipeline.json")
 ## Security
 
 Please review our [Security Policy](SECURITY.md) for responsible vulnerability reporting guidelines.
+# TODO: bug: read_csv does not short-circuit when nrows=0, unlike read_jsonl which returns an empty frame without any i/o (#2070)


im-anishraj:

PR #2702, opened 2026-07-06
Bug: sniff_delimiter misdetects punctuation in single-column free-text files (#2446)

### LINKED ISSUE #2446: Bug: sniff_delimiter misdetects punctuation in single-column free-text files
## Environment / context

- Package under test: `arnio==1.19.0` from PyPI
- Current source check: `origin/main` still uses the same `sniff_delimiter()` scoring logic in `arnio/io.py`
- Python: 3.13.9
- Platform: Windows
- Audit context: installed wheel tested outside the repository checkout

## Steps to reproduce

```python
from pathlib import Path
import tempfile
import arnio as ar

path = Path(tempfile.mkdtemp()) / "notes.csv"
path.write_text(
    "notes\n"
    "likes apples, bananas\n"
    "x, y, z\n",
    encoding="utf-8",
)

delimiter = ar.sniff_delimiter(path)
print(repr(delimiter))

# Use the detected delimiter with the parser.
ar.read_csv(path, delimiter=delimiter)
```

A semicolon-only variant has the same shape:

```python
path.write_text("notes\na;b;c\nx;y;z\n", encoding="utf-8")
delimiter = ar.sniff_delimiter(path)  # returns ';'
ar.read_csv(path, delimiter=delimiter)
```

## Expected result

`sniff_delimiter()` should not return a delimiter when the sample is structurally a single-column file and candidate punctuation appears only inside free-text values. It should raise a clear `ValueError`, similar to the existing single-column/no-delimiter behavior, so callers do not pass a bad delimiter into `read_csv()`.

## Actual result

`sniff_delimiter()` returns a candidate delimiter based on punctuation in the data rows:

```text
','
```

Using that delimiter immediately makes the same file fail to parse:

```text
CsvReadError: CSV row 2 has 2 fields; expected 1
```

For the semicolon variant, it returns `';'` and `read_csv()` fails because row 2 has 3 fields while the header has 1.

## Severity / impact

Medium. This affects delimiter auto-detection workflows for real CSV-like files that contain one free-text column with commas, semicolons, or pipe characters. The helper returns a confident-looking delimiter even though the detected delimiter cannot parse the file consistently.

## Likely affected area

- CSV delimiter sniffing
- CSV parser ergonomics
- Public Python API

## Suggested next action

Update `sniff_delimiter()` scoring to require structural consistency across sampled non-empty lines, including the header. A candidate delimiter should be rejected when the modal delimiter count is absent from the header or when applying it would produce inconsistent field counts for the sampled rows. Add regression tests for single-column free-text values containing comma, semicolon, tab, and pipe-like punctuation.

saurabhhhcodes:
Fixes #2446

### DIFF
--- a/README.md
+++ b/README.md
@@ -2161,3 +2161,4 @@ loaded_steps = ar.load_pipeline("my_pipeline.json")
 ## Security
 
 Please review our [Security Policy](SECURITY.md) for responsible vulnerability reporting guidelines.
+# TODO: bug: sniff_delimiter misdetects punctuation in single-column free-text files (#2446)


im-anishraj:

PR #2705, opened 2026-07-06
Performance: Remove inline pandas import in `_normalize_validator_result` to fix severe validation bottleneck (#2067)

### LINKED ISSUE #2067: Performance: Remove inline pandas import in `_normalize_validator_result` to fix severe validation bottleneck
### Workload

Running custom schema validators over large datasets (e.g., 1M+ rows) where a custom function is mapped across the dataset row-by-row.

In `schema.py` (around line 1419), `_normalize_validator_result()` is called for every single row via a lambda function. Inside this function (line 1436), there is a `try...except` block that executes an inline `import pandas as _pd` to check for `pd.NA`. 

Even with Python caching modules, doing an inline import and exception handling inside a per-row map operation on a large dataset causes a massive performance hit. Furthermore, the comment mentions doing this to "avoid importing pandas at module level," but `pandas` is already explicitly imported at the top of the file on line 18.

### Target improvement

Remove the inline `import pandas as _pd` inside `_normalize_validator_result`. Since `pandas` is already imported at the module level, this function can just directly reference `pd.NA` or use `pd.isna()`. This will eliminate the per-row import/exception overhead and vastly speed up custom validation execution times.

I would love to be assigned to this issue and fix it for GSSoC 2026!

### Acceptance criteria

- [ ] Reproducible benchmark or regression test added
- [ ] Improvement is measured before/after
- [ ] No correctness regressions in existing tests

saurabhhhcodes:
Fixes #2067

### DIFF
--- a/README.md
+++ b/README.md
@@ -2161,3 +2161,4 @@ loaded_steps = ar.load_pipeline("my_pipeline.json")
 ## Security
 
 Please review our [Security Policy](SECURITY.md) for responsible vulnerability reporting guidelines.
+# TODO: performance: remove inline pandas import in `_normalize_validator_result` to fix severe validation bottleneck (#2067)


im-anishraj:

PR #2711, opened 2026-07-06
Bug: Bug: two-pass CSV reading can silently inject null values when a source file changes between passes (#1990)

### LINKED ISSUE #1990: Bug: Bug: two-pass CSV reading can silently inject null values when a source file changes between passes
### What happened?

The CSV reader performs:

1. a type-inference pass
2. a data-ingestion pass

using two separate file opens.

If the underlying file changes between those passes, inferred column types may no longer match the ingested content.

When parsing then fails, values become nulls without any warning.

### Area

### Minimal reproduction

A deterministic standalone reproduction is difficult because it requires the CSV file to change between the reader's type-inference pass and ingestion pass.

The issue arises because csv_reader.cpp opens the same file twice. If the file is modified, replaced, or atomically swapped between those passes, inferred types can diverge from ingested content and parse failures become silent null values.

The bug was identified through source inspection of the two-pass read implementation.

### Expected behavior

A CSV read operation should observe a consistent snapshot of the source file.

If the file changes between schema inference and ingestion, the reader should either:

- fail fast
- verify file identity
- reuse a single buffered snapshot

### Actual behavior

The file is reopened for the second pass.

If content changes between passes, parse failures are converted into null values while the inferred schema remains unchanged.

The resulting frame appears valid but contains silently corrupted data.

### Impact

Produces incorrect data

### Arnio version

1.1.0

### Install method

Source inspection

### Operating System

Linux

### Python version

3.12.4

### Environment details

```shell
Issue identified through source inspection of the two-pass CSV reader implementation.

The report is based on analysis of separate schema-inference and ingestion passes that reopen the source file.
```

### Pre-flight checks

### Additional context

_No response_

saurabhhhcodes:
Fixes #1990

### DIFF
--- a/README.md
+++ b/README.md
@@ -2161,3 +2161,4 @@ loaded_steps = ar.load_pipeline("my_pipeline.json")
 ## Security
 
 Please review our [Security Policy](SECURITY.md) for responsible vulnerability reporting guidelines.
+# TODO: bug: bug: two-pass csv reading can silently inject null values when a source file changes between passes (#1990)


im-anishraj:

PR #2727, opened 2026-07-06
Bug: published 1.19.0 from_records silently drops data when columns is empty (#2430)

### LINKED ISSUE #2430: Bug: published 1.19.0 from_records silently drops data when columns is empty
## Environment / context
- Package: `arnio==1.19.0` installed from PyPI wheel
- Platform: Windows, Python 3.13.9 virtual environment
- Import path during repro: `.audit_venv\Lib\site-packages\arnio`
- Current main has regression coverage in `tests/test_frame.py` requiring this case to raise `ValueError`.

## Steps to reproduce
```python
import arnio as ar

frame = ar.from_records([{"a": 1}, {"a": 2}], columns=[])
print(frame.shape)
print(frame.columns)
print(frame.to_dict())
print(frame.is_empty)
```

## Expected result
Passing `columns=[]` with dict records should raise a clear `ValueError`, for example explaining that `columns` must not be empty and that callers should pass `columns=None` to infer keys.

## Actual result
The published wheel silently returns an empty frame:
```text
shape (0, 0)
columns []
to_dict {}
is_empty True
```

## Severity / impact
Medium. This can silently drop all records and columns when a caller accidentally passes an empty column list, making downstream validation, profiling, or exports operate on empty data instead of failing fast.

## Likely affected area
- `ArFrame.from_records`
- `arnio.from_records`
- Release/package parity for the published wheel

## Suggested next action
Include the current empty-column guard in the next release and add a wheel-level parity check for `from_records(..., columns=[])` so the published artifact cannot regress to silent empty-frame behavior.

saurabhhhcodes:
Fixes #2430

### DIFF
--- a/README.md
+++ b/README.md
@@ -2161,3 +2161,4 @@ loaded_steps = ar.load_pipeline("my_pipeline.json")
 ## Security
 
 Please review our [Security Policy](SECURITY.md) for responsible vulnerability reporting guidelines.
+# TODO: bug: published 1.19.0 from_records silently drops data when columns is empty (#2430)


im-anishraj:

PR #2742, opened 2026-07-06
Bug: published 1.19.0 select_dtypes raises instead of returning zero-column frames (#2409)

### LINKED ISSUE #2409: Bug: published 1.19.0 select_dtypes raises instead of returning zero-column frames
## Environment / context

- Package under test: `arnio==1.19.0` from PyPI
- Python: 3.13.9
- Platform: Windows
- Audit context: installed wheel tested outside the repository checkout
- Related source fix: #1954 fixed this on main by returning a zero-column frame that preserves row count.

## Steps to reproduce

```python
import arnio as ar
import pandas as pd

    "a": [1, 2],
    "b": ["x", "y"],
    "c": [True, False],
}))

result = frame.select_dtypes(include="float64")
print(result.shape)
print(result.columns)
```

## Expected result

Selecting a dtype that is absent should behave like a safe selection operation and return a zero-column frame with the original row count:

```python
(2, 0)
[]
```

Current source has tests covering this behavior.

## Actual result

On `arnio==1.19.0`, the call raises:

```text
ValueError: No columns match the dtype selection. Frame dtypes: {'a': 'int64', 'b': 'string', 'c': 'bool'}.
```

## Severity / impact

Medium. Users expect dtype filtering to be composable in pipelines. Raising on valid empty selections makes schema-dependent workflows brittle and prevents downstream logic from safely handling zero-column frames.

## Likely affected area

- `ArFrame.select_dtypes`
- pandas interoperability semantics
- PyPI/release artifact parity

## Suggested next action

Publish a patched wheel containing the current zero-column `select_dtypes` behavior and add a wheel parity test that verifies both `include` with absent dtypes and `exclude` that removes all columns preserve row count.

saurabhhhcodes:
Fixes #2409

### DIFF
--- a/README.md
+++ b/README.md
@@ -2161,3 +2161,4 @@ loaded_steps = ar.load_pipeline("my_pipeline.json")
 ## Security
 
 Please review our [Security Policy](SECURITY.md) for responsible vulnerability reporting guidelines.
+# TODO: bug: published 1.19.0 select_dtypes raises instead of returning zero-column frames (#2409)


im-anishraj:

PR #2756, opened 2026-07-06
DX: check_env reports ready even when the build toolchain is blocked (#2393)

### LINKED ISSUE #2393: DX: check_env reports ready even when the build toolchain is blocked
## Environment / Context
- Current `main` commit checked locally: `db905172`
- OS: Windows
- Python: 3.13.9
- Command: `python examples/check_env.py`
- Local build tools: CMake available, `cl` and `nmake` not on `PATH`

## Steps to Reproduce
1. Use a Windows shell where optional Python dependencies are installed but MSVC/NMake are not available on `PATH`.
2. Run:

```powershell
python examples/check_env.py
```

## Expected Result
The dashboard should make the blocked source-build state unambiguous. If required build tools are missing, the final summary should not say the environment is fully ready.

A clearer final message would distinguish optional runtime dependencies from source-build readiness, for example:

```text
All optional dependencies are installed, but source builds are blocked until the missing build tools are available.
```

## Actual Result
The dashboard correctly reports the build-tool failure:

```text
[BUILD BLOCKED] Source installs like `pip install -e ".[dev]"` may fail until the missing build tools are available.
```

But the final summary still says:

```text
All optional dependencies are successfully installed! You are ready to go.
```

## Severity / Impact
Low to medium. This can confuse new contributors because the same command reports a blocked local build and a ready environment. It is especially misleading after an editable install fails with `nmake` / `CMAKE_CXX_COMPILER` errors.

## Likely Affected Area
- `examples/check_env.py`
- Developer onboarding / Windows setup diagnostics

## Suggested Next Action
Update `print_dashboard()` so the final success message accounts for required build-tool status. Add a regression test where optional dependencies are installed but required build tools are missing, and assert that the final output does not include `You are ready to go`.

saurabhhhcodes:
Fixes #2393

### DIFF
--- a/README.md
+++ b/README.md
@@ -2161,3 +2161,4 @@ loaded_steps = ar.load_pipeline("my_pipeline.json")
 ## Security
 
 Please review our [Security Policy](SECURITY.md) for responsible vulnerability reporting guidelines.
+# TODO: dx: check_env reports ready even when the build toolchain is blocked (#2393)


im-anishraj:

